PR #1, opened 2017-10-23
Expand the readme, and add ability to configure nginx proxy

katebee:
## What does this change?

- Expand the Security HQ readme to give a bit more information on getting the project running for the first time

- Add nginx config file to the project, and put explanation in the readme based on the use of a nginx proxy in other projects

### What is the value of this?

Some stuff I wrote after getting started with the project; may or may not be beneficial to add to Master, or if it is better to just ensure that it is on GitHub and can be tracked down if needed.

Adding a nginx proxy means you can access the webapp via https://security.local.dev-gutools.co.uk

picture 46


adamnfish:
Excellent! I put some very minor points in comments, but no dramas, this looks good.

:+1:

PR #7, opened 2017-11-06
Resolve the unexpected errors when parsing SG open ports

katebee:
Trusted Advisor was failing if the `SGIds` did not contain a `vpcId`, this adds a new case to handle `EC2 classic`, and updates the tests to include one for this issue.

This also moves away from using `throw new RuntimeException` when attempting to parse the metadata from Trusted Advisor

### DIFF
--- a/hq/app/aws/support/TrustedAdvisor.scala
+++ b/hq/app/aws/support/TrustedAdvisor.scala
@@ -54,12 +54,16 @@ object TrustedAdvisor {
     handleAWSErrs(awsToScala(client.describeTrustedAdvisorCheckResultAsync)(request))
   }
 
-  def parseTrustedAdvisorCheckResult[A <: TrustedAdvisorCheckDetails](parseDetails: TrustedAdvisorResourceDetail => A)(result: DescribeTrustedAdvisorCheckResultResult): TrustedAdvisorDetailsResult[A] = {
-    TrustedAdvisorDetailsResult(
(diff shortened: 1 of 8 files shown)


hq/app/aws/support/TrustedAdvisor.scala
```diff
@@ -54,12 +54,15 @@ object TrustedAdvisor {
     handleAWSErrs(awsToScala(client.describeTrustedAdvisorCheckResultAsync)(request))
   }
 
-  def parseTrustedAdvisorCheckResult[A <: TrustedAdvisorCheckDetails](parseDetails: TrustedAdvisorResourceDetail => A)(result: DescribeTrustedAdvisorCheckResultResult): TrustedAdvisorDetailsResult[A] = {
-    TrustedAdvisorDetailsResult(
+  def parseTrustedAdvisorCheckResult[A <: TrustedAdvisorCheckDetails](parseDetails: TrustedAdvisorResourceDetail => Attempt[A], ec: ExecutionContext)(result: DescribeTrustedAdvisorCheckResultResult): Attempt[TrustedAdvisorDetailsResult[A]] = {
+    implicit val executionContext: ExecutionContext = ec
```
adamnfish: As discussed IRL this isn't ideal, probably a sign that we should simplify the approach slightly in the future.

adamnfish: (but fine for now)

adamnfish:
Looks fantastic to me, great fixes :+1:

PR #8, opened 2017-11-07
Use "clue" to provide test feedback on failed attempts

adamnfish:
Here's an improvement copied from another project where I used this approach (as discussed yesterday, Kate).

cc @katebee @abdusahin


abdusahin:
👍

PR #9, opened 2017-11-07
Reverse sort for accounts with flagged resources

adamnfish:
Now puts accounts with the highest number of flagged resources first.


abdusahin:
👍

PR #10, opened 2017-11-08
Remove unused links & tidy up

abdusahin:
- Removed old links
- Renamed AWS trusted advisor key names
- Removed dependencies


abdusahin:
@adamnfish / @katebee can you have a look pls ?

hq/app/aws/support/TrustedAdvisorExposedIAMKeys.scala
```diff
@@ -11,10 +11,10 @@ import scala.concurrent.ExecutionContext
 
 
 object TrustedAdvisorExposedIAMKeys {
-  val exposedIAMKeys = "12Fnkpl8Y5"
+  val AWS_EXPOSED_ACCESS_KEYS_IDENTIFIER = "12Fnkpl8Y5"
```
katebee: Minor point: I have mixed feelings about conventions on caps, I can see the benefits of visually distinguishing these parameters (particularly as objects grow in length), however, I think IntelliJ helps enough that it may not be necessary? 😄 

I generally perceive all-caps to indicate global values (such as those coming in from a .conf), however I appreciate this is a habit from other projects, (and JavaScript land). One consideration is that humans find text in all-caps harder to parse, since most adults read the shape of a word and `WRITING IN ALL CAPS` turns words into undifferentiated blocks, therefore [slowing down reading speed](http://uxmovement.com/content/all-caps-hard-for-users-to-read).

Maybe this is a good approach for now, but if we reach a point where a lot of code is written in all-caps we could reevaluate? 🤔

PR #11, opened 2017-11-10
Credentials auditing

abdusahin:
- Retrieve credentials audit report from AWS
- Add controller/view stubs for report display

### DIFF
--- a/hq/app/aws/iam/CredentialsReport.scala
+++ b/hq/app/aws/iam/CredentialsReport.scala
@@ -1,0 +1,81 @@
+package aws.iam
+
+import java.io.StringReader
+
+import model.{IAMCredential, IAMCredentialsReport}
+import com.github.tototoshi.csv._
+import org.joda.time.DateTime
+import org.joda.time.format.ISODateTimeFormat
+import com.amazonaws.regions.{Region, Regions}
+import com.amazonaws.services.identitymanagement.model.{GenerateCredentialReportResult, GetCredentialReportResult}
+
+
+object CredentialsReport {
+
+  def isComplete(report: GenerateCredentialReportResult) = report.getState == "COMPLETE"
+
+  def extractReport(result: GetCredentialReportResult): IAMCredentialsReport = {
+    val content = result.getContent
+    val dest = new Array[Byte](content.capacity())
+    content.get(dest, 0, dest.length)
+      IAMCredentialsReport(
+        new DateTime(result.getGeneratedTime),
+        parseCredentialsReport(new String(dest))
+      )
+  }
+
+  // parsing the credentials report CSV file
+  val isoDateTimeParser = ISODateTimeFormat.dateTimeParser().withZoneUTC()
+
+  def parseBoolean(cell: String): Option[Boolean] = {
+    if (cell == "true") Some(true)
+    else if (cell == "false") Some(false)
+    else None
+  }
+
+  def parseDateTimeOpt(cell: String): Option[DateTime] = {
+    if (cell == "N/A" || cell == "not_supported") None
+    else Some(isoDateTimeParser.parseDateTime(cell))
+  }
+
+  def parseRegionOpt(cell: String): Option[Region] = {
+    if (cell == "N/A") None
+    else Some(Region.getRegion(Regions.fromName(cell)))
+  }
+
+  def parseStrOpt(cell: String): Option[String] = {
+    if (cell == "N/A") None
+    else Some(cell)
+  }
+
+  def parseCredentialsReport(contents: String): List[IAMCredential] = {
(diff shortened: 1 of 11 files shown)


abdusahin:
@adamnfish , @katebee here is a PR for review

hq/app/aws/iam/CredentialsReport.scala
```diff
@@ -0,0 +1,81 @@
...
+    val dest = new Array[Byte](content.capacity())
+    content.get(dest, 0, dest.length)
+      IAMCredentialsReport(
+        new DateTime(result.getGeneratedTime),
+        parseCredentialsReport(new String(dest))
+      )
+  }
+
+  // parsing the credentials report CSV file
+  val isoDateTimeParser = ISODateTimeFormat.dateTimeParser().withZoneUTC()
```
adamnfish: I think this should be in [DateUtils](https://github.com/guardian/security-hq/blob/master/hq/app/logic/DateUtils.scala)

hq/app/aws/iam/CredentialsReport.scala
```diff
@@ -0,0 +1,81 @@
...
+    if (cell == "N/A") None
+    else Some(cell)
+  }
+
+  def parseCredentialsReport(contents: String): List[IAMCredential] = {
+    val report = CSVReader.open(new StringReader(contents)).allWithHeaders().map { row =>
+      IAMCredential(
+        user = row.getOrElse("user", "no username available"),
+        arn = row.getOrElse("arn", "no ARN available"),
+        creationTime = row.get("user_creation_time").flatMap(parseDateTimeOpt).get,
```
adamnfish: TODO error handling (as discussed IRL), but this is great for now.

hq/app/logic/Retry.scala
```diff
@@ -0,0 +1,24 @@
...
+import utils.attempt.{Attempt, Failure}
+
+import scala.concurrent.ExecutionContext
+object Retry {
+
+  def until[A](body: => Attempt[A], predicate: A => Boolean, failureMessage: String, maxAttempt: Int = 10)(implicit ec: ExecutionContext): Attempt[A] = {
+
+    def loop(numberOfTries: Int): Attempt[A] = {
+      if (numberOfTries >= maxAttempt ) {
+        Attempt.Left[A](Failure("MAX_ATTEMPT_LIMIT", failureMessage, 500))
```
adamnfish: Might be useful to put the friendly message in the log message as well. As it stands we're just going to see MAX_ATTEMPT_LIMIT in the logs where this happens, which isn't heaps of context.

hq/app/logic/Retry.scala
```diff
@@ -0,0 +1,24 @@
+package logic
+
+import utils.attempt.{Attempt, Failure}
+
+import scala.concurrent.ExecutionContext
+object Retry {
+
+  def until[A](body: => Attempt[A], predicate: A => Boolean, failureMessage: String, maxAttempt: Int = 10)(implicit ec: ExecutionContext): Attempt[A] = {
```
adamnfish: Very minor point, but it's a Scala convention to make functions require parentheses when they perform side-effects. This would just make it explicit that we're planning to do IO here. Parens-less pure access is fine, but we know `body` will go to AWS.

hq/conf/routes
```diff
@@ -7,6 +7,7 @@ GET        /                                          controllers.HQController.i
 
 GET        /iam                                       controllers.HQController.iam
 GET        /iam/:accountId                            controllers.HQController.iamAccount(accountId)
+GET        /iam/:accountId/cred-report                controllers.HQController.generateCredentialsReport(accountId)
```
adamnfish: Is there a reason to put this on a separate path? It feels like this is exactly what people will want to see when they go to `/iam/:accountId`.

hq/app/logic/Retry.scala
abdusahin: how about` s"MAX_ATTEMPT_LIMIT_REACHED: $failureMessage"`

hq/conf/routes
abdusahin: nope, just keeping it separate while testing.

hq/app/logic/Retry.scala
abdusahin: i think putting `()` will make `until` usage a bit uglier  as i have to use it like `{ () => IAMClient.generateCredentialsReport(client) }`.  
Retry is also not bound to AWS as it's generic at the moment. I think the convention around parentheses cares about type function declarations. (e.g. `def foo ` vs `def foo()`) . There is an example below.
https://docs.scala-lang.org/style/naming-conventions.html#parentheses

adamnfish:
:+1:

PR #14, opened 2017-11-13
 - Handle 'no_information' case for password last updated field

abdusahin:
- Introduce delay to the Retry logic

### DIFF
--- a/hq/app/controllers/HQController.scala
+++ b/hq/app/controllers/HQController.scala
@@ -10,7 +10,7 @@ import play.api._
 import play.api.libs.ws.WSClient
 import play.api.mvc._
 import utils.attempt.PlayIntegration.attempt
-
+import scala.concurrent.duration._
 import scala.concurrent.ExecutionContext
 
 
@@ -40,11 +40,12 @@ class HQController(val config: Configuration)
   }
 
   def generateCredentialsReport(accountId: String) = AuthAction.async {
+    val delay = 3.seconds
     attempt {
       for {
         account <- AWS.lookupAccount(accountId, accounts)
         client = IAMClient.client(account)
-         _ <- Retry.until(IAMClient.generateCredentialsReport(client), CredentialsReport.isComplete, "Failed to generate credentials report")
+         _ <- Retry.until(IAMClient.generateCredentialsReport(client), CredentialsReport.isComplete, "Failed to generate credentials report", delay)
         report <- IAMClient.getCredentialsReport(client)
       } yield Ok(views.html.iam.iamCredReport(report))
     }
--- a/hq/app/controllers/HQController.scala
+++ b/hq/app/controllers/HQController.scala
@@ -10,7 +10,7 @@ import play.api._
 import play.api.libs.ws.WSClient
 import play.api.mvc._
 import utils.attempt.PlayIntegration.attempt
-import scala.concurrent.duration._
+
 import scala.concurrent.ExecutionContext
 
 
@@ -40,16 +40,14 @@ class HQController(val config: Configuration)
   }
 
   def generateCredentialsReport(accountId: String) = AuthAction.async {
-    val delay = 3.seconds
     attempt {
       for {
         account <- AWS.lookupAccount(accountId, accounts)
-        client = IAMClient.client(account)
-         _ <- Retry.until(IAMClient.generateCredentialsReport(client), CredentialsReport.isComplete, "Failed to generate credentials report", delay)
-        report <- IAMClient.getCredentialsReport(client)
+        report <- IAMClient.getCredentialsReport(account)
       } yield Ok(views.html.iam.iamCredReport(report))
     }
   }
+
 
 
 }
--- a/hq/app/logic/Retry.scala
+++ b/hq/app/logic/Retry.scala
@@ -8,68 +8,24 @@ object Retry {
 
   /**
     * execute the body until the predicate is satisfied or maxAttempt is reached, with the delay between each body execution.
-    * @param body
-    * @param predicate
-    * @param failureMessage
-    * @param delay
-    * @param maxAttempt
-    * @param ec
-    * @tparam A
-    * @return
     */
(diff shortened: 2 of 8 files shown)


abdusahin:
@adamnfish can you have a look ?

hq/test/logic/RetryTest.scala
```diff
@@ -12,6 +14,7 @@ class RetryTest extends WordSpec with Matchers with AttemptValues {
     val Started = "STARTED"
     val maxAttempt = 10
     val predicateF = { status: String => status == Complete }
+    val testDelay = 100.millisecond
```
adamnfish: I'd divide this (and corresponding references) by 10 for the tests to keep them quick.

hq/test/logic/RetryTest.scala
```diff
@@ -22,14 +25,14 @@ class RetryTest extends WordSpec with Matchers with AttemptValues {
         body
       }
 
-      Retry.until(testBody, predicateF, failMessage).isFailedAttempt shouldBe true
+      Retry.until(testBody, predicateF, failMessage, testDelay).isFailedAttempt shouldBe true
```
adamnfish: If you aren't testing the time delays there's no reason to slow the tests down, why not use `Duration.Zero` for these?

hq/app/utils/attempt/Attempt.scala
```diff
@@ -155,4 +160,47 @@ object Attempt {
...
       Attempt(ferr.map(scala.Left(_)))
   }
+
+
+
+
+  /**
+    * @see https://stackoverflow.com/questions/16359849/scala-scheduledfuture
+    * */
+  object Delayed {
```
adamnfish: The formatting is off on this block because it's been pasted from SO. Can you make it consistent with the codebase?

hq/app/logic/Retry.scala
```diff
@@ -3,22 +3,29 @@ package logic
...
 import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.FiniteDuration
 object Retry {
 
-  def until[A](body: => Attempt[A], predicate: A => Boolean, failureMessage: String, maxAttempt: Int = 10)(implicit ec: ExecutionContext): Attempt[A] = {
+  /**
+    * execute the body until the predicate is satisfied or maxAttempt is reached, with the delay between each body execution.
+    */
+  def until[A](body: => Attempt[A], predicate: A => Boolean, failureMessage: String, delay: FiniteDuration , maxAttempt: Int = 10)(implicit ec: ExecutionContext): Attempt[A] = {
+    def determineBody(tryCnt: Int) : Attempt[A] =  if (tryCnt == 0) body else body.delay(delay)
```
adamnfish: Is there any reason not to inline this? The function definition is larger than the logic and it's separated so it can be tested. This means the indirection doesn't seem to achieve much?

hq/app/utils/attempt/Attempt.scala
```diff
@@ -47,6 +48,10 @@ case class Attempt[A] private (underlying: Future[Either[FailedAttempt, A]]) {
       scala.Left(apiErrors)
     }
   }
+
+  def delay(delay: FiniteDuration )(implicit ec: ExecutionContext): Attempt[A] = {
```
adamnfish: The formatting is a bit off here as well? It's a very minor point here.

hq/test/aws/iam/CredentialsReportTest.scala
```diff
@@ -99,6 +103,14 @@ class CredentialsReportTest extends FreeSpec with Matchers with OptionValues wit
         CredentialsReport.tryParsingReport(testReport).isFailedAttempt() shouldBe true
       }
 
+      "handles - 'no_information' data for password_last_used field" in {
```
adamnfish: Great catch

hq/app/utils/attempt/Attempt.scala
```diff
@@ -155,4 +160,47 @@ object Attempt {
...
+            )
+          }
+        }
+      )
+      prom.future
+    }
+    
+    def apply[A]( delay: FiniteDuration )( body: => A )(implicit ctx: ExecutionContext): Attempt[A] = {
+      Attempt.fromFuture( makeTask( body )( timer.schedule( _, delay.toMillis ) )) {
+        case th => FailedAttempt(Failure(th.getMessage, "Cannot execute the delayed task", 500))
```
adamnfish: There's a shortcut for making a `FailedAttempt` instance from a single `Failure` because it's such a common thing to do. Off the top of my head it's `.asAttempt`, but IntelliJ will help.

hq/app/controllers/HQController.scala
```diff
@@ -40,11 +40,12 @@ class HQController(val config: Configuration)
   }
 
   def generateCredentialsReport(accountId: String) = AuthAction.async {
+    val delay = 3.seconds
     attempt {
       for {
         account <- AWS.lookupAccount(accountId, accounts)
         client = IAMClient.client(account)
-         _ <- Retry.until(IAMClient.generateCredentialsReport(client), CredentialsReport.isComplete, "Failed to generate credentials report")
+         _ <- Retry.until(IAMClient.generateCredentialsReport(client), CredentialsReport.isComplete, "Failed to generate credentials report", delay)
```
adamnfish: This is *borderline* too complicated for the controller and I think the `delay` constant pushes it over the edge for my taste. This line's logic and string constants are specific to the credentials report and the IAM client's functionality. I'd probably prefer to put them in a method on the IAMClient object so that the controller can just ask for a result without needing to know about how it worked.

hq/app/logic/Retry.scala
abdusahin: 👍

hq/app/controllers/HQController.scala
abdusahin: will move to IAMClient

hq/test/logic/RetryTest.scala
abdusahin: 👍

hq/test/logic/RetryTest.scala
abdusahin: 👍

adamnfish:
:+1:

PR #17, opened 2017-11-23
Security groups sorting

abdusahin:
-Display security groups in-use
-Sort flags by alert level
-Sort flags by port priority

### DIFF
--- a/hq/app/aws/ec2/EC2.scala
+++ b/hq/app/aws/ec2/EC2.scala
@@ -56,7 +56,7 @@ object EC2 {
       sgResult <- TrustedAdvisorSGOpenPorts.getSGOpenPorts(supportClient)
       sgUsage <- getSgsUsage(sgResult, account)
       flaggedSgs = sgResult.flaggedResources.filter(_.status != "ok")
-    } yield flaggedSgs.map(sgOpenPortsDetail => sgOpenPortsDetail -> sgUsage.getOrElse(sgOpenPortsDetail.id, Set.empty))
+    } yield sortSecurityGroupsByInUse(flaggedSgs.map(sgOpenPortsDetail => sgOpenPortsDetail -> sgUsage.getOrElse(sgOpenPortsDetail.id, Set.empty)))
   }
 
   def allFlaggedSgs(accounts: List[AwsAccount])(implicit ec: ExecutionContext): Attempt[List[(AwsAccount, Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]])]] = {
@@ -65,6 +65,10 @@ object EC2 {
         flaggedSgsForAccount(account).asFuture.map(account -> _)
       }
     }
+  }
+
+  private [ec2] def sortSecurityGroupsByInUse(sgsPortFlags : List[(SGOpenPortsDetail, Set[SGInUse])]) = {
+    sgsPortFlags.sortWith{ case  ((_, s1), (_, s2)) => s1.size > s2.size }
   }
 
   def sortAccountByFlaggedSgs[L, R](accountsWithFlaggedSgs: List[(AwsAccount, Either[L, List[R]])]): List[(AwsAccount, Either[L, List[R]])] = {
--- a/hq/app/aws/ec2/EC2.scala
+++ b/hq/app/aws/ec2/EC2.scala
@@ -56,7 +56,7 @@ object EC2 {
       sgResult <- TrustedAdvisorSGOpenPorts.getSGOpenPorts(supportClient)
       sgUsage <- getSgsUsage(sgResult, account)
       flaggedSgs = sgResult.flaggedResources.filter(_.status != "ok")
-    } yield sortSecurityGroupsByInUse(flaggedSgs.map(sgOpenPortsDetail => sgOpenPortsDetail -> sgUsage.getOrElse(sgOpenPortsDetail.id, Set.empty)))
+    } yield sortSecurityGroupsByInUse(flaggedSgs, sgUsage)
   }
 
   def allFlaggedSgs(accounts: List[AwsAccount])(implicit ec: ExecutionContext): Attempt[List[(AwsAccount, Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]])]] = {
@@ -67,8 +67,10 @@ object EC2 {
     }
   }
 
-  private [ec2] def sortSecurityGroupsByInUse(sgsPortFlags : List[(SGOpenPortsDetail, Set[SGInUse])]) = {
-    sgsPortFlags.sortWith{ case  ((_, s1), (_, s2)) => s1.size > s2.size }
+  private [ec2] def sortSecurityGroupsByInUse(sgsPortFlags : List[SGOpenPortsDetail], sgUsage: Map[String, Set[SGInUse]]) = {
+    sgsPortFlags
+      .map(sgOpenPortsDetail => sgOpenPortsDetail -> sgUsage.getOrElse(sgOpenPortsDetail.id, Set.empty))
+      .sortWith{ case  ((_, s1), (_, s2)) => s1.size > s2.size }
   }
 
   def sortAccountByFlaggedSgs[L, R](accountsWithFlaggedSgs: List[(AwsAccount, Either[L, List[R]])]): List[(AwsAccount, Either[L, List[R]])] = {
--- a/hq/app/aws/iam/IAMClient.scala
+++ b/hq/app/aws/iam/IAMClient.scala
@@ -5,22 +5,23 @@ import aws.AwsAsyncHandler._
 import com.amazonaws.regions.{Region, Regions}
 import com.amazonaws.services.identitymanagement.model.{GenerateCredentialReportRequest, GenerateCredentialReportResult, GetCredentialReportRequest}
 import com.amazonaws.services.identitymanagement.{AmazonIdentityManagementAsync, AmazonIdentityManagementAsyncClientBuilder}
-import logic.Retry
-import model.{AwsAccount, IAMCredentialsReport}
+import logic.{ReportDisplay, Retry}
+import model.{AwsAccount, CredentialReportDisplay, IAMCredentialsReport}
 import utils.attempt.Attempt
-import scala.concurrent.duration._
+
 import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
 
 
 object IAMClient {
-  def client(account: AwsAccount, region: Region = Region.getRegion(Regions.EU_WEST_1)): AmazonIdentityManagementAsync = {
+  private def client(account: AwsAccount, region: Region = Region.getRegion(Regions.EU_WEST_1)): AmazonIdentityManagementAsync = {
     val auth = AWS.credentialsProvider(account)
     AmazonIdentityManagementAsyncClientBuilder.standard()
       .withCredentials(auth)
       .withRegion(Option(Regions.getCurrentRegion).getOrElse(region).getName).build()
   }
 
-  def generateCredentialsReport(client: AmazonIdentityManagementAsync)(implicit ec: ExecutionContext): Attempt[GenerateCredentialReportResult] = {
+  private def generateCredentialsReport(client: AmazonIdentityManagementAsync)(implicit ec: ExecutionContext): Attempt[GenerateCredentialReportResult] = {
     val request = new GenerateCredentialReportRequest()
     handleAWSErrs(awsToScala(client.generateCredentialReportAsync)(request))
   }
@@ -30,7 +31,7 @@ object IAMClient {
     handleAWSErrs(awsToScala(client.getCredentialReportAsync)(request)).flatMap(CredentialsReport.extractReport)
   }
 
-  def getCredentialsReport (account: AwsAccount)(implicit ec: ExecutionContext): Attempt[IAMCredentialsReport] = {
+  def getCredentialsReport(account: AwsAccount)(implicit ec: ExecutionContext): Attempt[IAMCredentialsReport] = {
     val delay = 3.seconds
     val client = IAMClient.client(account)
     for {
@@ -39,5 +40,12 @@ object IAMClient {
     } yield report
   }
 
+  def getAllCredentialReports(accounts: Seq[AwsAccount])(implicit executionContext: ExecutionContext): Seq[Attempt[(AwsAccount, CredentialReportDisplay)]] = {
+    accounts map { account =>
+      for {
+        credReport <- getCredentialsReport(account)
+        uiReport = ReportDisplay.toCredentialReportDisplay(credReport)
+      } yield account -> uiReport
+    }
+  }
 }
-
--- a/hq/app/aws/iam/IAMClient.scala
+++ b/hq/app/aws/iam/IAMClient.scala
@@ -7,9 +7,9 @@ import com.amazonaws.services.identitymanagement.model.{GenerateCredentialReport
 import com.amazonaws.services.identitymanagement.{AmazonIdentityManagementAsync, AmazonIdentityManagementAsyncClientBuilder}
 import logic.{ReportDisplay, Retry}
 import model.{AwsAccount, CredentialReportDisplay, IAMCredentialsReport}
-import utils.attempt.Attempt
+import utils.attempt.{Attempt, FailedAttempt}
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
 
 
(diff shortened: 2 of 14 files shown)


hq/app/aws/support/TrustedAdvisor.scala
```diff
@@ -11,10 +11,13 @@ import logic.DateUtils.fromISOString
 import utils.attempt.Attempt
 
 import scala.collection.JavaConverters._
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.ExecutionContext
 
 
 object TrustedAdvisor {
+  val portPriorityMap = Map("27017" -> 0, "27018" -> 0, "27019" -> 0, "6379" -> 0, "6380" -> 0,  "9000" -> 1, "7077" -> 1, "4040" -> 1, "8890" -> 1, "4041" -> 1, "4042" -> 1, "8080" -> 1, "22" -> 9)
```
katebee: `"22" -> 9`

Why is 22 getting assigned 9 here? Should it be 2?

abdusahin: chasing it to 9 in the test

hq/app/controllers/HQController.scala
```diff
@@ -23,8 +24,13 @@ class HQController(val config: Configuration)
...
   }
 
-  def iam = AuthAction {
-    Ok(views.html.iam.iam())
+  def iam = AuthAction.async  {
+    attempt {
+      val accountsAndReports = IAMClient.getAllCredentialReports(accounts)
+      for {
+        accountReportMaps <- Attempt.sequence(accountsAndReports.toList)
+       } yield Ok(views.html.iam.iam(accountReportMaps.toMap))
```
katebee: When running this locally and going to the IAM page, I'm getting an error:

`Unknown error while making an API call to AWS' AmazonIdentityManagement service`

I assume because I have only obtained the creds for a couple of accounts from Janus (and the others have expired). If this is the case, this won't be an issue in PROD (hopefully). However, maybe we could use a similar approach to the security-groups (which still gives you some results) to improve the DEV experience?

katebee: Additionally, would it be worth adding a custom error at this point to help Devs if they run into this?

hq/app/aws/support/TrustedAdvisor.scala
adamnfish: Would it be easier to just use the list index as the priority and make a decision about the relative importances of each port number? That will also help provide a more stable sort across accounts.

adamnfish: Some examples that appear to be missing are:

5432 (postgres)
5601 (kibana)
9300, 9200 (elastic search)
0-65535 (wide open)

Have a look at the existing results to find ports that are are flagged now. If any of those are important ones (especially databases) then make sure this prioritisation captures them.

hq/app/controllers/HQController.scala
adamnfish: It's because Janus credentials can't make IAM calls. It's super annoying, but to run this feature locally you'd need to temporarily make permanent credentials to authenticate those calls.

You're right it's probably worth providing a good error for this that describes the above. That might be a bit fiddly to add (because it's you only get a generic error and this is for a specific problem) but definitely doable.

hq/app/aws/ec2/EC2.scala
```diff
@@ -56,7 +56,7 @@ object EC2 {
       sgResult <- TrustedAdvisorSGOpenPorts.getSGOpenPorts(supportClient)
       sgUsage <- getSgsUsage(sgResult, account)
       flaggedSgs = sgResult.flaggedResources.filter(_.status != "ok")
-    } yield flaggedSgs.map(sgOpenPortsDetail => sgOpenPortsDetail -> sgUsage.getOrElse(sgOpenPortsDetail.id, Set.empty))
+    } yield sortSecurityGroupsByInUse(flaggedSgs.map(sgOpenPortsDetail => sgOpenPortsDetail -> sgUsage.getOrElse(sgOpenPortsDetail.id, Set.empty)))
```
adamnfish: This is for sure too complicated for a one-liner yield, but it's also possibly too much logic for an untested for-comp. Is there any reason the sort function doesn't take `flaggedSgs` so it can do the `map` (and `getOrElse`) step itself as well?

hq/app/aws/support/TrustedAdvisor.scala
```diff
@@ -54,6 +57,36 @@ object TrustedAdvisor {
     handleAWSErrs(awsToScala(client.describeTrustedAdvisorCheckResultAsync)(request))
   }
 
+  private[support] def sortSecurityFlags[A <: TrustedAdvisorCheckDetails](list: List[A]): List[A] = {
+    //    AWS Alert mappings
+    //    Green: Access to port 80, 25, 443, or 465 is unrestricted.
+    //    Red: Access to port 20, 21, 1433, 1434, 3306, 3389, 4333, 5432, or 5500 is unrestricted.
+    //    Yellow: Access to any other port is unrestricted.
+    //
+    //    FTP : 21
```
adamnfish: I think this documentation should be with the port definitions, not here?

hq/app/controllers/HQController.scala
abdusahin: we don't have this issue on PROD, it's happening because of multiple accounts configured in the local `~/.gu/security-hq.local` file. When I leave have only one account like below I managed to see some data. @adamnfish have you got any idea how to fix this one locally ?

```{   
 name = "security"
 id   = "security-test"
 roleArn = "dev-local"
}

File: hq/app/views/iam/iamCredReportBody.scala.html [hunk 5]
```diff
@@ -0,0 +1,49 @@
+@import logic.DateUtils
+@import model.CredentialReportDisplay
+@import logic.ReportDisplay
+@(report: CredentialReportDisplay)
+
+<div class="container">
```
katebee: Suggestion:

For the layout, I would take the greedy-width class off the [main container](https://github.com/guardian/security-hq/pull/17/files#diff-a704b7bf68dfc593b988548cb8d85c3cR16) of the IAM page, and place it on this DIV instead. This will mean the accordion width matches the header, but the tables inside each account can grab some extra width

hq/test/aws/support/TrustedAdvisorSGOpenPortsTest.scala
```diff
@@ -65,4 +68,57 @@ class TrustedAdvisorSGOpenPortsTest extends FreeSpec with Matchers with AttemptV
...
+  "sorting" - {
+
+    "security flags by port priority" in {
+      val portPriorityMap = Map("27017" -> 0, "27018" -> 0, "27019" -> 0, "6379" -> 0, "6380" -> 0,  "9000" -> 1, "7077" -> 1, "4040" -> 1, "8890" -> 1, "4041" -> 1, "4042" -> 1, "8080" -> 1, "22" -> 9)
+      val sgs: Gen[SGOpenPortsDetail] = for {
+        status <- Gen.oneOf("Ok", "Warning", "Error")
+        name <- Gen.alphaStr
+        id <- Gen.alphaStr
+        vpcId <- Gen.alphaStr
+        port <- Gen.oneOf(9000, 8080, 27017, 27018, 27019, 6379, 6380, 22, 4040, 8890, 4042, 4041 )
```
adamnfish: Probably OK to refer to the hard-coded port list here to reduce the burden of maintaining this test.

hq/app/views/fragments/statusSummary.scala.html
```diff
@@ -0,0 +1,14 @@
...
+@(report: CredentialReportDisplay)
+
+@if(ReportDisplay.reportStatusSummary(report).contains(Red)) {
+    <i class="material-icons right red-text" data-position="bottom" data-delay="50">error</i>
+}
+@if(ReportDisplay.reportStatusSummary(report).contains(Amber)) {
+    <i class="material-icons right amber-text" data-position="bottom" data-delay="50">warning</i>
+}
+@if(ReportDisplay.reportStatusSummary(report).contains(Green) && ReportDisplay.reportStatusSummary(report).size == 1) {
+    <i class="material-icons right green-text" data-position="bottom" data-delay="50">check</i>
```
katebee: These have a data-position, but no data-tooltip to set the text... Do some or all of these want a tooltip?

`data-tooltip="Example Label"`

hq/test/aws/support/TrustedAdvisorSGOpenPortsTest.scala
```diff
@@ -65,4 +68,57 @@ class TrustedAdvisorSGOpenPortsTest extends FreeSpec with Matchers with AttemptV
       }
     }
   }
+
+  "sorting" - {
```
adamnfish: It looks like this tests that the three separate sorts all work individually, but not that they are themselves correctly sorted. I realise it's a tricky test to write but it'd be good to know that e.g. unused results always go below used ones.

hq/test/aws/ec2/EC2Test.scala
```diff
@@ -115,6 +116,26 @@ class EC2Test extends FreeSpec with Matchers with Checkers {
...
+        id <- Gen.alphaStr
+        vpcId <- Gen.alphaStr
+        port <- Gen.oneOf(0 to 65000)
+        instanceId <- Gen.alphaStr
+        usage <- Gen.someOf(Seq(Instance(instanceId), UnknownUsage("unknown", "nic-1"), ELB(instanceId)) )
+      } yield SGOpenPortsDetail(status, "eu-west-1", name, id, vpcId, "tcp", port.toString, "Yellow", false ) -> usage.toSet[SGInUse]
+
+      forAll(Gen.listOf(sgsOpenPorts)) { detail =>
+        val sortedResult = EC2.sortSecurityGroupsByInUse(detail)
+        sortedResult should be (sortedResult.sortWith{ case  ((_, s1), (_, s2)) => s1.size > s2.size })
```
adamnfish: Does this test offer any value given that it is literally a copy paste of the implementation?

hq/app/views/fragments/statusSummary.scala.html
```diff
@@ -0,0 +1,14 @@
+@import model.CredentialReportDisplay
+@import logic.ReportDisplay
+@import model._
+@(report: CredentialReportDisplay)
+
+@if(ReportDisplay.reportStatusSummary(report).contains(Red)) {
+    <i class="material-icons right red-text" data-position="bottom" data-delay="50">error</i>
+}
+@if(ReportDisplay.reportStatusSummary(report).contains(Amber)) {
```
adamnfish: Make sure the design supports showing multiple reports if Red and Amber both exist.

hq/app/aws/iam/IAMClient.scala
```diff
@@ -39,5 +40,12 @@ object IAMClient {
     } yield report
   }
 
-}
-
+  def getAllCredentialReports(accounts: Seq[AwsAccount])(implicit executionContext: ExecutionContext): Seq[Attempt[(AwsAccount, CredentialReportDisplay)]] = {
```
adamnfish: As discussed IRL, this should probably have a different signature. Off the top of my head something like the following, but compare it to the Security Groups implementation.

`Attempt[List[(AwsAccount, Either[FailedAttempt, CredentialReportDisplay])]]`

hq/app/aws/support/TrustedAdvisor.scala
abdusahin: didn't add postgress, mysql, etc, as they're already flagged as Red. 
Wide open is also flagged as Red.

adamnfish: Sure, but since we're already doing a sort we could make it stable (consistent across changing data and different accounts) by implementing a sort that's 'stable' across the various ports.

hq/app/aws/ec2/EC2.scala
abdusahin: yeah, can deal with that.

hq/app/aws/support/TrustedAdvisor.scala
abdusahin: I guess I can add that info to ports and propagate it back to the UI for display

hq/app/aws/support/TrustedAdvisor.scala
abdusahin: not sure about deriving priority from the index. More importantly, I think wide port security group rule will break it.

hq/app/views/fragments/statusSummary.scala.html
abdusahin: I'm removing `data-position`

hq/app/aws/iam/IAMClient.scala
abdusahin: I'll use `Attempt.successfulAttempts` which is accepting a parameter of `attempts: List[Attempt[A]]` so will ignore the failing attempts.

hq/app/controllers/HQController.scala
```diff
@@ -23,8 +24,13 @@ class HQController(val config: Configuration)
     Ok(views.html.index(accounts))
   }
 
-  def iam = AuthAction {
-    Ok(views.html.iam.iam())
+  def iam = AuthAction.async  {
+    attempt {
+      val accountsAndReports = IAMClient.getAllCredentialReports(accounts)
+      for {
+        accountReportMaps <- Attempt.sequence(accountsAndReports.toList)
```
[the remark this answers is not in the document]
abdusahin: i think `Attempt.successfulAttempts` will solve this issue.

hq/app/views/fragments/statusSummary.scala.html
abdusahin: it does.

adamnfish:
Also, I think this still needs to be properly linked up (take away the cred-report route and use the IAM page for this report)

hq/app/views/fragments/statusSummary.scala.html
```diff
@@ -0,0 +1,14 @@
...
+    <i class="material-icons right amber-text" data-position="bottom" data-delay="50">warning</i>
+}
+@if(ReportDisplay.reportStatusSummary(report).contains(Green) && ReportDisplay.reportStatusSummary(report).size == 1) {
```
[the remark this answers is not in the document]
abdusahin: changed to `ReportDisplay.reportStatusSummary(report).forall(_ == Green)`

hq/app/views/iam/iam.scala.html
```diff
@@ -13,7 +13,19 @@ <h4 class="grey-text text-darken-4 center-on-small-only">For all accounts</h4>
     </div>
 
 } { @* Main content *@
-    <div class="container">
-        <p>Fetch IAM summary for entire AWS estate.</p>
+    <div class="container greedy-width">
```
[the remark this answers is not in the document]
abdusahin: removed as per Kate's suggestion

hq/test/aws/ec2/EC2Test.scala
```diff
@@ -115,6 +116,26 @@ class EC2Test extends FreeSpec with Matchers with Checkers {
...
   }
 
+  "security group" - {
+    "sort by usage" in {
+      val sgsOpenPorts = for {
+        status <- Gen.oneOf("Ok", "Warning", "Error")
+        name <- Gen.alphaStr
+        id <- Gen.alphaStr
+        vpcId <- Gen.alphaStr
+        port <- Gen.oneOf(0 to 65000)
```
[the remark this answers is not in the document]
abdusahin: port range is a bit tricky . I can change it so I can make it high priority if port range is detected.

hq/test/aws/ec2/EC2Test.scala
abdusahin: yes, it checks sort is stable. 
e.g. if two items has same priority their place shouldn't be changed by any sort.

hq/test/aws/support/TrustedAdvisorSGOpenPortsTest.scala
abdusahin: ok, i'll try to add one more test case with used and unused security groups.

hq/test/aws/ec2/EC2Test.scala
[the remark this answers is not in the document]
adamnfish: We don't want to make it high priority for *any* range. Definitely needs to explicitly flag wide open ports though. I think `9200-9300` isn't unheard of for Elastic Search, ideally the implementation would catch that via the `9200` and `9300` entries.

hq/test/aws/ec2/EC2Test.scala
adamnfish: But I'm talking about stable between accounts which is a different thing, not stable based on whatever order AWS provides them. Are MongoDB SGs always above Postgres, for example. It'll make it easier for InfoSec to review accoutns if they are always displayed the same way. This isn't a hard requirement now but I think it'd be useful. It'll be easier to implement now then change it later! If you don't agree this is important then no worries (or we can chat IRL, whatever you like).

abdusahin: i think that's doable, provided we put MongoDB before Postgres in the priority list. 
For that to work, I guess I can use an index based priority map so that each entry would have a unique position.

hq/app/views/iam/iam.scala.html
```diff
@@ -13,7 +13,19 @@ <h4 class="grey-text text-darken-4 center-on-small-only">For all accounts</h4>
...
+        <ul class="collapsible" data-collapsible="accordion">
+        @for((account, report) <- accounts) {
```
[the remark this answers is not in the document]
abdusahin: ok.

hq/test/aws/ec2/EC2Test.scala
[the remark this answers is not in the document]
abdusahin: i guess I can check if a port range is falling into the port priority list.

hq/app/views/iam/iamCredReportBody.scala.html
abdusahin: done

hq/app/aws/iam/IAMClient.scala
abdusahin: done.

abdusahin:
@adamnfish , @katebee , can you have a  look  again.

adamnfish:
:+1:

PR #18, opened 2017-11-24
Make resource details display correctly on Security Groups page

katebee:
##### Adds numbers and icons to show the number of resources hidden by the dropdown list:

picture 77

##### Makes the multiple resources dropdown the same height as the usage cta, and gives a bit more accommodation for moderately long resource names: 

picture 78

😎 🤘


hq/public/stylesheets/main.css
```diff
@@ -173,12 +177,31 @@ nav {
...
+    margin-right: 5px;
+    text-align: end;
+    width: 30px;
 }
 
 .usage-cta {
     display: block;
     background-color: #4E8098;
     padding: 0 .6rem;
+    white-space: nowrap;
```
katebee: This prevents long resource names from breaking onto a newline if they contain a hyphen.

hq/app/logic/SecurityGroupDisplay.scala
```diff
@@ -0,0 +1,23 @@
...
+import model.{ELB, Instance, SGInUse}
+
+
+object SecurityGroupDisplay {
+
+  case class ResourceIcons(instances: Int, elbs: Int, unknown: Int)
+
+  def resourceIcons(usages: List[SGInUse]): ResourceIcons = {
+
+    val instances = usages.collect {
```
abdusahin: You can traverse the `usages` list at once while collecting numbers for both Instance and ELB. It's not a big deal
but something like below would work.
```
val (ins, elb) = usages.collect {
    case Instance(_) => (1,0)
    case ELB(_) => (0, 1)
}.unzip
val instances = elb.sum
val elbs = elb.sum
```

hq/test/logic/SecurityGroupDisplayTest.scala
```diff
@@ -0,0 +1,22 @@
+package logic
+
+import model.{ELB, Instance, SGInUse, UnknownUsage}
+import org.scalatest.{FreeSpec, Matchers}
+import SecurityGroupDisplay._
+
+
+class SecurityGroupDisplayTest extends FreeSpec with Matchers {
+  "resourceIcons" - {
+    "returns 2 when there are two instances" in {
```
abdusahin: feels like this case isn't needed as the below one covers it.

hq/app/logic/SecurityGroupDisplay.scala
abdusahin: well there is also another method i can suggest so that there is no need to call `sum` later on.

```
val (instances, elbs) = usages.foldLeft(0,0) {
  case ( (ins, elb), Instance(_)) => (ins+1, elb)
  case ( (ins, elb), ELB(_)) => (ins, elb+1)
}
```

abdusahin:
👍

hq/app/logic/SecurityGroupDisplay.scala
katebee: Grand, planning to change this to:

```
    val (instances, elbs, unknown) = usages.foldLeft(0,0,0) {
      case ( (ins, elb, unk), Instance(_) ) => (ins+1, elb, unk)
      case ( (ins, elb, unk), ELB(_) ) => (ins, elb+1, unk)
      case ( (ins, elb, unk), _ ) => (ins, elb, unk+1)
    }
```

hq/test/logic/SecurityGroupDisplayTest.scala
katebee: Good point, I might swap it out for a test for when the list is empty

PR #20, opened 2017-11-27
Removed sbt step from bundle script

abdusahin:


adamnfish:
This doesn't achieve anything by itself so let's hold off on merging this until its value as clear as part of a solution to sorting out the build.

PR #21, opened 2017-11-29
Upgrade to Play 2.6.7

abdusahin:
- Play upgrade to 2.6.7
- Assets changes

### DIFF
--- a/hq/conf/application.conf
+++ b/hq/conf/application.conf
@@ -20,8 +20,7 @@
 # ~~~~~
 # The secret key is used to sign Play's session cookie.
 # This must be changed for production, but we don't recommend you change it in this file.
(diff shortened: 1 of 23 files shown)


hq/conf/application.conf
```diff
@@ -20,8 +20,9 @@
 # ~~~~~
 # The secret key is used to sign Play's session cookie.
 # This must be changed for production, but we don't recommend you change it in this file.
-play.crypto.secret = ${APPLICATION_SECRET}
-application.secret = ${APPLICATION_SECRET}
+play.http.secret.key ="skkqY6yoYpT:tk_lS`=/]El/KhI[VlE<;FPRrp^v[j_b^]rbEl@68`/E<^MT@xwo"
```
katebee: ? Is this supposed to be here?

abdusahin: i didn't realise we had a `APPLICATION_SECRET` key in the local `~/.gu/security-hq.local.conf`. Removing the default one.

PR #22, opened 2017-12-04
Yarn install all the things! Add Prettier, and let it fix everything 💅 🌈

katebee:
This is how it begins...

- New package.json to manage our node modules and scripts
- Make sass-lint consistent with the Prettier ruleset
- Run Prettier against the CSS and JS to fix and format it
- Update the README so that getting started with client-side development makes sense

Validation and fixing of the CSS/JS is not automated yet... but hopefully Soon™


hq/README.md
```diff
@@ -80,3 +80,34 @@ It is possible to run the application and setup nginx as a proxy; this will make
...
+  >
+  > *Mac OSX:* `$ brew install yarn`
+
+2. Install all the dependencies:
+
+`$ yarn`
+
+3. Run the setup for the dependencies:
+
+`$ yarn run setup`
```
abdusahin: that bit didn't work for me.  Am i missing something ?

abdusahin:
👍

hq/README.md
katebee: Nope, I deleted that script but forgot to remove the mention in the readme 😅 

Will fix it now, cheers.

PR #23, opened 2017-12-04
Make credential username hyperlink to AWS IAM console

abdusahin:


hq/app/views/iam/iamCredReportBody.scala.html
```diff
@@ -20,8 +20,9 @@
                     <th>Status</th>
                 </tr>
                 @for(cred <- report.humanUsers) {
+
                     <tr>
-                        <td><span>@views.html.fragments.username(cred.username)</span></td>
+                        <td><a target="_blank" href="https://console.aws.amazon.com/iam/home#/users/@{cred.username}">@views.html.fragments.username(cred.username)</a></td>
```
katebee: We probably can trust AWS, but if we want to be extra careful we could add `rel="noopener noreferrer"` to the anchor tag...

Fun fact:
> The page we're linking to gains partial access to the linking page via the window.opener object.

abdusahin: very interesting, 👐   web world is always complex.

katebee: 🤔 Do we want to target blank and change the default behaviour of links? (even speaking as someone who cmd-clicks every link because I love tabs)

If we are going to stick with target blank, maybe we could make the link an icon instead? Other apps use this icon to show that it opens in a new tab:

picture 83

https://material.io/icons/#ic_open_in_new

PR #24, opened 2017-12-04
Made security groups id a hyperlink to AWS security groups console

abdusahin:


abdusahin:
@katebee 
added an icon next to the hyperlink

PR #25, opened 2017-12-04
Switch to CSS grid, and add icons to expandables

katebee:
### UX: icons for expandable menus

We need more visual cues that expandable menus are clickable, so first step is to add an icon:

picture 84

### CSS: Grid layout for accounts list

- Less flex, more CSS grid
- Less CSS overall 🎉 🔥
- Deleting some html since the elements are no longer needed


abdusahin:
looks much better 👍

abdusahin:
deployed 🛫

PR #27, opened 2017-12-06
Provide upstart

aug24:

PR #28, opened 2017-12-06
Materialize update, working expandables and other frontend enhancements

katebee:
#### Updating the version of Materialize! 🚀 

- Fixes some bugs and missing functionalities
- Materialize is now included as a node module, as an explicit version
- Only include the minimised versions in the `/dist`... delete the others. The pretty versions still are available in `/node-modules` if needed
- Adds the necessary hacks to the HTML and CSS, to offset the layout changes between materialize versions (yay... more !important)
- Fixes some CSS discrepancies between Firefox and Chrome 😬 🔫 

#### New and improved JavaScript! 🌟 

- Fixes the bug where hovering and clicking on the security groups expandable would result in an annoying behaviour
- Clicking will now remove the hover effect, and swap to toggling the open and close by clicking

#### Moves IAM page errors into the expandable bodies! 💅

- Consistent with the Security Groups page
- More room for error messages

picture 85

#### Bonus floating menu for all your navigation needs!

The page with all the security groups now contains menu that becomes visible when you expand one of the accounts:

picture 86

Through the power of JS, it can scroll you to the top or end of the account, or close the active expandable. This might be helpful if an account has more than a dozen security groups...

...It also looks pretty


katebee:
These changes have been merged into branch https://github.com/guardian/security-hq/pull/27, and will be getting to master that way 😁 

🚢 🚢

PR #30, opened 2017-12-11
Pull VPC details of the Security Groups

abdusahin:
- Pull name of the VPC linked to the security group
- Open resources in new tab.

### DIFF
--- a/hq/README.md
+++ b/hq/README.md
@@ -57,7 +57,7 @@ AWS_ACCOUNTS=[
 ```
 
 ### AWS Security Policies
-See ```watched-account``` cloud formation template for the security policies needed to run security-hq.
+See `watched-account` template under `cloudformation` folder for the security policies needed to run security-hq.
 
 ### Running project
 From the root of the project:
--- a/hq/README.md
+++ b/hq/README.md
@@ -56,6 +56,18 @@ AWS_ACCOUNTS=[
 ]
 ```
 
+### Required AWS Security Policies
+The following security policies are needed for security-hq to run.
+```
+    "trustedadvisor:Describe*",
+    "support:*",
+    "ec2:DescribeNetworkInterfaces",
+    "ec2:DescribeVpcs",
+    "iam:GenerateCredentialReport",
+    "iam:GetCredentialReport"
+
+```
+
 ### Running project
 From the root of the project:
 
--- a/hq/README.md
+++ b/hq/README.md
@@ -56,17 +56,8 @@ AWS_ACCOUNTS=[
 ]
 ```
 
-### Required AWS Security Policies
-The following security policies are needed for security-hq to run.
-```
-    "trustedadvisor:Describe*",
-    "support:*",
-    "ec2:DescribeNetworkInterfaces",
-    "ec2:DescribeVpcs",
-    "iam:GenerateCredentialReport",
-    "iam:GetCredentialReport"
-
-```
+### AWS Security Policies
+See ```watched-account``` cloud formation template for the security policies needed to run security-hq.
 
 ### Running project
 From the root of the project:
--- a/hq/app/aws/ec2/EC2.scala
+++ b/hq/app/aws/ec2/EC2.scala
@@ -8,7 +8,7 @@ import cats.instances.set._
 import cats.syntax.semigroup._
 import com.amazonaws.auth.AWSCredentialsProviderChain
 import com.amazonaws.regions.Regions
-import com.amazonaws.services.ec2.model.{Instance => _, _}
+import com.amazonaws.services.ec2.model.{DescribeVpcsRequest, Vpc, DescribeNetworkInterfacesRequest, DescribeNetworkInterfacesResult, Filter, NetworkInterface}
 import com.amazonaws.services.ec2.{AmazonEC2Async, AmazonEC2AsyncClientBuilder}
 import model._
 import utils.attempt.{Attempt, FailedAttempt}
@@ -114,7 +114,7 @@ object EC2 {
   private[ec2] def parseNetworkInterface(ni: NetworkInterface): SGInUse = {
     val elb = Option(ni.getAttachment.getInstanceOwnerId).find(_ == "amazon-elb")
       .map(_ => ELB(ni.getDescription.stripPrefix("ELB ")))
-    val instance = Option(ni.getAttachment.getInstanceId).map(Instance)
+    val instance = Option(ni.getAttachment.getInstanceId).map(Ec2Instance)
 
     elb
       .orElse(instance)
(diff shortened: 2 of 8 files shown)


abdusahin:
@adamnfish , @katebee  can you have a look please ?

hq/app/views/fragments/openSecurityGroups.scala.html
```diff
@@ -59,8 +61,8 @@
 
                         </tr>
                         <tr>
-                            <th>VPC</th>
-                            <td>@flaggedSg.vpcId</td>
+                            <th>VPC Id/Name</th>
+                            <td>@flaggedSg.vpcId  / @flaggedSg.vpcName.getOrElse("N/A")</td>
```
katebee: Could we make the name a new row in the table, so we have separate rows for the VPC ID and VPC name?

This will give us less layout issues, and make the table content a bit clearer to new users.

hq/app/aws/ec2/EC2.scala
```diff
@@ -120,4 +122,26 @@ object EC2 {
         UnknownUsage(ni.getDescription, ni.getNetworkInterfaceId)
       )
   }
+
+  private[ec2] def addVpcName (flaggedSgs : List[SGOpenPortsDetail], vpcs: Map[String, Vpc]): List[SGOpenPortsDetail] = {
```
adamnfish: Try and format stuff correctly!

```scala
private[ec2] def addVpcName(flaggedSgs: List[SGOpenPortsDetail], vpcs: Map[String, Vpc]):
```

Elsewhere as well, I'll not call it out in every case.

hq/app/aws/ec2/EC2.scala
```diff
@@ -120,4 +122,26 @@ object EC2 {
...
+    Attempt.traverse(flaggedSgs.map(_.region).distinct) { region =>
+      val ec2Client = client(account, Regions.fromName(region) )
+      val vpcIds = flaggedSgs.filter(_.region == region).map(_.vpcId)
+      vpcsDetailsF(ec2Client, vpcIds)
+    }.map( _.fold(Map.empty)(_ ++ _) )
+  }
+
+  private def getVpcsDetails (client: AmazonEC2Async, vpcIds: List[String])(implicit ec: ExecutionContext): Attempt[Map[String, Vpc]] = {
+    val request = new DescribeVpcsRequest().withVpcIds(vpcIds.asJava)
+    handleAWSErrs(awsToScala(client.describeVpcsAsync)(request)).map( _.getVpcs.asScala.map(vpc => vpc.getVpcId -> vpc).toMap)
```
adamnfish: This looks a little complex for a single line, is it possible to break it up slightly?

hq/app/views/fragments/openSecurityGroups.scala.html
```diff
@@ -8,14 +8,16 @@
 @resourceList(flaggedSg: SGOpenPortsDetail, usages: Set[SGInUse]) = {
     @usages.map {
         case Instance(id) => {
-              <a class="btn usage-cta" href="https://@{flaggedSg.region}.console.aws.amazon.com/ec2/v2/home?region=@{flaggedSg.region}#Instances:search=@id">
+              <a target="_blank" rel="noopener noreferrer" class="btn usage-cta" href="https://@{flaggedSg.region}.console.aws.amazon.com/ec2/v2/home?region=@{flaggedSg.region}#Instances:search=@id">
```
adamnfish: I'm not sure we need blank targets, our users know how to use web browsers (and can choose to open in a new tab if they want to)?

hq/test/aws/ec2/EC2Test.scala
```diff
@@ -140,6 +142,38 @@ class EC2Test extends FreeSpec with Matchers with Checkers with PropertyChecks
     }
   }
 
+  "VPC" - {
```
adamnfish: The formatting's a bit off in the block. Pay attention to unneeded empty lines, indentation and whitespace in signatures.

hq/test/aws/ec2/EC2Test.scala
```diff
@@ -140,6 +142,38 @@ class EC2Test extends FreeSpec with Matchers with Checkers with PropertyChecks
...
+    val sgs3 = SGOpenPortsDetail("Ok", "eu-west-2", "name-3", "id-122", "vpc-3", "tcp", "1099", "Yellow", false )
+
+    val vpcsMap = Map(
+      "vpc-1" -> new Vpc().withVpcId("vpc-1").withTags(new Tag("Name", "name-1")),
+      "vpc-2" -> new Vpc().withVpcId("vpc-2").withTags(new Tag("Name", "name-2")),
+      "vpc-3" -> new Vpc().withVpcId("vpc-3")
+    )
+    val sgsList = List(sgs1, sgs2, sgs3)
+
+    "get vpcs"  in {
```
adamnfish: When you're writing `FreeSpec` tests the test message is the concatenation of all the blocks that lead to the test. In this case that is `"VPC"` and `"get vpcs"`, resulting in a test message of `"VPC get vpcs"`. This is not a helpful test message!

I generally create blocks to contain tests grouped by functionality. This means the test setup can be shared as well as giving the file logical grouping based on the scope. Within each method it can be helpful to group tests there as well for the same reasons.

None of that is set in stone, `FreeSpec` is deliberately flexible! However, whatever test runner you use you do need to make sure the test describes what you expect to happen, so that when it fails the developer has some context about what they'll need to fix.

An example from elsewhere in this file is `parseDescribeNetworkInterfacesResults returns nothing if no interfaces match search IDs` which describes what we're after (and shares some test setup in the scope as I've mentioned above).

hq/test/aws/ec2/EC2Test.scala
```diff
@@ -140,6 +142,38 @@ class EC2Test extends FreeSpec with Matchers with Checkers with PropertyChecks
...
+
+      def vpcsF (ec2: AmazonEC2Async, vpcIds : List[String]) : Attempt[Map[String, Vpc]] = {
```
adamnfish: Since this is only used once would it be easier(/clearer) to just hard-code the result? I'm not sure - could go either way on it! (I'm just always nervous when tests look like they need tests)

hq/README.md
```diff
@@ -56,6 +56,18 @@ AWS_ACCOUNTS=[
 ]
 ```
 
+### Required AWS Security Policies
```
adamnfish: While this is a very useful addition, would it not be better to point the developer at the actual policy (in cloudformation)? As it is, we now have two places that need to be kept up to date.

Speaking of, this permissions change needs to be made to the CloudFormation.
https://github.com/guardian/security-hq/blob/master/cloudformation/watched-account.template.yaml

adamnfish: (This one's a blocker because Security-HQ won't work if these permissions aren't set)

hq/app/aws/ec2/EC2.scala
```diff
@@ -56,7 +56,9 @@ object EC2 {
       sgResult <- TrustedAdvisorSGOpenPorts.getSGOpenPorts(supportClient)
       sgUsage <- getSgsUsage(sgResult, account)
       flaggedSgs = sgResult.flaggedResources.filter(_.status != "ok")
-    } yield sortSecurityGroupsByInUse(flaggedSgs, sgUsage)
+      vpcs <- getVpcs(account, flaggedSgs)(getVpcsDetails)
```
adamnfish: It looks to me like if we fail to fetch the VPC information this whole feature will fail. While that may be what we want, it's also possible that since this is just "extra" information we'll be ok to log the errors and carry on?

Have a think about whether you really mean for this to be required. If not, you could use something like `Attempt.successfulAttempts` to discard the failures (obviously making sure they're at least logged and we don't discard them completely).

On the other hand, that would raise the question of "what do we do about the failure" and "who's ever going to look at the logs", so I'm happy if the answer is "it's fine like this".

hq/app/aws/ec2/EC2.scala
```diff
@@ -8,7 +8,7 @@ import cats.instances.set._
 import cats.syntax.semigroup._
 import com.amazonaws.auth.AWSCredentialsProviderChain
 import com.amazonaws.regions.Regions
-import com.amazonaws.services.ec2.model.{DescribeNetworkInterfacesRequest, DescribeNetworkInterfacesResult, Filter, NetworkInterface}
+import com.amazonaws.services.ec2.model.{Instance => _, _}
```
adamnfish: This `Instance => _` import is a bad sign. Might mean we need to rethink our own model name because this clash is something that could easily cause lots of confusion.

hq/README.md
abdusahin: will update one, PR left old after moving to CF template.

hq/app/views/fragments/openSecurityGroups.scala.html
abdusahin: when me & Kate visited teams. we noticed most people didn't do `ctrl/cmd + click` and even asked for opening in a new tab.

hq/app/aws/ec2/EC2.scala
abdusahin: I cannot see how retrieving vpc details could fail, except a technical failure, which is a different matter. VPC details are retrieved like any other AWS api call. I think it feels like too cautious to add this safety guard.

hq/app/views/fragments/openSecurityGroups.scala.html
katebee: Yep - several people requested new tab default after clicking through to AWS, then realising that they had lost the tab for Security HQ.

Not a huge fan of target blank, but seems here it solves a fairly common UX issue. However, I would like to keep a watch on further usage!

abdusahin:
@adamnfish  is this good to go ?

hq/README.md
```diff
@@ -56,6 +56,9 @@ AWS_ACCOUNTS=[
 ]
 ```
 
+### AWS Security Policies
+See ```watched-account``` cloud formation template for the security policies needed to run security-hq.
```
katebee: Minor (don't let me prevent merge): 
- those triple quotes could be single ones
- we have been treating cloudformation as one word 😄

hq/app/aws/ec2/EC2.scala
adamnfish: OK!

hq/app/views/fragments/openSecurityGroups.scala.html
adamnfish: :eyes: :+1:

PR #31, opened 2017-12-11
Refresh security groups open ports report

abdusahin:
- Try to refresh security groups open ports report before displaying.

### DIFF
--- a/cloudformation/watched-account.template.yaml
+++ b/cloudformation/watched-account.template.yaml
@@ -32,6 +32,7 @@ Resources:
             Action:
             # Analyse security groups
             - trustedadvisor:Describe*
+            - trustedadvisor:Refresh*
             - support:*
             - ec2:DescribeNetworkInterfaces
             - ec2:DescribeVpcs
--- a/hq/app/aws/ec2/EC2.scala
+++ b/hq/app/aws/ec2/EC2.scala
@@ -2,6 +2,7 @@ package aws.ec2
 
 import aws.AWS
 import aws.AwsAsyncHandler.{awsToScala, handleAWSErrs}
+import aws.support.TrustedAdvisor.refreshTrustedAdvisorCheck
 import aws.support.{TrustedAdvisor, TrustedAdvisorSGOpenPorts}
 import cats.instances.map._
 import cats.instances.set._
@@ -10,6 +11,7 @@ import com.amazonaws.auth.AWSCredentialsProviderChain
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.ec2.model.{DescribeVpcsRequest, Vpc, DescribeNetworkInterfacesRequest, DescribeNetworkInterfacesResult, Filter, NetworkInterface}
 import com.amazonaws.services.ec2.{AmazonEC2Async, AmazonEC2AsyncClientBuilder}
+import com.amazonaws.services.support.model.RefreshTrustedAdvisorCheckResult
 import model._
 import utils.attempt.{Attempt, FailedAttempt}
 
@@ -50,10 +52,9 @@ object EC2 {
     }
   }
 
-  def flaggedSgsForAccount(account: AwsAccount, refresh: Boolean = true)(implicit ec: ExecutionContext): Attempt[List[(SGOpenPortsDetail, Set[SGInUse])]] = {
+  def flaggedSgsForAccount(account: AwsAccount)(implicit ec: ExecutionContext): Attempt[List[(SGOpenPortsDetail, Set[SGInUse])]] = {
     val supportClient = TrustedAdvisor.client(account)
     for {
-      _  <- if (refresh) TrustedAdvisorSGOpenPorts.refreshSGOpenPorts(supportClient) else Attempt.Right(Unit)
       sgResult <- TrustedAdvisorSGOpenPorts.getSGOpenPorts(supportClient)
       sgUsage <- getSgsUsage(sgResult, account)
       flaggedSgs = sgResult.flaggedResources.filter(_.status != "ok")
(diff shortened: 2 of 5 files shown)


abdusahin:
@adamnfish , @katebee can u have a look pls ?

hq/app/aws/ec2/EC2.scala
```diff
@@ -50,19 +50,27 @@ object EC2 {
     }
   }
 
-  def flaggedSgsForAccount(account: AwsAccount)(implicit ec: ExecutionContext): Attempt[List[(SGOpenPortsDetail, Set[SGInUse])]] = {
+  def flaggedSgsForAccount(account: AwsAccount, refresh: Boolean = true)(implicit ec: ExecutionContext): Attempt[List[(SGOpenPortsDetail, Set[SGInUse])]] = {
     val supportClient = TrustedAdvisor.client(account)
     for {
+      _  <- if (refresh) TrustedAdvisorSGOpenPorts.refreshSGOpenPorts(supportClient) else Attempt.Right(Unit)
```
adamnfish: Could the refresh be done from the call site, rather than here? This mixes the concerns of refreshing the report with fetching the flagged SGs. The fact that it takes `refresh` as a parameter suggests you think it's the call-site that should decide whether to refresh or not. In that case, is it true that having the caller refresh in the for comp of "things that need to be done" would be better than hiding it in this function?

adamnfish: Another reason for separating them is that I think we will probably move the refresh commands into a service that runs separately from user-generated web requests. While doing speculative work is always bad, the fact that it also simplifies this code means that might lend some weight to exposing the refresh as a single `Attempt` call separate to fetching the SGs for an account(s).

hq/app/aws/support/TrustedAdvisorSGOpenPorts.scala
```diff
@@ -60,4 +61,10 @@ object TrustedAdvisorSGOpenPorts {
         }
     }
   }
+
+  def refreshSGOpenPorts(client: AWSSupportAsync)(implicit ec: ExecutionContext): Attempt[RefreshTrustedAdvisorCheckResult]  = {
```
adamnfish: Pay attention to the formatting in this function (whitespace).

hq/app/aws/ec2/EC2.scala
abdusahin: did changes to refresh from the controller

abdusahin:
@adamnfish are you with the changes. ?

hq/app/aws/support/TrustedAdvisor.scala
```diff
@@ -120,4 +120,10 @@ object TrustedAdvisor {
       resourcesSuppressed = result.getResult.getResourcesSummary.getResourcesSuppressed
     )
   }
+
+  def refreshTrustedAdvisorCheck(client: AWSSupportAsync, checkId: String)(implicit ec: ExecutionContext): Attempt[RefreshTrustedAdvisorCheckResult] = {
```
adamnfish: There's a `getTrustedAdvisorChecks` function higher up in this Object. This function feels like the counterpart to that. As a result, I think this should be named with a plural and it should be placed alongside that function.

hq/app/aws/support/TrustedAdvisorSGOpenPorts.scala
```diff
@@ -60,4 +61,10 @@ object TrustedAdvisorSGOpenPorts {
         }
     }
   }
+
+  def refreshSGOpenPorts(client: AWSSupportAsync)(implicit ec: ExecutionContext): Attempt[RefreshTrustedAdvisorCheckResult] = {
```
adamnfish: If refresh is required here why is it not required for the other checks?

hq/app/aws/ec2/EC2.scala
```diff
@@ -61,12 +63,19 @@ object EC2 {
     } yield sortSecurityGroupsByInUse(flaggedSgsWithVpc, sgUsage)
   }
 
+  def refreshSGSReports(accounts: List[AwsAccount])(implicit ec: ExecutionContext): Attempt[List[Either[FailedAttempt, RefreshTrustedAdvisorCheckResult]]] = {
+    Attempt.traverseWithFailures(accounts) { account =>
```
adamnfish: Is there any value keeping all the failures if they're only going to be thrown away?

See [Line 30](https://github.com/guardian/security-hq/pull/31/files#diff-59b0ca98fa263a392fac7315f4bf946bR30) and [Line 41](https://github.com/guardian/security-hq/pull/31/files#diff-59b0ca98fa263a392fac7315f4bf946bR41) of the Controller.

It looks like this functionality would fail silently - there isn't even any logging I can see to let us know if there are problems with the refreshing.

cloudformation/watched-account.template.yaml
```diff
@@ -32,6 +32,7 @@ Resources:
             Action:
             # Analyse security groups
             - trustedadvisor:Describe*
+            - trustedadvisor:Refresh*
```
adamnfish: Good catch!

hq/app/aws/ec2/EC2.scala
katebee: I'm getting a `Unknown error while making an API call to AWS' AmazonEC2 service` on some areas of the dashboard

Whereas the application is logging:
```
Failed to execute AWS SDK operation
com.amazonaws.services.ec2.model.AmazonEC2Exception: You are not authorized to perform this operation
```

Might be helpful to have a more detailed error o n the dashboard?

hq/app/aws/ec2/EC2.scala
```diff
@@ -2,6 +2,7 @@ package aws.ec2
 
 import aws.AWS
 import aws.AwsAsyncHandler.{awsToScala, handleAWSErrs}
+import aws.support.TrustedAdvisor.refreshTrustedAdvisorCheck
```
katebee: 🤔 this is imported but I don't think it is used?

hq/app/aws/support/TrustedAdvisorSGOpenPorts.scala
abdusahin: I think it's actually required for other checks too. There is a refresh button next to each check on AWS Trusted Advisor -> Security page, so I imagine other checks needs refreshing too.

hq/app/aws/ec2/EC2.scala
abdusahin: will remove that one.

hq/app/aws/ec2/EC2.scala
abdusahin: @adamnfish regarding failing silently, I can make changes so that it won't retrieve SGS for the failing accounts, or simply can log the accounts failing refresh.

shtukas:
I am closing this as it has been redone here: https://github.com/guardian/security-hq/pull/60

PR #33, opened 2017-12-13
Header overhaul, including new sub-nav and links between account pages

katebee:
## 👉 Changes to the header - new Nav and Sub nav

- make the global nav more distinct from the account nav
- Now possible to jump between an account SG page, and the IAM report page
- Also applies to the global pages

picture 92

Responsive:

picture 102

## 👉 Other style fixes and improvements

- Black colour on links
- Main icon resizing at small breakpoint
- Margins and spacing
- CSS naming conventions
- Correct port number in the README
- Adjusts breakpoints of grid pattern on Security Groups page

picture 94


hq/app/views/iam/iam.scala.html
```diff
@@ -3,11 +3,16 @@
...
+    <div class="hq-sub-header">
+        <div class="container hq-sub-header__row">
+            <div class="hq-sub-header__name">
+                <h4 class="header light grey-text text-lighten-5">All accounts</h4>
+            </div>
+            <div class="hq-sub-header__tabs">
+                <ul class="tabs tabs-transparent">
+                    <li class="tab col s3"><a target="_self" href="/security-groups"><i class="material-icons left">vpn_lock</i>Security Groups</a></li>
+                    <li class="tab col s3"><a target="_self" class="active" href="/iam"><i class="material-icons left">vpn_key</i>IAM Report</a></li>
+                </ul>
```
katebee: This code is duplicated with almost no variation in a few places. It would be good to split the sub-navigation out into a template fragment, so it can be reused rather than copy-pasted.

Will leave this for a future PR 😁 😉

hq/public/stylesheets/main.css
```diff
@@ -142,13 +207,13 @@ nav {
...
 
-@media (min-width: 740px) {
+@media (min-width: 940px) {
   .sg-container {
     grid-template-columns: repeat(2, 1fr);
   }
 }
 
-@media (min-width: 1140px) {
+@media (min-width: 1400px) {
```
katebee: Was running the branch with the VPC and cloudformation stack details and noticed that the SG cards now need a bit more horizontal room, so have given them some breathing space as part of this PR.

PR #34, opened 2017-12-18
Display Cloudformation  stack id & name

abdusahin:
Display Cloudformation  stack id & name on security groups page.

### DIFF
--- a/hq/app/aws/ec2/EC2.scala
+++ b/hq/app/aws/ec2/EC2.scala
@@ -55,11 +55,15 @@ object EC2 {
     for {
       sgResult <- TrustedAdvisorSGOpenPorts.getSGOpenPorts(supportClient)
       sgUsage <- getSgsUsage(sgResult, account)
-      flaggedSgs <- updateSgsWithTags(account, sgResult.flaggedResources.filter(_.status != "ok")){ (ec2Client, request) =>
-        handleAWSErrs(awsToScala(ec2Client.describeSecurityGroupsAsync)(request))
-      }
-      vpcs <- getVpcs(account, flaggedSgs)(getVpcsDetails)
-      flaggedSgsWithVpc = addVpcName(flaggedSgs, vpcs)
+      flaggedSgs = sgResult.flaggedResources.filter(_.status != "ok")
+      flaggedSgsIds = flaggedSgs.map(_.id)
+      regions = flaggedSgs.map(sg => Regions.fromName(sg.region)).distinct
+      clients = regions.map(client(account, _))
+      describeSecurityGroupsResults <- Attempt.traverse(clients)(EC2.describeSecurityGroups(flaggedSgsIds))
+      sgTagDetails = describeSecurityGroupsResults.flatMap(extractTagsForSecurityGroups).toMap
+      enrichedFlaggedSgs = enrichSecurityGroups(flaggedSgs, sgTagDetails)
+      vpcs <- getVpcs(account, enrichedFlaggedSgs)(getVpcsDetails)
+      flaggedSgsWithVpc = addVpcName(enrichedFlaggedSgs, vpcs)
     } yield sortSecurityGroupsByInUse(flaggedSgsWithVpc, sgUsage)
   }
 
@@ -94,6 +98,30 @@ object EC2 {
     }
   }
 
+  private def describeSecurityGroups(sgIds: List[String])(client: AmazonEC2Async)(implicit ec: ExecutionContext): Attempt[DescribeSecurityGroupsResult] = {
+    val request = new DescribeSecurityGroupsRequest()
+      .withFilters(new Filter("group-id", sgIds.asJava))
+    handleAWSErrs(awsToScala(client.describeSecurityGroupsAsync)(request))
+  }
+
+  private[ec2] def extractTagsForSecurityGroups(describeSecurityGroupsResult: DescribeSecurityGroupsResult): Map[String, List[Tag]] = {
+    val sgs = describeSecurityGroupsResult.getSecurityGroups.asScala.toList
+    sgs.map { sg =>
+      sg.getGroupId -> sg.getTags.asScala.toList
+    }.toMap
+  }
+
+  private[ec2] def enrichSecurityGroups(sGOpenPortsDetails: List[SGOpenPortsDetail], sgTagDetails: Map[String, List[Tag]]): List[SGOpenPortsDetail] = {
+    sGOpenPortsDetails.map { sGOpenPortsDetail =>
+      val temp = for {
+        tags <- sgTagDetails.get(sGOpenPortsDetail.id)
+        cfStackNameTag <- tags.find(_.getKey == "aws:cloudformation:stack-name")
+        cfStackIdTag <- tags.find(_.getKey == "aws:cloudformation:stack-id")
+      } yield sGOpenPortsDetail.copy(stackName = Some(cfStackNameTag.getValue), stackId = Some(cfStackIdTag.getValue))
+      temp.getOrElse(sGOpenPortsDetail)
+    }
+  }
+
   private[ec2] def getSgsUsageForRegion(sgIds: List[String], client: AmazonEC2Async)(implicit ec: ExecutionContext): Attempt[DescribeNetworkInterfacesResult] = {
     val request = new DescribeNetworkInterfacesRequest()
         .withFilters(new Filter("group-id", sgIds.asJava))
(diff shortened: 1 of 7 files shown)


abdusahin:
@adamnfish . @katebee , @aug24  here is the PR for displaying Cloudformation stack name on SGS page.

adamnfish:
Hard to see how this differs to https://github.com/guardian/security-hq/pull/30 so I'll defer review until that one is changed and merged.

katebee:
@abdusahin are all these changes included in #30? Do you want to resolve the conflicts, or can this be closed?

abdusahin:
@katebee resolved the merge conflicts

hq/app/views/fragments/openSecurityGroups.scala.html
```diff
@@ -73,6 +73,17 @@
                             <th>Region</th>
                             <td>@flaggedSg.region</td>
                         </tr>
+
+                        <tr>
+                            <th>Cloudformation Stack</th>
+                            <td>@flaggedSg.stackId match {
+                                case Some(id) => {
+                                    <a target="_blank" rel="noopener noreferrer" href="https://@{flaggedSg.region}.console.aws.amazon.com/cloudformation/home?region=@{flaggedSg.region}#/stack/detail?stackId=@helper.urlEncode(id)">@flaggedSg.stackName.get</a>
```
katebee: `@flaggedSg.stackName.get`...
Is it possible to have a stackId, without a stackName? If so, is there a chance of trying to `get` a None here?

katebee: `<a target="_blank" rel="noopener noreferrer"`
I would rather use target blank sparingly, and if we do, use it alongside the `open_in_new` md icon to indicate that we are overriding the default behaviour of the browser.

Not a blocker to getting this merged, however it would be good to overview/reduce the usage of target=_blank in the application. (Maybe the links could be target=_self then the icon could be target=_blank?)

hq/app/aws/ec2/EC2.scala
```diff
@@ -123,6 +125,36 @@ object EC2 {
       )
   }
 
+  def updateSgsWithTags(account: AwsAccount, sgIds: List[SGOpenPortsDetail])
+                       (sgsDetailsFn: (AmazonEC2Async, DescribeSecurityGroupsRequest) => Attempt[DescribeSecurityGroupsResult])
```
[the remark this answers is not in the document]
adamnfish: Also seems like it's doing quite a few different things. I'ma stare at it and see if it can be simplified.

hq/test/aws/ec2/EC2Test.scala
```diff
@@ -140,6 +141,22 @@ class EC2Test extends FreeSpec with Matchers with Checkers with PropertyChecks w
         sortedResult should be(sortedResult.sortWith { case ((_, s1), (_, s2)) => s1.size > s2.size })
       }
     }
+
+    "add cloudformation stackId & stackName" in {
```
adamnfish: Good to check the happy path, but with only one test we're missing quite a few states.

hq/app/views/fragments/openSecurityGroups.scala.html
adamnfish: I think that's a great idea, pull that in as a consistent approach.

hq/test/aws/ec2/EC2Test.scala
```diff
@@ -140,6 +141,22 @@ class EC2Test extends FreeSpec with Matchers with Checkers with PropertyChecks w
...
+      val sgs1 = new SecurityGroup().withGroupId("id-1").withTags(new Tag().withKey("aws:cloudformation:stack-id").withValue("stack-id-1"), new Tag().withKey("aws:cloudformation:stack-name").withValue("stack-name-1"))
+      val sgs2 = new SecurityGroup().withGroupId("id-2").withTags(new Tag().withKey("aws:cloudformation:stack-id").withValue("stack-id-2"), new Tag().withKey("aws:cloudformation:stack-name").withValue("stack-name-2"))
+      val sgs3 = new SecurityGroup().withGroupId("id-3")
+      val result = new DescribeSecurityGroupsResult().withSecurityGroups(sgs1, sgs2, sgs3)
+      val sgsOpenPort1 = SGOpenPortsDetail("warning", "eu-west-1", "name-1", "id-1", "vpc-1", "tcp", "8080", "Red", false )
+      val sgsOpenPort2 = SGOpenPortsDetail("warning", "eu-west-2", "name-2", "id-2", "vpc-2", "tcp", "8080", "Red", false )
+      val sgsOpenPort3 = SGOpenPortsDetail("warning", "eu-west-1", "name-4", "id-4", "vpc-4", "tcp", "8080", "Red", false )
```
katebee: Minor: Is there a reason that `sgsOpenPort3` is using `4`?   `"name-4", "id-4", "vpc-4"` would it make more sense to use `3`?

hq/test/aws/ec2/EC2Test.scala
```diff
@@ -140,6 +141,22 @@ class EC2Test extends FreeSpec with Matchers with Checkers with PropertyChecks w
...
+      val account = AwsAccount("id-1", "security", "security-role")
+      EC2.updateSgsWithTags(account, List(sgsOpenPort1, sgsOpenPort2, sgsOpenPort3))((_, _) => Attempt.Right(result) ).value().toSet shouldBe
+        Set(
+        sgsOpenPort1.copy(stackId = Some("stack-id-1"), stackName = Some("stack-name-1")),
+        sgsOpenPort2.copy(stackId = Some("stack-id-2"), stackName = Some("stack-name-2")),
+        sgsOpenPort3)
+    }
```
katebee: Will it be okay if there is a Security Group with a stackId, but no stackName?

Tried this:
```
    "add cloudformation stackId & stackName" in {
      val sgs1 = new SecurityGroup().withGroupId("id-1").withTags(new Tag().withKey("aws:cloudformation:stack-id").withValue("stack-id-1"), new Tag().withKey("aws:cloudformation:stack-name").withValue("stack-name-1"))
      val sgs2 = new SecurityGroup().withGroupId("id-2").withTags(new Tag().withKey("aws:cloudformation:stack-id").withValue("stack-id-2"), new Tag().withKey("aws:cloudformation:stack-name").withValue("stack-name-2"))
      val sgs3 = new SecurityGroup().withGroupId("id-3").withTags(new Tag().withKey("aws:cloudformation:stack-id").withValue("stack-id-3"))
      val sgs4 = new SecurityGroup().withGroupId("id-4")
      val result = new DescribeSecurityGroupsResult().withSecurityGroups(sgs1, sgs2, sgs3, sgs4)
      val sgsOpenPort1 = SGOpenPortsDetail("warning", "eu-west-1", "name-1", "id-1", "vpc-1", "tcp", "8080", "Red", false )
      val sgsOpenPort2 = SGOpenPortsDetail("warning", "eu-west-2", "name-2", "id-2", "vpc-2", "tcp", "8080", "Red", false )
      val sgsOpenPort3 = SGOpenPortsDetail("warning", "eu-west-1", "name-3", "id-3", "vpc-3", "tcp", "8080", "Red", false )
      val sgsOpenPort4 = SGOpenPortsDetail("warning", "eu-west-1", "name-4", "id-4", "vpc-4", "tcp", "8080", "Red", false )
      val account = AwsAccount("id-1", "security", "security-role")
      EC2.updateSgsWithTags(account, List(sgsOpenPort1, sgsOpenPort2, sgsOpenPort3, sgsOpenPort4))((_, _) => Attempt.Right(result) ).value().toSet shouldBe
        Set(
        sgsOpenPort1.copy(stackId = Some("stack-id-1"), stackName = Some("stack-name-1")),
        sgsOpenPort2.copy(stackId = Some("stack-id-2"), stackName = Some("stack-name-2")),
        sgsOpenPort3.copy(stackId = Some("stack-id-3")),
        sgsOpenPort4)
    }
  }
```
It passes, but that does mean it is possible for the template to end up with a `Some('stackId)`, None` scenario, so we should probably use a getOrElse, or explicitly check on the existence of the stackName [on line 81 here](https://github.com/guardian/security-hq/pull/34/files#diff-7d6df40145447f3722e44ed02bf342c8R81)

hq/app/utils/attempt/Attempt.scala
```diff
@@ -67,6 +67,14 @@ object Attempt {
     as.foldRight[Attempt[List[B]]](Right(Nil))(f(_).map2(_)(_ :: _))
   }
 
+  /** Traverses the given list `List[A]` with the function f, A => Attempt[List[B]]` and flattens the generated result into Attempt[List[B]]`
```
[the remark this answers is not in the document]
abdusahin: I actually used your suggestion initially, but it felt like too long at some point when the code gets longer.

hq/app/views/fragments/openSecurityGroups.scala.html
abdusahin: @katebee I'm guessing it's unlikely to have stackId without a stack name, but i'll do a mini change for that.

hq/app/aws/ec2/EC2.scala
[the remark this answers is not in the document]
abdusahin: Added the function for the testing purpose, can be changed, however we might lose the test case.
I feel like it'd better to have AWS layer separated since more features are added.

hq/app/views/fragments/openSecurityGroups.scala.html
abdusahin: @katebee , This can be changed as below.

```
<td>@(flaggedSg.stackId, flaggedSg.stackName) match {
      case (Some(id), Some(name)) => {
        <a href="https://@{flaggedSg.region}.console.aws.amazon.com/cloudformation/home?region=@{flaggedSg.region}#/stack/detail?stackId=@helper.urlEncode(id)">@name</a>
      }
      case _ => {}
    }
</td>
```

hq/test/aws/ec2/EC2Test.scala
abdusahin: @katebee I put a mini code block that checks for stack name emptiness.

hq/app/aws/ec2/EC2.scala
```diff
@@ -92,6 +98,30 @@ object EC2 {
...
+  private[ec2] def extractTagsForSecurityGroups(describeSecurityGroupsResult: DescribeSecurityGroupsResult): Map[String, List[Tag]] = {
+    val sgs = describeSecurityGroupsResult.getSecurityGroups.asScala.toList
+    sgs.map { sg =>
+      sg.getGroupId -> sg.getTags.asScala.toList
+    }.toMap
+  }
+
+  private[ec2] def enrichSecurityGroups(sGOpenPortsDetails: List[SGOpenPortsDetail], sgTagDetails: Map[String, List[Tag]]): List[SGOpenPortsDetail] = {
+    sGOpenPortsDetails.map { sGOpenPortsDetail =>
+      val temp = for {
```
katebee: Ah! I should rename this to something more descriptive

hq/app/aws/ec2/EC2.scala
```diff
@@ -55,11 +55,15 @@ object EC2 {
     for {
       sgResult <- TrustedAdvisorSGOpenPorts.getSGOpenPorts(supportClient)
       sgUsage <- getSgsUsage(sgResult, account)
-      flaggedSgs <- updateSgsWithTags(account, sgResult.flaggedResources.filter(_.status != "ok")){ (ec2Client, request) =>
-        handleAWSErrs(awsToScala(ec2Client.describeSecurityGroupsAsync)(request))
-      }
-      vpcs <- getVpcs(account, flaggedSgs)(getVpcsDetails)
-      flaggedSgsWithVpc = addVpcName(flaggedSgs, vpcs)
+      flaggedSgs = sgResult.flaggedResources.filter(_.status != "ok")
```
adamnfish: This is ok for now but the for comp is a little on the long side. There's tension between the positive impacts of "be explicit in the for comp and avoid deeply nested function calls" and "keep the for comprehension as simple as possible because it's hard to test".

While there's no "correct" answer, there are a couple of areas where we have sequential "pure" computations (`=` rather than `<-`) where the intermediate result is not used elsewhere. This is a sign that we're doing logic in the for comp. For example:

```scala
sgTagDetails = describeSecurityGroupsResults.flatMap(extractTagsForSecurityGroups).toMap
enrichedFlaggedSgs = enrichSecurityGroups(flaggedSgs, sgTagDetails)
```

I think we could probably refactor this to do all the IO up front and then combine some of the pure work we do into a function that we can test. Let's look at doing this afterwards though, particularly as we're aware the info might be coming from different locations (cached data) soon.

PR #35, opened 2017-12-18
Support run command

aug24:


hq/app/views/documentation/unknown.scala.html
```diff
@@ -0,0 +1,9 @@
+@(file: String)(implicit assets: AssetsFinder)
+
+<h2>Not found</h2>
+<p>No documentation found on "@file".</p>
+<p>Consider searching on
+    <a target="_blank" href="https://www.google.com/search?q=@file">Google</a>, or
```
katebee: I know that we have used target blank elsewhere, but I would prefer avoiding it as far as possible.

Both of these would be okay opening in same window

adamnfish:
This looks very similar to the other PR (which I have reviewed in detail). I'll skip this one for now as a result!

PR #36, opened 2017-12-18
Add documentation service

aug24:


hq/markdown/snyk.md
```diff
@@ -0,0 +1,227 @@
...
+Snyk is a managed service which uses a dependency analysis tool and a vulnerability database to inform you of
+vulnerabilities in your application.
+
+## Installing
+
+Snyk requires node, and `npm`.  To install on a mac, please use:
+
+```
+brew install npm
+npm install -g snyk
```
adamnfish: What's the unicode char at the start of the line here? I'm seeing a character error (box)

hq/markdown/snyk.md
```diff
@@ -0,0 +1,227 @@
...
+```
+sbt test
+snyk test --file=build.sbt [--show-vulnerable-paths=(true|false)]
+```
+
+### Gotchas
+
+#### With SBT 1.0.0+
+
+It appears that sbt expects to invoke `dependency-tree`.  After sbt 1.0, this command appears to be `dependencyTree`.
```
adamnfish: Be good to raise this issue with Snyk if we haven't already!

hq/markdown/snyk.md
```diff
@@ -0,0 +1,227 @@
...
+step for more than ten seconds or so, you should check your `build.sbt` is good.
+
+## Use with Nodejs
+
+### Pre-requisites
+
+Ensure your node packages are up to date with the following:
+
+```
+rm node_modules
```
adamnfish: You sure this is required? Could impact other stuff in the project?

hq/markdown/snyk.md
```diff
@@ -0,0 +1,227 @@
...
+```
+$ git diff build.sbt
+diff --git a/build.sbt b/build.sbt
+index 434dbb2..d123098 100644
+--- a/build.sbt
++++ b/build.sbt
+@@ -29,7 +29,11 @@ libraryDependencies ++= Seq(
+   "org.typelevel" %% "cats" % "0.8.1",
+-  "my.library" %% "cleverstuff" % "0.0.99"
++  "my.library" %% "cleverstuff" % "1.0.0+"
```
adamnfish: We try to use specific versions for our build definitions rather than ranges (e.g. `+`).

hq/markdown/snyk.md
```diff
@@ -0,0 +1,227 @@
...
+  'SNYK-JAVA-ORGAPACHEHTTPCOMPONENTS-31517':
+    - '* > org.apache.httpcomponents:httpclient':
+      reason: 'No remediation available; moving to HTTPS anyway'
+      expires: 2018-01-11T00:00:00Z
+```
+
+Note that the entry includes an expiry date.  Good practice would suggest that this should be of reasonable short duration.
+Long expiry times will mask problems.
+
+## Reviewing Expired Exemptions
```
adamnfish: If we're encouraging this approach we'll need to work with InfoSec to make sure they have visibility of these exemptions and can be involved in the decision to defer resolution.

hq/markdown/snyk.md
```diff
@@ -0,0 +1,227 @@
...
+
+When reviewing accepted risks, it would be prudent to examine the git history to discover how long a problem has existed.
```
adamnfish: In particular here, while teams are responsible for the security of their own products, InfoSec are the best place to go for guidance on what would be appropriate.

hq/app/logic/DocumentUtil.scala
```diff
@@ -0,0 +1,26 @@
...
+import com.vladsch.flexmark.parser.Parser
+import com.vladsch.flexmark.util.options.MutableDataSet
+import java.io.InputStream
+import play.twirl.api.Html
+
+object DocumentUtil {
+
+  def convert(markdownfile: String): Option[Html] = {
+    val filename = s"/$markdownfile.md"
+    var is = getClass.getResourceAsStream(filename)
```
adamnfish: Make sure this is safe. Your old version had a whitelist of possible documentation paths and I think that is probably worth keeping.

For example, if the conf directory is on the classpath an attacker could use this to read the production config file.

hq/markdown/unknown.md
```diff
@@ -0,0 +1,3 @@
+# apples
```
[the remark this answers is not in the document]
katebee: I love it, can we keep it? 😂 🍎 🍏

hq/app/views/documentation/unknown.scala.html
```diff
@@ -0,0 +1,9 @@
+@(file: String)(implicit assets: AssetsFinder)
+
+<h2>Not found</h2>
+<p>No documentation found on "@file".</p>
+<p>Consider searching on
+    <a target="_blank" href="https://www.google.com/search?q=@file">Google</a>, or
```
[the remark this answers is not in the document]
katebee: Same security issue here as in the other PR:

For the SGs and IAM links to AWS, several people asked it these could open in a new tab, so the knowledge may not be that common and it might be worthwhile keeping those as target blank links 😅

However, I think we should encourage people to discover ctrl click / cmd click. So I feel happy that these could default to opening in the same tab

adamnfish:
Also, can we have descriptions on Pull Requests, even if they're very brief? We've chatted IRL about this but it's good to have confirmed what the goals of the change are. Especially true for a tool like this one where there might be other people in the department interested in what's changing.

hq/markdown/index.md
```diff
@@ -0,0 +1,3 @@
+# List of help pages
+
+   1. How to implement [Snyk](snyk) Vulnerability Analysis
```
katebee: Am I missing something here, or should `(snyk)` be a url?

hq/markdown/snyk.md
aug24: God knows, but it's gone now.

hq/markdown/snyk.md
aug24: I've briefly discussed this with Kate and we will do so.

hq/app/logic/DocumentUtil.scala
aug24: Only files ending .md can be fetched; is that sufficient whitelisting?

hq/app/logic/DocumentUtil.scala
```diff
@@ -0,0 +1,26 @@
...
+    if (is == null) None else convertInputStream(is)
+  }
+
+  private def convertInputStream(is: InputStream) = {
```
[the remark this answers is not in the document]
aug24: I rewrote that whole class in more scala idiom code.  It was irritating me that it was obviously a chunk of Java.

hq/app/views/documentation/unknown.scala.html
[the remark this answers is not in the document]
aug24: I'll take it out then.

hq/app/views/documentation/unknown.scala.html
```diff
@@ -0,0 +1,9 @@
+@(file: String)(implicit assets: AssetsFinder)
+
+<h2>Not found</h2>
+<p>No documentation found on "@file".</p>
```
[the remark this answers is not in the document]
aug24: You thinking people could inject html here?  I'm not sure what other mechanism we might be worried about.

hq/markdown/run-command.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+### Run a command
+
+Specify the command and target:
+```
+export command="uname -a"
+```
+```
+export instance="i-0fe40a72847b61cc6"
```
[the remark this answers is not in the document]
aug24: I'll bear that in mind.

hq/markdown/snyk.md
aug24: It shouldn't do any harm.  It was intended to prevent checking vulnerabilities on present dependencies, then building/packaging with newer.

hq/markdown/snyk.md
aug24: OK.  I'll change the example.

hq/markdown/unknown.md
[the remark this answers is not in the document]
aug24: I noticed that couple of hours ago and removed it :)

hq/markdown/index.md
aug24: It is a _relative_ url.

hq/markdown/snyk.md
katebee: Minor: Some projects use Yarn, and have custom install commands ..they may also have a lot of dependencies so reinstalling the node modules would be something I would groan about doing.

Is there any way this could be optional, or more generic? 

```
Ensure your node packages are up to date by using the appropriate command for your project. It will probably be either: `yarn upgrade` or `npm update`
```

Would an update be enough?

katebee: Will `rm` work? Or are we going to need to `rm -r` if node_modules is a directory?

hq/markdown/unknown.md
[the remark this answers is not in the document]
katebee: Reminds me of this:

https://github.com/guardian/frontend/blob/master/docs/99-archives/crepes.md

hq/markdown/snyk.md
aug24: Added to Trello backlog

hq/markdown/snyk.md
aug24: Fair points.  Amended.

hq/markdown/snyk.md
aug24: Added to Trello.

hq/app/views/documentation/unknown.scala.html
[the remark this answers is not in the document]
aug24: I've had a bit of a play with this.  As far as I can see, Play either correctly handles % escapes or refuses the request altogether if it includes <, > etc.

PR #39, opened 2017-12-20
Give Documentation page the new nav, add links to the docs in header and footer

katebee:
#### 👉 Give the Docs page the new header added in #33 

- Consistent design, header and headings taking less vertical space

#### 👉 Link to the Docs in the header, and add a burger and side nav for mobile/tablet:

- Making documentation discoverable by linking to it in the header

picture 115

dec-19-2017 16-56-36

#### 👉 Linking to the docs in the footer and adding icons for each link:

- Making documentation discoverable by linking to it in the footer

picture 116

#### 👉 Truncating long resource names on the security groups page:

- Improving visual display at various breakpoints

picture 117

### DIFF
--- a/hq/app/views/main.scala.html
+++ b/hq/app/views/main.scala.html
@@ -22,10 +22,9 @@
 
         @views.html.header()
 
-        @pageHeader
-
         <main>
-        @content
+          @pageHeader
+          @content
         </main>
 
         <footer class="page-footer grey darken-4">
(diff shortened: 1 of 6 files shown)


hq/app/views/main.scala.html
```diff
@@ -22,10 +22,9 @@
 
         @views.html.header()
 
-        @pageHeader
-
         <main>
-        @content
+          @pageHeader
```
katebee: Moving pageheader into one of the html sections, so it is not alone for Christmas 🎄🎁

(also, it should be a child of either `header`, `main` or `footer`)

adamnfish: Great spot!

PR #40, opened 2017-12-20
- Handle empty VPC id in security groups

abdusahin:


hq/app/aws/ec2/EC2.scala
```diff
@@ -124,24 +124,23 @@ object EC2 {
...
-    for {
-      sgs <- flaggedSgs
-      vpc <- vpcs.collectFirst { case (id, v) if id == sgs.vpcId => v }
-      vpcN = vpc.getTags.asScala.collectFirst { case tag if tag.getKey == "Name" => tag.getValue }
-      sgsUpdated = sgs.copy(vpcName = vpcN)
-    } yield sgsUpdated
+    def vpcName(vpc: Vpc) = vpc.getTags.asScala.collectFirst { case tag if tag.getKey == "Name" => tag.getValue }
+
+    flaggedSgs.map {
+      case s if s.vpcId.nonEmpty => s.copy(vpcName = vpcs.get(s.vpcId) flatMap vpcName)
```
katebee: As discussed IRL... 😀 

If we are certain that the `.get(` here cannot fail if the `vpcId` doesn't match anything in the `vpcs`, then this looks okay to me 👍

hq/test/aws/ec2/EC2Test.scala
```diff
@@ -146,28 +146,29 @@ class EC2Test extends FreeSpec with Matchers with Checkers with PropertyChecks w
...
       "returns empty vpc details" in {
-        EC2.getVpcs(AwsAccount("security-test", "security", "security-test"), sgsList)((_, _) => Attempt.Right(Map.empty)).value shouldBe Map.empty
+        EC2.getVpcs(AwsAccount("security-test", "security", "security-test"), sgsList)( _ => Attempt.Right(Map.empty)).value shouldBe Map.empty
       }
     }
 
     "addVpcName" - {
       "updates SGSOpenPortDetails with vpc name if vpc ids match" in {
-        EC2.addVpcName(sgsList, vpcsMap) shouldBe List(sgs1.copy(vpcName = Some("name-1")), sgs2.copy(vpcName = Some("name-2"))) :+ sgs3
+        EC2.addVpcName(sgsList, vpcsMap) shouldBe List(sgs1.copy(vpcName = Some("name-1")), sgs2.copy(vpcName = Some("name-2"))) :+ sgs3 :+ sgs4
```
katebee: What would happen if I pass this:

`val sgs5 = SGOpenPortsDetail("Ok", "eu-west-2", "name-3", "id-122", "vpc-missing", "tcp", "1099", "Yellow", false)`

Assuming `vpc-missing` is a non-empty string, but cannot be found in our vpcsMap?

(Please disregard me if this could never happen in the wild!)

abdusahin: Since `vpcs.get(s.vpcId)` will return `None` the `vpcName` function won't be called.

PR #43, opened 2017-12-22
Cache security groups

abdusahin:
## What does this change?
Caches security groups, IAM exposed keys and Credential Report

## What is the value of this?
Helps rendering page very quickly and avoids hitting AWS for every page request.

To improve page display performance


hq/app/utils/attempt/Failure.scala
```diff
@@ -59,6 +59,23 @@ object Failure {
...
+    Failure(
+      s"Cannot retrieve exposed IAM keys for $accountId",
+      s"Cannot retrieve exposed IAM keys for $accountId",
+      500)
+def cannotGetCredentialReports(accountId: String): Failure =
+    Failure(
+      s"Cannot retrieve credential report for $accountId",
+      s"Cannot retrieve credential report for $accountId",
+      500)
+
```
katebee: Maybe all of these could use [awsError](https://github.com/guardian/security-hq/pull/43/commits/0fbfa567d3e22b4deb6dd694d422b8e1b9bb6ebd#diff-18fd0f60f6cd85e31e36fa7086700be8R28) with the relevant string ("Security Groups" / "exposed IAM keys") being passed in? This would remove some duplication. 😄

katebee: Or maybe we could build on awsError to make a serviceError that could be more specific?

Would something like this work:

```
  def cacheServiceError(accountId: String, serviceNameOpt: Option[String]): Failure = {
    val details = serviceNameOpt.fold("Cache service error, unable to retrieve required information for $accountId") { serviceDetails =>
      s"Cache service error, unable to retrieve $serviceDetails for $accountId"
    }
    val friendlyMessage = serviceNameOpt.fold("Unable to retrieve required information for $accountId") { serviceDetails =>
      s"Unable to retrieve $serviceDetails for $accountId"
    }
  }
```

Or is it not worth it? (YMMV) 🙃

hq/app/service/SecurityGroupsService.scala
```diff
@@ -0,0 +1,43 @@
...
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+
+object SecurityGroupsService {
+
+  type EitherFlaggedSgs = Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]]
+  type FlaggedSgs = List[(AwsAccount, EitherFlaggedSgs)]
+
+  private val sgsBox = Box[FlaggedSgs](List.empty)
```
katebee: 🤔 Thinking out loud:

For the securityGroupService, the ideal data structure of the Box (sgsBox) would be:
```
Box[Map[String, Attempt[Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]]]]](Map.empty)
```

This would allow us to skip the filtering step in the controller (for single accounts) and minimise the amount transformation required

PR #44, opened 2017-12-22
Cache security groupsv2

abdusahin:
- Caches security groups, IAM exposed keys and Credential Report
- Helps rendering page very quickly and avoids hitting AWS for every page request.
- Improves page display performance


hq/app/utils/attempt/Failure.scala
```diff
@@ -59,6 +59,23 @@ object Failure {
...
+    Failure(
+      s"Cannot retrieve exposed IAM keys for $accountId",
+      s"Cannot retrieve exposed IAM keys for $accountId",
+      500)
+def cannotGetCredentialReports(accountId: String): Failure =
+    Failure(
+      s"Cannot retrieve credential report for $accountId",
+      s"Cannot retrieve credential report for $accountId",
+      500)
+
```
katebee: Maybe all of these could use [awsError](https://github.com/guardian/security-hq/pull/43/commits/0fbfa567d3e22b4deb6dd694d422b8e1b9bb6ebd#diff-18fd0f60f6cd85e31e36fa7086700be8R28) with the relevant string ("Security Groups" / "exposed IAM keys") being passed in? This would remove some duplication. 😄 

Or maybe we could build on awsError to make a serviceError that could be more specific?

Would something like this work:

```
  def cacheServiceError(accountId: String, serviceNameOpt: Option[String]): Failure = {
    val details = serviceNameOpt.fold("Cache service error, unable to retrieve required information for $accountId") { serviceDetails =>
      s"Cache service error, unable to retrieve $serviceDetails for $accountId"
    }
    val friendlyMessage = serviceNameOpt.fold("Unable to retrieve required information for $accountId") { serviceDetails =>
      s"Unable to retrieve $serviceDetails for $accountId"
    }
  }
```

Or is it not worth it? (YMMV) 🙃 

(copying some comments from the closed PR)

hq/app/service/CacheService.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  def update[T](actorSystem: ActorSystem)(awsWorkerFn: (List[AwsAccount]) => Attempt[T])(boxUpdaterFn: Attempt[T] => Attempt[Unit])(implicit ec: ExecutionContext) = {
+    val cacheInterval = Config.getAwsApiCacheInterval(config)
+    val subscription = Observable.interval(10.millisecond, cacheInterval).doOnEach { _ =>
+      boxUpdaterFn(awsWorkerFn(accounts))
+    }.subscribe
+
+    actorSystem.registerOnTermination(
+      subscription.unsubscribe()
+    )
```
katebee: It is always difficult to walk the line between DRY and code coupling; in this case, I would prefer an approach where the caches use their own update functions. This would improve the maintainability and readability, and simplify the method since we no longer have to pass the necessary method into it.

We ran into difficulties when trying this out and then debugging the cache services (e.g. which call to update() failed?). I also believe that the

If we want to update the caches at different intervals, that would require further abstraction.

hq/app/AppComponents.scala
```diff
@@ -8,12 +12,15 @@ import play.api.mvc.{AnyContent, BodyParser, ControllerComponents}
...
+import service._
 
 
 class AppComponents(context: Context)
   extends BuiltInComponentsFromContext(context)
   with CSRFComponents
-  with AhcWSComponents with AssetsComponents {
+  with AhcWSComponents with AssetsComponents  {
+
+  val cacheService = new CacheService(configuration)
```
katebee: What are the benefits and risks of having many cache services vs. one? I have a (maybe unfounded) preference of splitting the responsibilities into two services:
```
val sgsUpdater = new SecurityGroupsUpdater(configuration)
val credUpdater = new CredentialServiceUpdater(configuration)
```
Would this have any impact on performance? Would it make it harder or easier to expand functionality or debug?

hq/app/service/CacheService.scala
```diff
@@ -0,0 +1,58 @@
...
+import utils.attempt.{Attempt, FailedAttempt}
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+class CacheService(config: Configuration) {
+  private val accounts = Config.getAwsAccounts(config)
+
+  def update[T](actorSystem: ActorSystem)(awsWorkerFn: (List[AwsAccount]) => Attempt[T])(boxUpdaterFn: Attempt[T] => Attempt[Unit])(implicit ec: ExecutionContext) = {
+    val cacheInterval = Config.getAwsApiCacheInterval(config)
```
katebee: Is it worth bringing this out of the method to join `accounts` at the class level, for consistency? (line 15)
```
private val cacheInterval = Config.getAwsApiCacheInterval(config)
```

hq/app/service/CacheService.scala
```diff
@@ -0,0 +1,58 @@
...
+
+object CacheService {
+  type EitherFlaggedSgs = Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]]
+  type FlaggedSgs = List[(AwsAccount, EitherFlaggedSgs)]
+  type ExposedKeys = List[Either[FailedAttempt, (AwsAccount, TrustedAdvisorDetailsResult[ExposedIAMKeyDetail])]]
+  type CredentialReport = Seq[(AwsAccount, Either[FailedAttempt, CredentialReportDisplay])]
+
+  private val exposedKeysBox = Box[ExposedKeys](List.empty)
+  private val credentialReportBox = Box[CredentialReport](List.empty)
+  private val sgsBox = Box[FlaggedSgs](List.empty)
```
katebee: It might be worth revisiting these boxes (another PR?) once we have a clear idea for our data structures. Ideally, the boxes would store the data in the most helpful shape ready for use. For example:

The ideal data structure of the sgsBox might be:
```
Box[Map[String, Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]]]]]
```
Would this allow us to skip the filtering step in the controller (for single accounts) and minimise the amount transformation required?

hq/app/AppComponents.scala
abdusahin: @katebee performance wise, I don't expect any difference at all really. Can be separated if you like.

katebee:
We have split these changes into two separate PRs (#53 and #54)  so I will close this one 😊

PR #50, opened 2018-01-08
Expansion of the EC2 run readme, and addition of an example shell script

katebee:
### 👉 Adding more links to external resources

Saving everyone a few clicks

picture 137

### 👉 Reordering content and adding expanding some sections

I think this order makes it easier to follow along and learn as you go. I ran through the guide and had some questions, so adding some verbosity for any AWS/EC2 beginners.

### 👉 Changing the command used in the wait_for_command function

Script was not working as expected but was giving the helpful error:

```
Invalid choice: 'get-command-invocation', maybe you meant:

  * list-command-invocations
```

### 👉 Copying the shell script into an example file

The docs can link out to the example, and this will make it easier for others to copy the content

picture 138

### 👉 Adding line breaks before triple backticks

No visible change to web view. Some markdown viewers require this to format correctly.

Nope

### DIFF
--- a/hq/markdown/run-command.md
+++ b/hq/markdown/run-command.md
@@ -3,32 +3,43 @@
 ## Prerequisites for this tutorial
 
  * jq - the command line json manipulation program
- * aws - the command line amazon tool
+ * aws - the command line amazon tool (awscli)
  * some understanding of bash
  * some understanding of IAM, EC2 and Cloudformation
 
-## Create IAM client permissions
+OS X users can download jq and awscli using Homebrew:
+
+```
+brew install jq
+brew install awscli
+
+```
+
+
+## Using EC2run on a configured instance
+
+If you have not yet setup EC2 run on the instance, then go to the second section....
+
+## Setting up a instance with EC2run
+
+
+
+### Create IAM client permissions
 
 Use an admin account, then you can skip this. If not, it is task 1 in doc 1 below
 
-## Installation
+### Installation
 
-Choose the target account and region.  For example:
-```
-export region="eu-west-1"
-export profile="security"
-```
-
-### Create an IAM role (this will need to be done in cloudformation)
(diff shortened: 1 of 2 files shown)


ssm-util.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
```
adamnfish: I'd consider 'auto-formatting' this in IntelliJ or something along those lines. At least, the indentation is inconsistent in `wait_for_command` which might be worth a tweak.

Minor point though and no worries.

hq/markdown/run-command.md
```diff
@@ -92,14 +122,14 @@ for this_instance in $instance; do
    while [[ $responseCode -eq -1 ]]; do
       printf "."
       sleep 1
-      responseCode=$(aws --region $region --profile $profile ssm get-command-invocation \
+      responseCode=$(aws --region $region --profile $profile ssm list-command-invocations \
```
adamnfish: This is a significant change!

cc @aug24 for approval

shtukas: Let's wait for @aug24 input, but the proposed version is correct. (What I think happened, but this is only a theory, is that @aug24 might have looked that the Windows PowerShell version, but @katebeeis correctly reporting the `aws` cli version.)

The AWS documentstion for `aws` says 

```
aws ssm list-command-invocations --command-id "command ID" --details
``` 

and @katebee proposes

```
aws ssm list-command-invocations \
    --region $region --profile $profile \
    --command-id "$command_id" \
    --instance-id "$this_instance" \
         | jq -r '.ResponseCode'
```

@adamnfish

shtukas:
All good for me. I noticed that the `read_command_output` adopt the optimistic case that the command output will have less than 2500 lines, but as a proof of concept it's all right 

:+1:

PR #53, opened 2018-01-09
Add a cache service and integrate it with Security Groups

katebee:
## What does this change?

Adds caching to the Security Groups

The service now stores the AWS response on flagged security groups into a [Box](https://github.com/guardian/box) 📦

Every 5 minutes, the service will refresh the Box content

The application can get the data from the Box, rather than running off to AWS each time. 

## What is the value of this?

Speeds up the page load for the all accounts security groups page, and for the single account pages

No long trips over the wire delaying page load of the Security Groups 🏃💨

Nope

### DIFF
--- a/hq/app/services/CacheService.scala
+++ b/hq/app/services/CacheService.scala
@@ -1,0 +1,49 @@
+package services
+
+import aws.ec2.EC2
+import com.gu.Box
+import config.Config
+import model.{AwsAccount, SGInUse, SGOpenPortsDetail}
+import play.api.inject.ApplicationLifecycle
+import play.api.{Configuration, Environment, Mode}
+import rx.lang.scala.Observable
+import utils.attempt.{FailedAttempt, Failure}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
+
+
+class CacheService(config: Configuration, lifecycle: ApplicationLifecycle, environment: Environment)(implicit ec: ExecutionContext) {
+  private val sgsBox: Box[Map[AwsAccount, Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]]]] = Box(Map.empty)
+  private val accounts = Config.getAwsAccounts(config)
+
+  def getAllSgs(): Map[AwsAccount, Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]]] = sgsBox.get()
+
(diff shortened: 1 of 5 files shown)


hq/app/services/CacheService.scala
```diff
@@ -0,0 +1,49 @@
...
+      Left(Failure("unable to find account data in the cache", "No security group data available", 500, Some(awsAccount.id)).attempt)
+    )
+  }
+
+  private def refreshSgsBox(): Unit = {
+    println("Started refresh of the Security Groups")
+    for {
+      allFlaggedSgs <- EC2.allFlaggedSgs(accounts)
+    } yield {
+      println("Sending the refreshed data to the Security Groups Box")
```
katebee: Should these be switched over to using Logger? Or shall I delete both printlns?

adamnfish: Yes they definitely should, good sport.

I think `info` is probably a good log level for this.

hq/app/services/CacheService.scala
```diff
@@ -0,0 +1,49 @@
...
+      sgsBox.send(allFlaggedSgs.toMap)
```
adamnfish: If a call fails, we'll write that failure into the cache, potentially overwriting real data. This might be what we want (always see the state of things) or it might not (we might decide serving stale is ok in case of problems). I'm not sure which is best and we shouldn't change it here, but this is something for us to consider.

hq/app/services/CacheService.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
+
+
+class CacheService(config: Configuration, lifecycle: ApplicationLifecycle, environment: Environment)(implicit ec: ExecutionContext) {
+  private val sgsBox: Box[Map[AwsAccount, Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]]]] = Box(Map.empty)
+  private val accounts = Config.getAwsAccounts(config)
+
+  def getAllSgs(): Map[AwsAccount, Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]]] = sgsBox.get()
```
adamnfish: We did talk about initialising this data. Happy to defer this to another PR if you like, but I feel like it might be worth doing this to distinguish between accounts that aren't found and accounts that haven't yet been updated.

shtukas:
Very nice +1

hq/app/services/CacheService.scala
katebee: Yep, sounds sensible! 👍 
I would like to do a followup PR for setting starting values in each box

PR #54, opened 2018-01-10
Expand the cache service to include Exposed Keys and Credentials Report

katebee:
## What does this change?

Adding methods to help get exposed keys for all accounts

Stores the AWS responses for Credentials and Exposed Keys into two separate [Boxes](https://github.com/guardian/box) 📦 📦

Adds a new Failure that can be reused across the cache service:

picture 142

## What is the value of this?

Speeds up the page load for the IAM pages

Adding the startingCache to each Box means no more blank pages while waiting for each service to start:

picture 150

Nope

## Any additional notes?

Related PR: #53

### DIFF
--- a/hq/app/services/CacheService.scala
+++ b/hq/app/services/CacheService.scala
@@ -1,9 +1,11 @@
 package services
 
 import aws.ec2.EC2
+import aws.iam.IAMClient
+import aws.support.TrustedAdvisorExposedIAMKeys
 import com.gu.Box
 import config.Config
-import model.{AwsAccount, SGInUse, SGOpenPortsDetail}
+import model._
 import play.api.inject.ApplicationLifecycle
 import play.api.{Configuration, Environment, Logger, Mode}
 import rx.lang.scala.Observable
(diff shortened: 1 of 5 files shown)


hq/app/services/CacheService.scala
```diff
@@ -14,18 +16,59 @@ import scala.concurrent.duration._
 
 
 class CacheService(config: Configuration, lifecycle: ApplicationLifecycle, environment: Environment)(implicit ec: ExecutionContext) {
-  private val sgsBox: Box[Map[AwsAccount, Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]]]] = Box(Map.empty)
   private val accounts = Config.getAwsAccounts(config)
+  private val startingCache = accounts.map(acc => (acc, Left(Failure.cacheServiceError(acc.id, "cache").attempt))).toMap
```
katebee: Discussed using type aliases for the caches to remove some duplication. For example:
`type Tmp[Report] = Map[AwsAccount, Either[FailedAttempt, Report]]`

Having tried it, I have a minor preference for the original. The lines are longer this way, but scanning along a line to understand a val is easier to me than switching to reading 3-5 above then back.

hq/app/services/CacheService.scala
```diff
@@ -14,18 +16,59 @@ import scala.concurrent.duration._
...
+  private val credentialsBox: Box[Map[AwsAccount, Either[FailedAttempt, CredentialReportDisplay]]] = Box(startingCache)
+  private val exposedKeysBox: Box[Map[AwsAccount, Either[FailedAttempt, List[ExposedIAMKeyDetail]]]] = Box(startingCache)
+  private val sgsBox: Box[Map[AwsAccount, Either[FailedAttempt, List[(SGOpenPortsDetail, Set[SGInUse])]]]] = Box(startingCache)
+
```
katebee: Discussed using type aliases for the caches to remove some duplication. For example:
`type Tmp[Report] = Map[AwsAccount, Either[FailedAttempt, Report]]`

Having tried it, I have a minor preference for the original. The lines are longer this way, but scanning along a line to understand a val is easier to me than switching to reading 3-5 above then back.

shtukas: +1

PR #55, opened 2018-01-11
Add Snyk feedback

aug24:
## What does this change?

Snyk have fed back on the changes required for SBT 1.0.  The information has been added to the snyk readme.

## What is the value of this?

It specifies a cleaner solution to using Snyk with sbt >= 1.0

No.

## Any additional notes?

No.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -33,6 +33,11 @@ lazy val hq = (project in file("hq")).
       "com.vladsch.flexmark" % "flexmark-all" % "0.28.20",
       "io.reactivex" %% "rxscala" % "0.26.5",
       "com.gu" %% "box" % "0.1.0",
+      "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11",
+      "com.fasterxml.jackson.core" % "jackson-annotations" % "2.8.11",
+      "com.fasterxml.jackson.dataformat" % "jackson-datatype-jsr310" % "2.8.10",
+      "com.fasterxml.jackson.dataformat" % "jackson-dataformat-cbor" % "2.8.10",
+      "com.fasterxml.jackson.dataformat" % "jackson-dataformat-guava" % "2.8.10",
       "org.scalatestplus.play" %% "scalatestplus-play" % "3.1.2" % Test,
       "org.scalacheck" %% "scalacheck" % "1.13.4" % Test,
       "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % "1.1.6" % Test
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -13,7 +13,8 @@
     "eslint-plugin-prettier": "2.4.0",
     "materialize-css": "0.100.2",
     "prettier": "1.8.0",
-    "sass-lint": "1.12.1"
+    "sass-lint": "1.12.1",
+    "snyk": "1.68.0"
```
katebee: If we are adding new dependencies, can we update the yarn.lock file?

 install and rebuild the yarn.lock file:
`yarn install`

`git add yarn.lock`

build.sbt
```diff
@@ -80,7 +83,9 @@ lazy val lambdaCommon = (project in file("lambda/common")).
       "com.typesafe.scala-logging" %% "scala-logging" % "3.5.0",
       "ch.qos.logback" %  "logback-classic" % "1.2.3",
       "com.amazonaws" % "aws-java-sdk-config" % "1.11.246",
-      "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.10"
+      "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11",
+      "com.fasterxml.jackson.core" % "jackson-annotations" % "2.8.11",
+      "com.fasterxml.jackson.dataformat" % "jackson-dataformat-cbor" % "2.8.10",
```
katebee: Do we need to have these three twice? (lines 36, 37 and 38 above)

hq/markdown/snyk.md
```diff
@@ -7,7 +7,11 @@ vulnerabilities in your application.
 
 ## Installing
 
-Snyk requires node, and `npm`.  To install on a mac, please use:
+Snyk requires node, and `npm`.
```
katebee: Personal preference, however I prefer Yarn... Maybe:

> Snyk requires node, and a Node package manager such as `npm` or `Yarn`

hq/markdown/snyk.md
```diff
@@ -7,7 +7,11 @@ vulnerabilities in your application.
...
+
+### Developer Install
+
+To install on a mac, please use:
```
katebee: Brew does not come as standard on MacOS therefore it may be worth adding a line about getting Homebrew?

> OS X users may want to use [Homebrew](https://brew.sh/)

package.json
```diff
@@ -13,7 +13,8 @@
     "eslint-plugin-prettier": "2.4.0",
     "materialize-css": "0.100.2",
     "prettier": "1.8.0",
-    "sass-lint": "1.12.1"
+    "sass-lint": "1.12.1",
+    "snyk": "1.69.3"
```
katebee: Can the update to the yarn.lock file be included?

https://yarnpkg.com/lang/en/docs/yarn-lock/

It should auto-generate based on changes to package.json when you run `yarn install`

build.sbt
shtukas: Good catch!

PR #56, opened 2018-01-11
Let the reports for exposed keys and credentials fail/success independently

katebee:
#### 👉 New logic to show warning(s) on IAM page if one or both of the datasets is unavailable

This PR allows the IAM account pages on Security HQ to show a warning if either the credentials or exposed keys results in a failed attempt, while still showing the other report it that succeeds. 

For example, if the Exposed Keys cache service does not start, or fails to fetch data from AWS, the error makes it to the dashboard along with the available data (credentials report):

##### BEFORE
picture 155

##### AFTER
picture 153

Worst case scenario, the user gets two bright error messages:

picture 154

#### 👉 Refactoring templates
- Making a new fragment since the pattern matching adds noise
- Deleting some extra rows from the templates, and fixing indents to match other pages

Nope

## Any additional notes?

Tested with demo exposed keys, hence the template changes to make the page layouts more consistent and aesthetically pleasing 💅

### DIFF
--- a/hq/app/views/iam/iamAccount.scala.html
+++ b/hq/app/views/iam/iamAccount.scala.html
@@ -1,8 +1,9 @@
 @import model.{AwsAccount, CredentialReportDisplay, ExposedIAMKeyDetail}
-@(awsAccount: AwsAccount, exposedIamKeys: Either[utils.attempt.FailedAttempt, List[ExposedIAMKeyDetail]], credentialsReport: Either[utils.attempt.FailedAttempt, CredentialReportDisplay])(implicit assets: AssetsFinder)
+@import utils.attempt.FailedAttempt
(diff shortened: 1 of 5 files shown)


hq/app/views/iam/iamAccount.scala.html
```diff
@@ -1,7 +1,21 @@
 @import model.{AwsAccount, CredentialReportDisplay, ExposedIAMKeyDetail}
-@(awsAccount: AwsAccount, exposedIamKeys: List[ExposedIAMKeyDetail], report: CredentialReportDisplay)(implicit assets: AssetsFinder)
+@(awsAccount: AwsAccount, exposedIamKeys: Either[utils.attempt.FailedAttempt, List[ExposedIAMKeyDetail]], credentialsReport: Either[utils.attempt.FailedAttempt, CredentialReportDisplay])(implicit assets: AssetsFinder)
```
adamnfish: If you `import utils.attempt.FailedAttempt` above, this signature will be simplified.

PR #60, opened 2018-01-17
A Better Security Groups Cache Refresh

shtukas:
## What does this change?

Currently the CacheService runs `refreshSgsBox()` every 5 minutes. This adds a step ( `EC2.refreshSGSReports` ) to that `refreshSgsBox()` by which we instruct AWS to update/refresh its security group open ports report. 

## What is the value of this?

Increase the correctness (up-to-date-ness if I may) of the cache.

Yes. watched-account.template.yaml has been update to allow Trusted Advisor to run `Refresh`.

### DIFF
--- a/cloudformation/watched-account.template.yaml
+++ b/cloudformation/watched-account.template.yaml
@@ -32,6 +32,7 @@ Resources:
             Action:
             # Analyse security groups
             - trustedadvisor:Describe*
+            - trustedadvisor:Refresh*
             - support:*
             - ec2:DescribeNetworkInterfaces
             - ec2:DescribeSecurityGroups
--- a/hq/app/aws/ec2/EC2.scala
+++ b/hq/app/aws/ec2/EC2.scala
@@ -10,8 +10,11 @@ import com.amazonaws.auth.AWSCredentialsProviderChain
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.ec2.model._
 import com.amazonaws.services.ec2.{AmazonEC2Async, AmazonEC2AsyncClientBuilder}
+import com.amazonaws.services.support.model.RefreshTrustedAdvisorCheckResult
 import model._
 import utils.attempt.{Attempt, FailedAttempt}
+
+
 
 import scala.collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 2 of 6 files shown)


shtukas:
Cloud Formation update has not been done yet

aug24:
+1 here.

cloudformation/watched-account.template.yaml
```diff
@@ -32,6 +32,7 @@ Resources:
             Action:
             # Analyse security groups
             - trustedadvisor:Describe*
+            - trustedadvisor:Refresh*
```
adamnfish: Good catch :-)

hq/app/aws/ec2/EC2.scala
```diff
@@ -10,9 +10,12 @@ import com.amazonaws.auth.AWSCredentialsProviderChain
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.ec2.model._
 import com.amazonaws.services.ec2.{AmazonEC2Async, AmazonEC2AsyncClientBuilder}
+import com.amazonaws.services.support.model.RefreshTrustedAdvisorCheckResult
 import model._
 import utils.attempt.{Attempt, FailedAttempt}
 
+
+
```
adamnfish: Is this on purpose?

hq/app/aws/ec2/EC2.scala
```diff
@@ -67,6 +70,13 @@ object EC2 {
     } yield sortSecurityGroupsByInUse(flaggedSgsWithVpc, sgUsage)
   }
 
+  def refreshSGSReports(accounts: List[AwsAccount])(implicit ec: ExecutionContext): Attempt[List[Either[FailedAttempt, RefreshTrustedAdvisorCheckResult]]] = {
```
adamnfish: This returns all the failures and successes, but we're only throwing the result away. Should we be doing something with any failures we get here and if not, could we have this function return something simpler?

hq/app/services/CacheService.scala
```diff
@@ -15,6 +15,7 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
 
 
+
```
adamnfish: Was this on purpose?

shtukas:
Cloud Formation update has been done on the Root account

PR #63, opened 2018-01-24
Use configraun

aug24:
## What does this change?

Addition of snyk pages, and fetching of snyk token from AWS parameter store.

## What is the value of this?

Easy availability of snyk reports.
Dogfood example of how to use parameter store.

Yes. The instances will now need an instance role which permits self-description and read from parameter store.

## Have you committed your changes to the CloudFormation templates?

Yes

## Has the CloudFormation or StackSet update been completed?

No

### DIFF
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -1,0 +1,71 @@
+package api
+
+import model._
+import play.api.libs.json._
+import play.api.libs.ws.WSClient
+import utils.attempt.{Attempt, FailedAttempt, Failure}
+
+import scala.concurrent.ExecutionContext
+import scala.util.control.NonFatal
+
+object Snyk {
+
+  def getSnykOrganisations(token: Token, wsClient: WSClient)(implicit ec:ExecutionContext) = {
+
+    val snykOrgUrl = "https://snyk.io/api/v1/orgs"
+
+    val futureResponse = wsClient.url(snykOrgUrl)
+      .addHttpHeaders("Authorization" -> s"token ${token.value}")
+      .get
+
+    Attempt.fromFuture(futureResponse) { case NonFatal(e) => {
+      val failure = Failure(e.getMessage, "Could not read organisations from Snyk", 502, None, Some(e))
+      FailedAttempt(failure)
+    }}
+  }
+
+  def getProjects(token: Token, id: String, wsClient: WSClient)(implicit ec:ExecutionContext) = {
+        val snykProjectsUrl = s"https://snyk.io/api/v1/org/${id}/projects"
+        val a = wsClient.url(snykProjectsUrl)
+          .addHttpHeaders("Authorization" -> s"token ${token.value}")
+          .get()
+
+    Attempt.fromFuture(a) { case NonFatal(e) => {
+      val failure = Failure(e.getMessage, "Could not read projects from Snyk", 502, None, Some(e))
+      FailedAttempt(failure)
+    }}
+  }
+
+  def getProjectVulnerabilities(id: String, projects: List[SnykProject], token: Token, wsClient: WSClient)(implicit ec:ExecutionContext) = {
+    val projectVulnerabilityResponses = projects
+      .map(project => {
+        val snykProjectUrl = s"https://snyk.io/api/v1/org/${id}/project/${project.id}/issues"
+
+        val projectIssuesFilter = Json.obj(
+          "filters" -> Json.obj(
+            "severity" -> JsArray(List(
+              JsString("high"), JsString("medium"), JsString("low")
+            )),
+            "types" -> JsArray(List(
+              JsString("vuln")
+            )),
+            "ignored" -> "false",
+            "patched" -> "false"
+          )
+        )
+
+        wsClient.url(snykProjectUrl)
+          .addHttpHeaders("Authorization" -> s"token ${token.value}")
+          .post(projectIssuesFilter)
+
+      })
+    Attempt.traverse(projectVulnerabilityResponses)(
+      projectVulnerabilityResponse => Attempt.fromFuture(projectVulnerabilityResponse) {
+        case NonFatal(e) => {
+          val failure = Failure(e.getMessage, "Could not read project vulnerabilities from Snyk", 502, None, Some(e))
+          FailedAttempt(failure)
+        }
+      }
+    )
+  }
+}
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -2,7 +2,7 @@ package api
 
 import model._
 import play.api.libs.json._
-import play.api.libs.ws.WSClient
+import play.api.libs.ws.{WSClient, WSResponse}
 import utils.attempt.{Attempt, FailedAttempt, Failure}
 
 import scala.concurrent.ExecutionContext
@@ -10,7 +10,7 @@ import scala.util.control.NonFatal
 
 object Snyk {
 
-  def getSnykOrganisations(token: Token, wsClient: WSClient)(implicit ec:ExecutionContext) = {
+  def getSnykOrganisations(token: Token, wsClient: WSClient)(implicit ec:ExecutionContext): Attempt[WSResponse] = {
 
     val snykOrgUrl = "https://snyk.io/api/v1/orgs"
 
@@ -18,28 +18,28 @@ object Snyk {
       .addHttpHeaders("Authorization" -> s"token ${token.value}")
       .get
 
-    Attempt.fromFuture(futureResponse) { case NonFatal(e) => {
+    Attempt.fromFuture(futureResponse) { case NonFatal(e) =>
       val failure = Failure(e.getMessage, "Could not read organisations from Snyk", 502, None, Some(e))
       FailedAttempt(failure)
-    }}
+    }
   }
 
-  def getProjects(token: Token, id: String, wsClient: WSClient)(implicit ec:ExecutionContext) = {
-        val snykProjectsUrl = s"https://snyk.io/api/v1/org/${id}/projects"
+  def getProjects(token: Token, id: String, wsClient: WSClient)(implicit ec:ExecutionContext): Attempt[WSResponse] = {
+        val snykProjectsUrl = s"https://snyk.io/api/v1/org/$id/projects"
         val a = wsClient.url(snykProjectsUrl)
           .addHttpHeaders("Authorization" -> s"token ${token.value}")
           .get()
 
-    Attempt.fromFuture(a) { case NonFatal(e) => {
+    Attempt.fromFuture(a) { case NonFatal(e) =>
       val failure = Failure(e.getMessage, "Could not read projects from Snyk", 502, None, Some(e))
       FailedAttempt(failure)
-    }}
+    }
   }
 
-  def getProjectVulnerabilities(id: String, projects: List[SnykProject], token: Token, wsClient: WSClient)(implicit ec:ExecutionContext) = {
+  def getProjectVulnerabilities(id: String, projects: List[SnykProject], token: Token, wsClient: WSClient)(implicit ec:ExecutionContext): Attempt[List[WSResponse]] = {
     val projectVulnerabilityResponses = projects
       .map(project => {
-        val snykProjectUrl = s"https://snyk.io/api/v1/org/${id}/project/${project.id}/issues"
+        val snykProjectUrl = s"https://snyk.io/api/v1/org/$id/project/${project.id}/issues"
 
         val projectIssuesFilter = Json.obj(
           "filters" -> Json.obj(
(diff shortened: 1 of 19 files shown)


hq/app/views/snyk/snyk.scala.html
```diff
@@ -0,0 +1,62 @@
...
+            </thead>
+
+            <tbody>
+            @for(project <- results.sortWith((a, b) =>
+                a.high > b.high
+                        || a.high == b.high && (
+                        a.medium > b.medium
+                                || a.medium == b.medium && (
+                                a.low > b.low
+                                        || a.low == b.low && a.name < b.name)))) {
```
katebee: Yay! Sorting 🎉 

Not a blocker, but could this logic be moved to a separate logic file over in Scala land?
 Maybe logic/SnykDisplay.scala?

hq/conf/routes
```diff
@@ -9,11 +9,12 @@ GET        /iam                                       controllers.HQController.i
 GET        /iam/:accountId                            controllers.HQController.iamAccount(accountId)
 GET        /security-groups                           controllers.SecurityGroupsController.securityGroups
 GET        /security-groups/:accountId                controllers.SecurityGroupsController.securityGroupsAccount(accountId)
+GET        /snyk                                      controllers.SnykController.snyk
 GET        /login                                     controllers.AuthController.login
 GET        /loginError                                controllers.AuthController.loginError
 GET        /oauthCallback                             controllers.AuthController.oauthCallback
 GET        /logout                                    controllers.AuthController.logout
-GET        /documentation/                            controllers.UtilityController.documentation(file = "index")
+GET        /documentation                             controllers.UtilityController.documentation(file = "index")
```
katebee: Will this work? I tried dropping this `/` before but it seemed to break the linking to the subpages once on the documentation page...

aug24: Yes, it works; I did it to bring into line with all the others.

However I failed to update the header....  hold on.

aug24: Reverted.

hq/app/views/snyk/snyk.scala.html
aug24: Done.

shtukas:
+1

hq/app/views/header.scala.html
```diff
@@ -37,7 +40,10 @@ <h2 class="header grey-text text-lighten-5">Security HQ</h2>
                             <a href="/iam"><i class="material-icons black-text">vpn_key</i>Credentials audits</a>
                         </li>
                         <li class="main-nav">
-                            <a href="/documentation/"><i class="material-icons black-text">description</i>Documentation</a>
+                            <a href="/snyk"><i class="material-icons black-text">lock_open</i>Snyk</a>
+                        </li>
+                        <li class="main-nav">
+                            <a href="/documentation"><i class="material-icons black-text">description</i>Documentation</a>
```
katebee: Going to need to put the `/` back here tooo

> /documentation

shtukas: Let me do that :)

hq/app/logic/SnykDisplay.scala
```diff
@@ -0,0 +1,52 @@
...
+    })
+    Attempt.traverse(b)(a => a)
+  }
+
+  def labelProjects(projects: List[SnykProject], responses: List[SnykProjectIssues]) = {
+    projects.zip(responses).map(a => a._2.withName(a._1.name).withId(a._1.id))
+  }
+
+  def sortProjects(projects: List[SnykProjectIssues]): List[SnykProjectIssues] =
+    projects.sortWith((a, b) =>
```
adamnfish: `sortBy` is useful.

```scala
sortBy(spi => (spi.high, spi.medium, spi.low, spi.name))
```

Will sort by high and break ties with medium, then break ties with low, then break ties with name. This assumes `high`, `medium`, `low` and `name` are all themselves comparable, but give you're using `>` that must already be true.

hq/app/model/Serializers.scala
```diff
@@ -0,0 +1,41 @@
+package model
+
+import play.api.libs.functional.syntax._
+import play.api.libs.json.{Format, JsPath, Json, Reads}
+
+trait Serializers {
```
adamnfish: Why is this a `trait`?

hq/app/model/models.scala
```diff
@@ -135,3 +135,23 @@ case class MachineUser  (
...
+case class Organisation(value: String) extends AnyVal
+
+case class SnykOrg(name: String, id: String)
+
+case class SnykProject(name: String, id: String)
+
+case class SnykIssue(title: String, id: String, severity: String)
+
+case class SnykProjectIssues(name: String, id: String, ok: Boolean, vulnerabilities: List[SnykIssue])  {
+  def withName(name: String) = new SnykProjectIssues(name, this.id, this.ok, this.vulnerabilities)
```
adamnfish: These `with` methods are mixing the definition of the data structure with the program's logic. It's only used once and `copy` exists, why not just do that where you need it?

hq/conf/application.conf
```diff
@@ -27,10 +27,15 @@ stage = ${STAGE}
...
 
+stack = security
+app = security-hq
+
 hq {
   # default to empty list if the config is missing (for now)
   account = []
   accounts = ${?AWS_ACCOUNTS}
+  organisation = guardian
+  snyk_token = ${SNYK_TOKEN}
```
adamnfish: If the snyk token is a required config property what's parameter store being used for?

hq/app/api/Snyk.scala
```diff
@@ -0,0 +1,71 @@
...
+            "patched" -> "false"
+          )
+        )
+
+        wsClient.url(snykProjectUrl)
+          .addHttpHeaders("Authorization" -> s"token ${token.value}")
+          .post(projectIssuesFilter)
+
+      })
+    Attempt.traverse(projectVulnerabilityResponses)(
```
adamnfish: The syntax is a bit unusual for Scala here.

```scala
Attempt.traverse(projectVulnerabilityResponses) { projectVulnerabilityResponse =>
  ...
}
```
^- would be more canonical

hq/app/controllers/SnykController.scala
```diff
@@ -0,0 +1,75 @@
...
+import model._
+
+class SnykController(val config: Configuration)
+                    (implicit
+                     val ec: ExecutionContext,
+                     val wsClient: WSClient,
+                     val bodyParser: BodyParser[AnyContent],
+                     val controllerComponents: ControllerComponents,
+                     val assetsFinder: AssetsFinder,
+                     val awsClient: AWSSimpleSystemsManagement,
```
adamnfish: Why are these two implicit?

hq/app/controllers/SnykController.scala
```diff
@@ -0,0 +1,75 @@
...
+
+          results = SnykDisplay.labelProjects(projects, parsedVulnerabilitiesResponse)
+          sortedResult = SnykDisplay.sortProjects(results)
+        } yield Ok(views.html.snyk.snyk(sortedResult))
+      }
+    }
+
+  }
+
+  def getToken = configraun.getAsString("/snyk/token") match {
```
adamnfish: Not sure these belong in the controller (they aren't actions), but not the end of the world if you really want them here.

hq/app/logic/SnykDisplay.scala
```diff
@@ -0,0 +1,52 @@
...
+
+import scala.concurrent.ExecutionContext
+
+import model._
+
+object SnykDisplay extends Serializers {
+
+  def getOrganisationId(s: String, organisation: Organisation) = {
+    val id = for {
+      orglist <- (Json.parse(s) \ "orgs").asOpt[List[SnykOrg]]
```
adamnfish: It looks like there are some unhandled error cases in this, e.g. parsing failure. We want to try and avoid exceptions getting thrown so this might just need a try block.

hq/app/logic/SnykDisplay.scala
```diff
@@ -0,0 +1,52 @@
...
+import play.api.libs.json.Reads._
+import utils.attempt.{Attempt, FailedAttempt, Failure}
+
+import scala.concurrent.ExecutionContext
+
+import model._
+
+object SnykDisplay extends Serializers {
+
+  def getOrganisationId(s: String, organisation: Organisation) = {
```
adamnfish: It is super helpful to have return type annotations on non-trivial functions (we don't always have IntelliJ to help e.g. in Github)

hq/markdown/parameter-store.md
```diff
@@ -0,0 +1,57 @@
+# SSM Parameters
```
adamnfish: Probably worth mentioning that this uses configraun, which (I think) only applies to EC2 instances (not Lambdas).

hq/test/logic/SnykDisplayTest.scala
```diff
@@ -0,0 +1,209 @@
...
+
+}
+
+object SnykDisplayTest  {
+
+  val mockBadResponseWithMessage = s"""{\"error\": \"some nice error\"}"""
+
+  val mockBadResponseWithoutMessage = s"""{\"toughluck\": \"no use\"}"""
+
+  val mockGoodOrganisationResponse =
```
adamnfish: A nice technique is to put test JSON into the test resources directory and load it from there in your tests. Escaping JSON is not fun and it's a lot of noise for the test files.

Also, I may be mistaken but it looks like your " escapes aren't required? Within a `"""`, `"` is a literal.

hq/test/utils/attempt/AttemptValues.scala
```diff
@@ -42,5 +42,11 @@ trait AttemptValues extends Matchers {
         _ => false
       )
     }
+    def getFailedAttempt()(implicit ec: ExecutionContext): Option[FailedAttempt] = {
```
adamnfish: Why would this return an `Option`? I think this should work more like the `value()` method above (which is in turn inspired by scalatest conventions like `EitherValues`. If you want a failed attempt then it's a test failure if you didn't get one (rather than a None that you then need to handle).

You're calling `.get` anyway so this way would also give much better test error messages.

hq/app/controllers/SnykController.scala
adamnfish: I've also noticed that other configraun happens at start time. That seems very reasonable, why aren't these done the same way?

hq/app/controllers/SnykController.scala
```diff
@@ -0,0 +1,75 @@
...
+    case Left(a:ConfigraunError) => {
+      val failure = Failure(a.message, "Could not read Snyk token from aws parameter store", 500, None, Some(a.e))
+      Attempt.fromOption(None, FailedAttempt(failure))
+    }
+    case Right(a:String) => Attempt.fromOption(Some(new Token(a)), null)
```
adamnfish: `null`?!

Why isn't this just `Attempt.Right(new Token(a))`?

hq/app/controllers/SnykController.scala
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  def getOrganisation = configraun.getAsString("/snyk/organisation") match {
+    case Left(a:ConfigraunError) => {
+      val failure = Failure(a.message, "Could not read Snyk organisation from aws parameter store", 500, None, Some(a.e))
+      Attempt.fromOption(None, FailedAttempt(failure))
```
adamnfish: Why not `Attempt.Left(failure)`?

hq/app/api/Snyk.scala
aug24: Done.

hq/app/controllers/SnykController.scala
aug24: The awsClient shouldn't be there at all.  The configraun client is ... arguable, I suppose.

It could be a 'proper' variable, but I imagined it would be sent everywhere and had some idea that would be a good reason for making it implicit.  I've made it 'proper' instead.

hq/app/controllers/SnykController.scala
aug24: The stack/stage/app work happens at start time because configraun needs them to initialise.

I would usually rather have the snyk-specific code in a snyk-specific class, but I notice that the SnykController is only instantiated when /snyk is requested, so perhaps it would be better to force it up front in the AppController.

Discuss?  / Confirm? pls

hq/app/controllers/SnykController.scala
aug24: Hadn't found that syntax.  Done.

hq/app/controllers/SnykController.scala
aug24: Ditto

hq/app/logic/SnykDisplay.scala
aug24: Done.

hq/app/logic/SnykDisplay.scala
aug24: Thank you!

aug24: Man, that looks better!

hq/app/model/Serializers.scala
aug24: Now this one I would like to discuss.  I appear to need correctly name implicits, in my SnykDisplay object.  I wanted to have a separate class for the serializers as they aren't likely to change and are just 'noise' from the PoV of SnykDisplay.

hq/conf/application.conf
aug24: It isn't any more, that should have been removed.

hq/markdown/parameter-store.md
aug24: It doesn't _have_ to use Configraun.

PR #66, opened 2018-01-25
Sorting for the credentials reports, and the account names

katebee:
#### 👉 Sort IAM credentials reports by number of issues

The all accounts IAM credentials report is currently sorted according to the account name...

This PR will make SHQ order the the accounts by the severity and number of alerts:

```
# errors > # warnings > account name
```

picture 169

#### 👉 Sorting accounts within accounts!

Users were once listed in alphabetical order, now the ones with severe alerts will appear at the top of their section! \* 

##### \* see additional notes

picture 175

## What is the value of this?

Metrics and motivation
 

Nope

## Any additional notes?

The IAM body template currently builds from two distinct lists (humanUsers and machineUsers). Some refactoring of how the users are stored and the table is generated will be required before they can be sorted _together_

These changes only apply to the All Accounts page, since this is a good place to get an overview of how things are doing; I have left the original sorting in place on the individual account pages.

### DIFF
--- a/hq/app/controllers/HQController.scala
+++ b/hq/app/controllers/HQController.scala
@@ -8,7 +8,7 @@ import play.api.libs.ws.WSClient
 import play.api.mvc._
 import services.CacheService
 import utils.attempt.PlayIntegration.attempt
-import logic.ReportDisplay.sortByReportSummary
+import logic.ReportDisplay.sortAccountsByReportSummary
 
 
 import scala.concurrent.ExecutionContext
@@ -25,7 +25,7 @@ class HQController (val config: Configuration, cacheService: CacheService)
 
   def iam = authAction {
     val accountsAndReports = cacheService.getAllCredentials()
-    val sortedAccountsAndReports = sortByReportSummary(accountsAndReports.toList)
+    val sortedAccountsAndReports = sortAccountsByReportSummary(accountsAndReports.toList)
     Ok(views.html.iam.iam(sortedAccountsAndReports))
   }
 
--- a/hq/app/controllers/HQController.scala
+++ b/hq/app/controllers/HQController.scala
@@ -8,6 +8,8 @@ import play.api.libs.ws.WSClient
 import play.api.mvc._
 import services.CacheService
 import utils.attempt.PlayIntegration.attempt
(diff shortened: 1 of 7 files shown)


hq/app/views/iam/iamCredReportBody.scala.html
```diff
@@ -29,13 +29,12 @@
                     <td>@views.html.fragments.allKeyStatus(cred.key1Status, cred.key2Status)</td>
                     <td>@views.html.fragments.mfa(cred.hasMFA)</td>
                     <td>
-                        <span>
-                            @ReportDisplay.toDayString(cred.lastActivityDay)
-                        </span>
+                        <span>@ReportDisplay.toDayString(cred.lastActivityDay)</span>
```
katebee: Doing this for consistency with the other template. I think this table and the use of fragments could be refactored a bit, and I would like to allow machineUsers and humanUsers to sort together.

Targets for a future PR! 😁

hq/app/logic/ReportDisplay.scala
```diff
@@ -83,4 +85,24 @@ object ReportDisplay {
...
+  }
+
+  private def sortByStatus(u1: AwsUser, u2: AwsUser) = {
+    def statusCode(status: ReportStatus): Int = status match {
+      case Red => 2
+      case Amber => 1
+      case _ => 0
+    }
+    statusCode(u1.reportStatus) > statusCode(u2.reportStatus)
+  }
```
katebee: Not sure where these methods should live, but pretty sure it is not here! 😅 

Suggestions welcome!

hq/app/controllers/HQController.scala
```diff
@@ -8,6 +8,8 @@ import play.api.libs.ws.WSClient
 import play.api.mvc._
 import services.CacheService
 import utils.attempt.PlayIntegration.attempt
+import logic.ReportDisplay.sortAccountsByReportSummary
```
katebee: Mentioned below, but sortAccountsByReportSummary should probably not be in ReportDisplay...

shtukas: +1

PR #70, opened 2018-01-29
User icons and helper modals for the IAM credentials reports

katebee:
#### 👉 More information on the Credentials report pages

- Icon to indicate the inferred use type (human/machine)
- Help icon after the alert warning, to trigger display of the help modal

picture 177

#### 👉 Popup modals where verbose details and links to AWS documentation can be placed

Appears on the all accounts and individual accounts pages and can be slightly different depending on the user type. Also contains some information about how user types are inferred.

picture 172

#### 👉 Code maintenance and small fixes

- deleting small template fragment and moving lines to main template
- deleting unused css, and reordering css slightly
- removing unused data attributes from the html
- tweaks wording of the no active access key warning

## What is the value of this?

Providing more information and support when SHQ users are looking at the IAM credentials report

Nope 

## Any additional notes?

Wording and content of the modals can be iterated upon. 😄 

Flagging inactive users is a helpful way of encouraging old credentials to be deleted. I have arbitrarily set the age check to be a year, however I feel it could be shorter. We may want to reduce this, and update the help modal messaging?

### DIFF
--- a/hq/public/stylesheets/main.css
+++ b/hq/public/stylesheets/main.css
@@ -33,11 +33,6 @@ nav {
   box-shadow: none;
 }
 
-.is-hidden,
-[hidden] {
-  display: none !important;
-}
-
 .space-between {
   justify-content: space-between;
 }
@@ -50,6 +45,10 @@ nav {
 
 .link_new_tab {
   vertical-align: bottom;
+}
+
+.large-icon {
+  font-size: 36px !important;
 }
 
 /* navigation and header */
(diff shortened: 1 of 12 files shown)


hq/public/stylesheets/main.css
```diff
@@ -302,6 +301,10 @@ table.iam-report__table th {
   box-sizing: initial;
 }
 
+.iam-modal__trigger {
+  height: 24px;
```
adamnfish: Make sure MaterializeCSS can't already do what you're after here.

http://materializecss.com/icons.html

PR #64, opened 2018-01-25
Add Snyk report and Configraun

aug24:
## What does this change?

Adds a Snyk Report page
Fetches some configuration via Configraun configuration engine

## What is the value of this?

Exposes the results of snyk testing for vulnerabilities, including vulnerabilities found after the 
initial test.

Configraun use removes a 'secret' from the configuration file and stores it in AWS Parameter Store.
We should be moving towards this approach generally.  This implementation provides a dogfood implementation for others to use.

## Will this require CloudFormation and/or updates to the AWS StackSet?

Yes.
Template changes are included.

## Any additional notes?

No

PR #72, opened 2018-01-30
Add Snyk report and Configraun (Attempt 3)

aug24:
## What does this change?

Adds a new page showing the Snyk report; fetches Snyk token for same from AWS parameter store via configraun.

## What is the value of this?

Expose Snyk vulnerability reports.
Dogfood approach to using SSN Parameter Store.

Yes.

### DIFF
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -1,7 +1,11 @@
+import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagement
+import com.gu.configraun.Configraun
+import com.gu.configraun.aws.AWSSimpleSystemsManagementFactory
+import com.gu.configraun.models._
 import controllers._
 import filters.HstsFilter
 import play.api.ApplicationLoader.Context
-import play.api.BuiltInComponentsFromContext
+import play.api.{BuiltInComponentsFromContext, Logger}
 import play.api.libs.ws.WSClient
 import play.api.libs.ws.ahc.AhcWSComponents
 import play.api.mvc.{AnyContent, BodyParser, ControllerComponents}
@@ -24,6 +28,29 @@ class AppComponents(context: Context)
     csrfFilter,
     new HstsFilter()
   )
+  implicit val awsClient: AWSSimpleSystemsManagement = AWSSimpleSystemsManagementFactory("eu-west-1", "security")
+
+  val configraun: Configuration = {
+
+    configuration.getOptional[String]("stage") match {
+      case Some("DEV") =>
+        val stack = configuration.get[String]("stack")
+        val app = configuration.get[String]("app")
+        val stage = "DEV"
+        Configraun.loadConfig(Identifier(Stack(stack), App(app), Stage.fromString(stage).get)) match {
+          case Left(a) =>
+            Logger.error(s"Unable to load Configraun configuration from AWS (${a.message})")
+            sys.exit(1)
+          case Right(a) => a
+        }
+      case _ => Configraun.loadConfig match {
+        case Left(a) =>
+          Logger.error(s"Unable to load Configraun configuration from AWS tags (${a.message})")
+          sys.exit(1)
+        case Right(a: com.gu.configraun.models.Configuration) => a
+      }
+    }
+  }
 
   val cacheService = new CacheService(configuration, applicationLifecycle, environment)
 
@@ -31,6 +58,7 @@ class AppComponents(context: Context)
     httpErrorHandler,
     new HQController(configuration, cacheService),
     new SecurityGroupsController(configuration, cacheService),
+    new SnykController(configuration, configraun),
     new AuthController(environment, configuration),
     new UtilityController(),
     assets
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -11,13 +11,10 @@ import scala.util.control.NonFatal
 object Snyk {
 
   def getSnykOrganisations(token: SnykToken, wsClient: WSClient)(implicit ec:ExecutionContext): Attempt[WSResponse] = {
-
     val snykOrgUrl = "https://snyk.io/api/v1/orgs"
-
     val futureResponse = wsClient.url(snykOrgUrl)
       .addHttpHeaders("Authorization" -> s"token ${token.value}")
       .get
-
     Attempt.fromFuture(futureResponse) { case NonFatal(e) =>
       val failure = Failure(e.getMessage, "Could not read organisations from Snyk", 502, None, Some(e))
       FailedAttempt(failure)
@@ -29,7 +26,6 @@ object Snyk {
     val a = wsClient.url(snykProjectsUrl)
         .addHttpHeaders("Authorization" -> s"token ${token.value}")
         .get()
-
     Attempt.fromFuture(a) { case NonFatal(e) =>
       val failure = Failure(e.getMessage, "Could not read projects from Snyk", 502, None, Some(e))
       FailedAttempt(failure)
@@ -40,7 +36,6 @@ object Snyk {
     val projectVulnerabilityResponses = projects
       .map(project => {
         val snykProjectUrl = s"https://snyk.io/api/v1/org/$id/project/${project.id}/issues"
-
         val projectIssuesFilter = Json.obj(
           "filters" -> Json.obj(
             "severity" -> JsArray(List(
@@ -53,7 +48,6 @@ object Snyk {
             "patched" -> "false"
           )
         )
-
         wsClient.url(snykProjectUrl)
           .addHttpHeaders("Authorization" -> s"token ${token.value}")
           .post(projectIssuesFilter)
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -1,0 +1,71 @@
+package api
+
+import model._
+import play.api.libs.json._
+import play.api.libs.ws.{WSClient, WSResponse}
+import utils.attempt.{Attempt, FailedAttempt, Failure}
+
+import scala.concurrent.ExecutionContext
+import scala.util.control.NonFatal
+
+object Snyk {
+
+  def getSnykOrganisations(token: SnykToken, wsClient: WSClient)(implicit ec:ExecutionContext): Attempt[WSResponse] = {
+
+    val snykOrgUrl = "https://snyk.io/api/v1/orgs"
+
+    val futureResponse = wsClient.url(snykOrgUrl)
+      .addHttpHeaders("Authorization" -> s"token ${token.value}")
+      .get
+
+    Attempt.fromFuture(futureResponse) { case NonFatal(e) =>
+      val failure = Failure(e.getMessage, "Could not read organisations from Snyk", 502, None, Some(e))
+      FailedAttempt(failure)
+    }
+  }
+
+  def getProjects(token: SnykToken, id: String, wsClient: WSClient)(implicit ec:ExecutionContext): Attempt[WSResponse] = {
+    val snykProjectsUrl = s"https://snyk.io/api/v1/org/$id/projects"
+    val a = wsClient.url(snykProjectsUrl)
+        .addHttpHeaders("Authorization" -> s"token ${token.value}")
+        .get()
+
+    Attempt.fromFuture(a) { case NonFatal(e) =>
+      val failure = Failure(e.getMessage, "Could not read projects from Snyk", 502, None, Some(e))
+      FailedAttempt(failure)
+    }
+  }
+
+  def getProjectVulnerabilities(id: String, projects: List[SnykProject], token: SnykToken, wsClient: WSClient)(implicit ec:ExecutionContext): Attempt[List[WSResponse]] = {
+    val projectVulnerabilityResponses = projects
+      .map(project => {
+        val snykProjectUrl = s"https://snyk.io/api/v1/org/$id/project/${project.id}/issues"
+
+        val projectIssuesFilter = Json.obj(
+          "filters" -> Json.obj(
+            "severity" -> JsArray(List(
+              JsString("high"), JsString("medium"), JsString("low")
+            )),
+            "types" -> JsArray(List(
+              JsString("vuln")
+            )),
+            "ignored" -> "false",
+            "patched" -> "false"
+          )
+        )
+
+        wsClient.url(snykProjectUrl)
+          .addHttpHeaders("Authorization" -> s"token ${token.value}")
+          .post(projectIssuesFilter)
+
+      })
+    Attempt.traverse(projectVulnerabilityResponses) {
+      projectVulnerabilityResponse =>
+        Attempt.fromFuture(projectVulnerabilityResponse) {
+          case NonFatal(e) =>
+            val failure = Failure(e.getMessage, "Could not read project vulnerabilities from Snyk", 502, None, Some(e))
+            FailedAttempt(failure)
+        }
+    }
+  }
+}
--- a/hq/app/controllers/SnykController.scala
+++ b/hq/app/controllers/SnykController.scala
@@ -1,0 +1,71 @@
+package controllers
+
+
+import auth.SecurityHQAuthActions
+import play.api.Configuration
+import play.api.libs.ws.WSClient
+import play.api.mvc._
+
+import scala.concurrent.ExecutionContext
+import logic.SnykDisplay
+import api.Snyk
+import com.gu.configraun.Errors.ConfigraunError
+import utils.attempt.{Attempt, Failure}
+import utils.attempt.PlayIntegration.attempt
+import model._
+
+class SnykController(val config: Configuration, val configraun: com.gu.configraun.models.Configuration)
+                    (implicit
+                     val ec: ExecutionContext,
+                     val wsClient: WSClient,
+                     val bodyParser: BodyParser[AnyContent],
+                     val controllerComponents: ControllerComponents,
+                     val assetsFinder: AssetsFinder)
+  extends BaseController  with SecurityHQAuthActions
+{
+
+  def snyk: Action[AnyContent] = {
+
+    Action.async {
+      attempt {
+        for {
+          token <- getSnykToken
+          requiredOrganisation <- getSnykOrganisation
+
+          organisationResponse <- Snyk.getSnykOrganisations(token, wsClient)
+          organisation <- SnykDisplay.getOrganisation(organisationResponse.body, requiredOrganisation)
+
+          projectResponse <- Snyk.getProjects(token, organisation.id, wsClient)
+          projects <- SnykDisplay.getProjectIdList(projectResponse.body)
+          labelledProjects = SnykDisplay.labelOrganisations(projects, organisation)
+
+          vulnerabilitiesResponse <- Snyk.getProjectVulnerabilities(organisation.id, projects, token, wsClient)
+          vulnerabilitiesResponseBodies = vulnerabilitiesResponse.map(a => a.body)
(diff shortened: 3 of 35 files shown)


package.json
```diff
@@ -22,6 +22,8 @@
     "eslint": "eslint --max-warnings 0 --format codeframe --ignore-pattern '**/*.min.js' hq/public/javascripts/",
     "linter-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
     "prettier": "prettier -o --write './**/*.{js,css}'",
-    "sass-lint": "sass-lint -v -q hq/public/stylesheets/*.css"
+    "sass-lint": "sass-lint -v -q hq/public/stylesheets/*.css",
+    "snyk-test": "PATH=/path/to/sbt/executable/folder:$PATH snyk test --debug --org=guardian --json --file=build.sbt",
+    "snyk-monitor": "PATH=/path/to/sbt/executable/folder:$PATH snyk monitor --debug --org=guardian --file=build.sbt"
```
katebee: Should we have Snyk listed in the devDependencies? I may have missed the convo - does Snyk ave to be installed globally, or can we use it as a local installation?

hq/app/views/snyk/snyk.scala.html
```diff
@@ -0,0 +1,54 @@
...
+                    </td>
+                    <td>
+                        <a class="black-text" href="https://snyk.io/org/@projectIssues.project.get.organisation.get.name/project/@projectIssues.project.get.id/?disclosure=all&severity=high">@projectIssues.high</a>
+                    </td>
+                    <td>
+                        <a class="black-text" href="https://snyk.io/org/@projectIssues.project.get.organisation.get.name/project/@projectIssues.project.get.id/?disclosure=all&severity=medium">@projectIssues.medium</a>
+                    </td>
+                    <td>
+                        <a class="black-text" href="https://snyk.io/org/@projectIssues.project.get.organisation.get.name/project/@projectIssues.project.get.id/?disclosure=all&severity=low">@projectIssues.low</a>
+                    </td>
```
katebee: Black text on cards and within nav is okay (and is the current pattern), since most users will expect those areas to be clickable. However, within a table or body text, links should be visually distinct (at least a different colour, ideally also with text decoration). https://github.com/guardian/security-hq/pull/64#discussion_r163819446

I feel that we should reuse the styling/design from the credentials report here:

picture 178

Not a blocker - happy for this to be a change for a future PR

hq/markdown/snyk.md
```diff
@@ -20,6 +24,52 @@ Snyk then requires an authentication token to be set up.  Github auth is almost
...
+    "snyk-test": "PATH=/path/to/sbt/executable/folder:$PATH snyk test --debug --org=guardian --json --file=build.sbt",
+    "snyk-monitor": "PATH=/path/to/sbt/executable/folder:$PATH snyk monitor --debug --org=guardian --file=build.sbt",
+  }
+```
+
+### Node Builds
+
+```
+  "devDependencies": {
+    "snyk": "1.68.0"
```
katebee: Minor: is there a reason that the version numbers are different?

package.json
katebee: If we do add Snyk to the devDeps

Can the update to the yarn.lock file also be included?

https://yarnpkg.com/lang/en/docs/yarn-lock/

It should auto-generate based on changes to package.json when you run `yarn install`

hq/markdown/snyk.md
```diff
@@ -7,7 +7,11 @@ vulnerabilities in your application.
 
 ## Installing
 
-Snyk requires node, and `npm`.  To install on a mac, please use:
+Snyk requires node, and `npm`.
+
+### Developer Install
+
+To install on a mac, please use:
```
katebee: Brew does not come as standard on MacOS therefore it may be worth adding a line about getting Homebrew?

OS X users may want to use Homebrew

hq/markdown/snyk.md
```diff
@@ -7,7 +7,11 @@ vulnerabilities in your application.
 
 ## Installing
 
-Snyk requires node, and `npm`.  To install on a mac, please use:
+Snyk requires node, and `npm`.
```
katebee: Personal preference, however I prefer Yarn... (and have set up this project to use Yarn) Maybe:

> Snyk requires node, and a Node package manager such as `npm` or `Yarn`

I realise the build servers are currently using NPM, however I am hoping to remove this inconsistency

hq/app/views/snyk/snyk.scala.html
```diff
@@ -0,0 +1,54 @@
...
+
+@main(List("Snyk")) { @* Header *@
+    <div class="container">
+        <div class="row">
+            <h1 class="header center-on-small-only grey-text text-lighten-5">Security HQ</h1>
+            <div class="col s12 m9">
+                <h4 class="grey-text text-darken-4 center-on-small-only">Snyk Results</h4>
+            </div>
+        </div>
+    </div>
```
katebee: There was a bit of a design change with the header and subheader, so the h1 should no longer be required. We will probably want something similar to the docs page subheader:

```
  <div class="hq-sub-header">
      <div class="container hq-sub-header__row">
          <div class="hq-sub-header__name">
              <h4 class="header light grey-text text-lighten-5">Snyk Results</h4>
          </div>
      </div>
  </div>
```

Not a blocker, can be target for subsequent PR 😄

hq/app/controllers/SnykController.scala
```diff
@@ -0,0 +1,71 @@
...
+
+          results = SnykDisplay.labelProjects(labelledProjects, parsedVulnerabilitiesResponse)
+          sortedResult = SnykDisplay.sortProjects(results)
+        } yield Ok(views.html.snyk.snyk(sortedResult))
+      }
+    }
+
+  }
+
+  def getSnykToken: Attempt[SnykToken] = configraun.getAsString("/snyk/token") match {
```
adamnfish: I mentioned on the last PR that these functions perhaps shouldn't be in the controller (they aren't actions). Do we need them at all? If this config is required why don't we fail at start time rather than at runtime whenever this is requested?

hq/app/AppComponents.scala
```diff
@@ -24,13 +28,37 @@ class AppComponents(context: Context)
     csrfFilter,
     new HstsFilter()
   )
+  implicit val awsClient: AWSSimpleSystemsManagement = AWSSimpleSystemsManagementFactory("eu-west-1", "security")
+
+  val configraun: Configuration = {
+
+    configuration.getOptional[String]("stage") match {
+      case Some("DEV") =>
```
adamnfish: It looks like this means we all have to use the exact same config in DEV. That means we can't (for example) have different accounts configured to match the credentials we have available. It also means we can't try things by changing config without it affecting others.

hq/app/logic/SnykDisplay.scala
```diff
@@ -0,0 +1,50 @@
...
+import scala.concurrent.ExecutionContext
+
+import model._
+import model.Serializers._
+
+object SnykDisplay {
+
+  def getOrganisation(s: String, requiredOrganisation: SnykOrganisationName): Attempt[SnykOrganisation] = {
+    val organisation = for {
+      orglist <- (Json.parse(s) \ "orgs").asOpt[List[SnykOrganisation]]
```
adamnfish: These all still throw runtime exceptions is the JSON is not well-formed.

hq/app/logic/SnykDisplay.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  def labelOrganisations(projects: List[SnykProject], snykOrg: SnykOrganisation): List[SnykProject] = {
+    projects.map(project => project.copy(organisation = Some(snykOrg)))
+  }
+
+  def labelProjects(projects: List[SnykProject], responses: List[SnykProjectIssues]): List[SnykProjectIssues] = {
+    projects.zip(responses).map(a => a._2.copy(project = Some(a._1)))
+  }
+
+  def sortProjects(projects: List[SnykProjectIssues]): List[SnykProjectIssues] =
```
adamnfish: This should have a test instead of a comment

hq/app/model/Serializers.scala
```diff
@@ -0,0 +1,30 @@
...
+import play.api.libs.functional.syntax._
+import play.api.libs.json.{Format, JsPath, Json, Reads}
+
+object Serializers {
+
+  implicit val snykErrorFormat: Reads[SnykError] = Json.reads[SnykError]
+
+  implicit val snykOrgFormat: Reads[SnykOrganisation] = Json.reads[SnykOrganisation]
+
+  implicit val snykProjectFormat: Reads[SnykProject] = (
```
adamnfish: Are you sure the macro won't work here as well? Options in the target case class denote optional fields, it'll be `Some` if present, `None` if absent.

hq/markdown/snyk.md
```diff
@@ -185,11 +252,14 @@ The worst case scenario is that the library is still vulnerable, has no alternat
 may wish to build and release anyway, ideally reporting the situation to a risk register. This is most likely to happen when
 a new vulnerability is discovered and the library publisher has not had chance to respond.
 
-This is cleanly achieved by creating a `.snyk` file which can act both as an exemption and the risk register itself.
+For private repositories, this is cleanly achieved by creating a `.snyk` file which can act both as an exemption and the risk register itself.
 Please try to give meaningful reasons for allowing the exemption.
 
 The `.snyk` file can then be added to the repository and the build should continue.
 
+__However, for public repositories, this would mean that the vulnerability is effectively advertised right in
```
adamnfish: Do we care about this? The information is public anyway and if it means teams are more aware of it (and fix it more quickly) that is arguably better?

Be worth discussing this in the team, and maybe with InfoSec as well.

hq/markdown/temporary-ssh.md
```diff
@@ -0,0 +1,29 @@
+# Temporary SSH
+
+In order to move towards use of temporary credentials, a facility has been provided to locally generate a keypair and push the
```
adamnfish: This is very waffly. Attention is scarce, can we try and be as concise as possible in the docs?

hq/test/logic/SnykDisplayTest.scala
```diff
@@ -0,0 +1,291 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class SnykDisplayTest extends FreeSpec with Matchers with AttemptValues {
+
+  "find organisation" in {
+    val organisation = SnykDisplay.getOrganisation(SnykDisplayTest.mockGoodOrganisationResponse, SnykOrganisationName("guardian"))
+    organisation.value shouldBe SnykOrganisation("guardian", "1111111111")
+  }
+
+  "fail to find organisation (nice)" in {
```
adamnfish: What do (nice) and (not nice) mean? Will it be clear to someone else that comes to this file later when the tests are failing?

hq/markdown/snyk.md
aug24: Done.

hq/markdown/snyk.md
aug24: Done.

hq/test/logic/SnykDisplayTest.scala
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    "find project 2 in project list" in {
+      val projects = SnykDisplay.getProjectIdList(SnykDisplayTest.mockGoodProjectResponse)
+      projects.value().exists(p => p.name == "project2") shouldBe true
+    }
+
+  }
+
+  "fail to find project list" - {
```
adamnfish: Should this not be part of the group above?

hq/test/logic/SnykDisplayTest.scala
```diff
@@ -0,0 +1,291 @@
+package logic
+
+import model._
+import org.scalatest.{FreeSpec, Matchers}
+import utils.attempt.AttemptValues
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class SnykDisplayTest extends FreeSpec with Matchers with AttemptValues {
+
+  "find organisation" in {
```
adamnfish: Can we group these tests?

hq/test/logic/SnykDisplayTest.scala
```diff
@@ -0,0 +1,291 @@
...
+      projects.isFailedAttempt() shouldBe true
+    }
+    "fail to find vulnerability list (not nice) - has message" in {
+      val projects = SnykDisplay.parseProjectVulnerabilities(List(SnykDisplayTest.mockBadResponseWithoutMessage))
+      projects.getFailedAttempt.failures.head.friendlyMessage shouldBe """Could not read Snyk response ({"toughluck": "no use"})"""
+    }
+  }
+
+  "label organisation" - {
+    "label organisation - first id" in {
```
adamnfish: The test message should say what property we're checking so that when it fails the developer sees something helpful.

hq/test/logic/SnykDisplayTest.scala
```diff
@@ -0,0 +1,291 @@
...
+     """.stripMargin
+
+  val goodOrganisation =
+    SnykOrganisation("name0", "id0")
+
+  val goodProjects = List(
+    SnykProject("name1", "id1", None),
+    SnykProject("name2", "id2", None)
+  )
+  val goodVulnerabilities = List(
```
adamnfish: This is only used in one block of tests so I don't see a reason for it to be hidden away down here? Maybe stick it in the block so the tests can share it? Generally, better to have the test setup where it's used so there are fewer places to look to comprehend the tests.

hq/app/api/Snyk.scala
```diff
@@ -0,0 +1,71 @@
...
+import play.api.libs.ws.{WSClient, WSResponse}
+import utils.attempt.{Attempt, FailedAttempt, Failure}
+
+import scala.concurrent.ExecutionContext
+import scala.util.control.NonFatal
+
+object Snyk {
+
+  def getSnykOrganisations(token: SnykToken, wsClient: WSClient)(implicit ec:ExecutionContext): Attempt[WSResponse] = {
+
```
adamnfish: Lots of whitespace! Do we need it all?

package.json
aug24: Done.

hq/app/AppComponents.scala
aug24: It means the DEV config _for objects fetched via Configraun_ will be universal, yes.

Currently that is only the Snyk token and Organisation name.

hq/app/api/Snyk.scala
aug24: Removed.

hq/test/logic/SnykDisplayTest.scala
adamnfish: (+general grouping in this file)

hq/app/logic/SnykDisplay.scala
aug24: Done.

hq/app/logic/SnykDisplay.scala
aug24: Done.

hq/test/logic/SnykDisplayTest.scala
aug24: Done.

hq/test/logic/SnykDisplayTest.scala
aug24: Done.

hq/test/logic/SnykDisplayTest.scala
aug24: Done

hq/app/controllers/SnykController.scala
aug24: I don't like moving them out of the controller into AppComponents, as I think AppComponents shouldn't need to know what configuration elements the SnykController needs.  However, I don't like runtime failures either.

Suggestions gratefully received.

hq/app/model/Serializers.scala
aug24: They aren't returned as part of the json response (they had to be specified in the request).  They will definitively always be None.  The values are then 'enriched' on to the returned object after parsing.

hq/test/logic/SnykDisplayTest.scala
aug24: It does ... it checks the first id.

hq/test/logic/SnykDisplayTest.scala
aug24: Done.  In fact, the whole lot eventually found better homes and the addition object went away.

hq/markdown/temporary-ssh.md
aug24: Done.

hq/app/controllers/SnykController.scala
adamnfish: AppComponents is responsible for starting the application so by that measure it would need to be aware of what the application requires (and thus, what the SnykController needs).

Leave it as-is for now and we can explore this together when we're both in the office. I do think the config changes here will need a bit of thought either way, so fine to tackle this as part of that process.

hq/test/logic/SnykDisplayTest.scala
adamnfish: The failure message for this test will be precisely:

    label organisation label organisation - first id

The duplication of label organisation is unhelpful and there's not much info beyond that.

You know what that means because you wrote it, but the person dealing with a failing test will not see the implementation, only the above string. What do you think they will be able to infer from it?

What property of the id are you testing? In this case you are checking that the resulting id is equal to `id0`. What's the significance of `id0`? It's not in the `goodProjects` test data so where is it coming from and why?

The message doesn't need to be an essay, but it should help guide the understanding of someone that has to fix the project. I like to think of tests as compiled comments, if you were going to write an inline comment on the relevant line of code, what would it say?

hq/app/logic/SnykDisplay.scala
```diff
@@ -0,0 +1,74 @@
...
+    Attempt.fromOption(organisation, FailedAttempt(f))
+  }
+
+  private def parseJsonToError(s: String): SnykError = try {
+    Json.parse(s).asOpt[SnykError].getOrElse(SnykError(s))
+  } catch {
+    case _: JsonParseException => SnykError(s)
+  }
+
+  private def parseJsonToOrganisationList(s: String): Option[List[SnykOrganisation]] = try {
```
adamnfish: If this returned an Attempt you'd have the opportunity to give a more precise error message?

hq/app/logic/SnykDisplay.scala
```diff
@@ -0,0 +1,74 @@
...
+    val organisation = for {
+      orglist <- parseJsonToOrganisationList(s)
+      org <- orglist.find(_.name == requiredOrganisation.value)
+    } yield org
+    val error = parseJsonToError(s)
+    val f = Failure(s"Unable to find organisation from $s", s"Could not read Snyk response (${error.error})", 502, None, None)
+    Attempt.fromOption(organisation, FailedAttempt(f))
+  }
+
+  private def parseJsonToError(s: String): SnykError = try {
```
adamnfish: Why do we parse the JSON twice? (Same pattern below as well)

hq/app/logic/SnykDisplay.scala
adamnfish: (which might eliminate the need to parse the JSON twice)

hq/markdown/snyk.md
aug24: Toned down.

hq/app/logic/SnykDisplay.scala
aug24: It parses it once in an attempt to get the data from a correct response, then once to get an error message from a failed response.  If all fails, it will just use the entire response body.

PR #73, opened 2018-01-31
Support for filtering suppressed SGs, add filter options to views

katebee:
## What does this change?

Currently SHQ is not using the `isSuppressed` value from the TrustedAdvisor (TA) result.  If a Security Group warning is excluded/suppressed via the TA console `isSuppressed` will be `true`

> isSuppressed specifies whether the AWS resource was ignored by Trusted Advisor because it was marked as suppressed by the user.

This is a MVP at adding some filtering and visual cues to accommodate `isSuppressed`

#### 👉 Visually distinguish Security Groups when they are suppressed

- Add icon and tooltip to the card title
- Make the card background grey

picture 190

#### 👉 All accounts page

- Add optional badges to show the number of excluded/suppressed Security Groups (SGs)
- Adds **permanent** card panel, with explainer text and checkboxes to toggle which SGs to display
- Defaults to showing flagged SGs with the ignored SGs hidden

picture 191

#### 👉 Individual account pages

- Adds **conditional** card panel about ignored Security Groups (SGs) **if** there are ignored SGs
- Defaults to showing flagged SGs with the ignored SGs hidden

picture 189

## What is the value of this?

SHQ probably still wants oversight of suppressed reports, however, suppressed reports add noise to the SGs list and are probably not helpful to individual teams.

Nope

## Any additional notes?

This PR is interested in splitting flaggedSgs into two subsets that I have called `suppressed` and `flagged`. Arguably flaggedSgs should then be called something else, or maybe these subsets should be `excluded` and `included`...Naming things is hard. ¯\\_(ツ)\_/¯

Accounts are still sorted according to the total number of SG reports (including suppressed). We may want to change this (target for future PR).

#### I can confirm that checkboxes are 100% more aesthetically pleasing than using toggles here

picture 187

### DIFF
--- a/hq/public/stylesheets/main.css
+++ b/hq/public/stylesheets/main.css
@@ -1,6 +1,7 @@
 /* COLOUR SCHEME
 DARK GREY:  #212121
 TEAL:       #009688
+LIGHT GREY: #F5F5F5
 NOT-WHITE:  #F7F0F0
 HIGHLIGHT:  #8AF3FF
 
@@ -200,8 +201,16 @@ nav {
 
 /* security group pages */
 
+.sg-filter__label {
+  margin-left: 20px;
+}
+
 .sg-header__name {
-  margin-right: auto;
+  flex-grow: 2;
+}
+
+.sg-header span.sg-header__badge {
+  margin-left: 10px;
 }
 
 .sg-container {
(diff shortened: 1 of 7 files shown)


hq/public/stylesheets/main.css
```diff
@@ -200,8 +201,16 @@ nav {
...
+  margin-left: 20px;
+}
+
 .sg-header__name {
-  margin-right: auto;
+  flex-grow: 2;
+}
+
+.sg-header span.sg-header__badge {
+  margin-left: 10px;
```
katebee: Boo specificity! 👎 
Unfortunately, this is necessary to win a fight with materializecss because multiple badges should not collide (IMHO)

shtukas: 😄

shtukas:
Reviewed it with Kate on her computer. Good work :)

PR #78, opened 2018-02-02
Scale out and debug memory usage

aug24:
## What does this change?

Allows for two instances (minimum) up to four during deployments.
Adds a flag to collate memory usage in the JRE

## What is the value of this?

To mitigate and investigate the instance failure due to memory exhaustion.

Yes

## Any additional notes?

These changes can be reverted once the memory issues are understood and resolved.

### DIFF
--- a/cloudformation/security-hq.template.yaml
+++ b/cloudformation/security-hq.template.yaml
@@ -272,7 +272,7 @@ Resources:
         Ref: LaunchConfig
       MinSize: 2
       MaxSize: 4
-      DesiredCapacity: 1
+      DesiredCapacity: 2
       HealthCheckType: ELB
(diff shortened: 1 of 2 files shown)


cloudformation/security-hq.template.yaml
```diff
@@ -265,8 +265,8 @@ Resources:
         Ref: Subnets
       LaunchConfigurationName:
         Ref: LaunchConfig
-      MinSize: 1
-      MaxSize: 2
+      MinSize: 2
+      MaxSize: 4
       DesiredCapacity: 1
```
adamnfish: Desired needs to be 2 as well

aug24: Damn.

PR #80, opened 2018-02-02
Feedback on snyk instructions

aug24:
## What does this change?

Improvements to Snyk documentation based on Kate's feedback.

## What is the value of this?

Help people to use Snyk.

No.

### DIFF
--- a/hq/markdown/snyk.md
+++ b/hq/markdown/snyk.md
@@ -7,7 +7,7 @@ vulnerabilities in your application.
 
 ## Installing
 
-Snyk requires node, and a Node package manager such as npm or Yarn.
+Snyk requires node, and `npm`.
 
 ### Developer Install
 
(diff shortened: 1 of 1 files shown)


hq/markdown/snyk.md
```diff
@@ -51,7 +49,7 @@ on the PATH.  As a result it is necessary to pre-pend the sbt location onto the
 
 ```
   "devDependencies": {
-    "snyk": "1.69.3"
+    "snyk": "1.68.0"
```
katebee: 🤔 Is there a reason this is changing?

aug24: Recommendations from Snyk because the 1.69 releases remove minor bugs.  At least one of those bugs was involved in the difficulties in setting up Snyk for SHQ.

We should use the versions we then suggest others use, for dogfood reasons.

katebee: > We should use the versions we then suggest others use

Agree 💯 

...However, SHQ specifies `"snyk": "1.69.3"` in the package.json so this should be the one we are using?

Also, earlier in this file ([line 40](https://github.com/guardian/security-hq/pull/80/files/27d812f38bda4a7ec8a4de4a9d97b0b648e40e20#diff-ca96dc326cfaa20524b64a8facbcf9b6R40)) version `"1.69.3"` is also specified

picture 195

Is there a reason we mention two different versions?

PR #83, opened 2018-02-06
Upversion play-googleauth

aug24:
## What does this change?

Upversions Play GoogleAuth library, and implements call to new method for setting up google auth.

## What is the value of this?

Should finally kill the anti-forgery token problem.

No.

## Any additional notes?

No.

### DIFF
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -2,6 +2,7 @@ import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagement
 import com.gu.configraun.Configraun
 import com.gu.configraun.aws.AWSSimpleSystemsManagementFactory
 import com.gu.configraun.models._
+import config.Config
 import controllers._
 import filters.HstsFilter
 import play.api.ApplicationLoader.Context
@@ -52,14 +53,15 @@ class AppComponents(context: Context)
     }
   }
 
-  val cacheService = new CacheService(configuration, applicationLifecycle, environment)
+  private val cacheService = new CacheService(configuration, applicationLifecycle, environment)
+  private val googleAuthConfig = Config.googleSettings(httpConfiguration, configuration)
 
   override def router: Router = new Routes(
     httpErrorHandler,
-    new HQController(configuration, cacheService),
-    new SecurityGroupsController(configuration, cacheService),
-    new SnykController(configuration, configraun),
-    new AuthController(environment, configuration),
+    new HQController(configuration, cacheService, googleAuthConfig),
+    new SecurityGroupsController(configuration, cacheService, googleAuthConfig),
+    new SnykController(configuration, configraun, googleAuthConfig),
+    new AuthController(environment, configuration, googleAuthConfig),
     new UtilityController(),
     assets
   )
--- a/hq/app/config/Config.scala
+++ b/hq/app/config/Config.scala
@@ -3,7 +3,7 @@ package config
 import java.io.FileInputStream
 
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
-import com.gu.googleauth.{GoogleAuthConfig, GoogleGroupChecker, GoogleServiceAccount}
+import com.gu.googleauth.{AntiForgeryChecker, GoogleAuthConfig, GoogleGroupChecker, GoogleServiceAccount}
 import io.jsonwebtoken.SignatureAlgorithm
 import model.{AwsAccount, DEV, PROD, Stage}
 import play.api.Configuration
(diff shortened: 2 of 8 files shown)


hq/app/config/Config.scala
```diff
@@ -20,16 +22,20 @@ object Config {
...
-  def googleSettings(implicit config: Configuration): GoogleAuthConfig = {
+  def googleSettings(httpConfiguration: HttpConfiguration, config: Configuration): GoogleAuthConfig = {
     val clientId = requiredString(config, "auth.google.clientId")
     val clientSecret = requiredString(config, "auth.google.clientSecret")
     val domain = requiredString(config, "auth.domain")
     val redirectUrl = s"${requiredString(config, "host")}/oauthCallback"
-    GoogleAuthConfig(
+    val secret: String = httpConfiguration.secret.secret
+    val signatureAlgorithm: SignatureAlgorithm = SignatureAlgorithm.forName(httpConfiguration.session.jwt.signatureAlgorithm)
+    GoogleAuthConfig.withSecret(
```
rtyley: Without creating and using a new `GoogleAuthConfig.withSecret()` method, you have two alternative approaches that both use the existing constructor `GoogleAuthConfig.apply()` method in `play-googleauth`.

The closest to your current code is just to construct an `AntiForgeryChecker`, like this :

```
    GoogleAuthConfig(
      clientId,
      clientSecret,
      redirectUrl,
      domain,
      antiForgeryChecker = AntiForgeryChecker(secret,signatureAlgorithm)
    )
```

...this avoids the need to make a new `GoogleAuthConfig.withSecret()` method, but is also strongly-typed - there are fewer `String` arguments floating around to risk getting accidentally swapped.

Better still is to drop the code manually extracting `secret` & `signatureAlgorithm`, and use the `AntiForgeryChecker.borrowSettingsFromPlay(httpConfiguration)` method:

```
    GoogleAuthConfig(
      clientId,
      clientSecret,
      redirectUrl,
      domain,
      antiForgeryChecker = 
        AntiForgeryChecker.borrowSettingsFromPlay(httpConfiguration)
    )
```

hq/app/AppComponents.scala
```diff
@@ -52,14 +53,15 @@ class AppComponents(context: Context)
...
   override def router: Router = new Routes(
     httpErrorHandler,
-    new HQController(configuration, cacheService),
-    new SecurityGroupsController(configuration, cacheService),
-    new SnykController(configuration, configraun),
-    new AuthController(environment, configuration),
+    new HQController(configuration, cacheService, googleAuthConfig),
+    new SecurityGroupsController(configuration, cacheService, googleAuthConfig),
+    new SnykController(configuration, configraun, googleAuthConfig),
+    new AuthController(environment, configuration, googleAuthConfig),
```
rtyley: Minor note here but I'm not sure why it's become necessary to add `googleAuthConfig` to all the controllers? One note is that it might be better to pass a more tightly-focused object like a configured `AuthAction`, if that's what the `Controller` needs.

You could also go the full hog and create a `ControllerComponents` object if you want, like we do in Ophan: https://github.com/guardian/ophan/blame/fd7474fe9c632/dashboard/app/configuration/ApplicationComponents.scala#L71

aug24: Beyond the scope of this change, but I will discuss with @adamnfish

adamnfish: I agree that if all the required info is available in the config then passing the config would still be enough, but this does introduce a new piece of info that comes from the Play environment (the choice of secret and algorithm). I think doing this at the edge of the program and being explicit in the wiring makes sense. Certainly, creating `ControllerComponents` means you're still passing it into all the controllers and I'm not sure the indirection (and reduced specificity) that comes with using an intermediate data structure to hold the wiring gains much for us in this small program.

hq/app/controllers/SnykController.scala
```diff
@@ -10,11 +10,13 @@ import scala.concurrent.ExecutionContext
 import logic.SnykDisplay
 import api.Snyk
 import com.gu.configraun.Errors.ConfigraunError
+import com.gu.googleauth.GoogleAuthConfig
 import utils.attempt.{Attempt, Failure}
 import utils.attempt.PlayIntegration.attempt
 import model._
 
-class SnykController(val config: Configuration, val configraun: com.gu.configraun.models.Configuration)
+class SnykController(val config: Configuration, val configraun: com.gu.configraun.models.Configuration,
```
adamnfish: This name clash feels like something we should look into in Configraun at some point (not now).

aug24: Yes.  There are several different 'configuration' objects in there now.  Google's, Configraun's and SHQ's.

PR #84, opened 2018-02-06
Using ami-68630a11 (with java8, awscli tools and ssm-agent already installed)

shtukas:
## What does this change?

Here we set security-hq to use AMI ami-68630a11 which was baked in AMIgo. This AMI comes with java8, awscli tools and the ssm-agent already installed. Accordingly the `UserData` section of the cloudformation has been updated.


shtukas:
Closing this as a new (better) one is coming.

PR #85, opened 2018-02-06
Moving to AMIgo backed AMIs

shtukas:
## What does this change?

This change updates the `hq/conf/riff-raff.yaml` file to make it possible for Riff-Raff to use the latest available AMI of the `security-hq` AMIgo recipe.

This recipe puts `aws-tools`, `java8` and `ssm-agent` onto `ubuntu-trusty`.

Accordingly, `cloudformation/security-hq.template.yaml` has been updated  by removing the duplicate work. 

Note: Ideally we would not have `apt-get -y update` and `apt-get -y upgrade` running as part of the launch config, but this is not the concern of this PR. This problem, namely the fact that a backed AMI needs this, will be addressed separately. 

Note: Since the cloud formation file is picked up as part of the deployment, there is here no requirement to do it manually beforehand in AWS console.

### DIFF
--- a/cloudformation/security-hq.template.yaml
+++ b/cloudformation/security-hq.template.yaml
@@ -233,18 +233,10 @@ Resources:
         'Fn::Base64': !Sub |
           #!/bin/bash -ev
           locale-gen en_GB.UTF-8
-          add-apt-repository -y ppa:openjdk-r/ppa
           apt-get -y update
           apt-get -y upgrade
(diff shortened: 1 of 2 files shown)


shtukas:
@aug24 @adamnfish @sihil

shtukas:
@adamnfish Understood!
... and tested by deploying the branch to PROD, so merging now.

PR #87, opened 2018-02-07
Correct path to sbt on team city

aug24:
## What does this change?

Correct path for sbt executable (local wrapper script for Security HQ)
Adds command alias required by Snyk for sbt 1.0+
Turns off colour on sbt.
Upversions vulnerable library.

## What is the value of this?

As is, Snyk fails to (1) find sbt; (2) invoke dependencyTree, (3) parse coloured output: this change corrects each of those so that Snyk tests can run.  

SHQ then did not pass the test; the upversioned library corrects this.

No

## Any additional notes?

No

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,6 +34,7 @@ lazy val hq = (project in file("hq")).
       "com.vladsch.flexmark" % "flexmark-all" % "0.28.20",
       "io.reactivex" %% "rxscala" % "0.26.5",
       "com.gu" %% "box" % "0.1.0",
+      "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11",
       "org.scalatestplus.play" %% "scalatestplus-play" % "3.1.2" % Test,
       "org.scalacheck" %% "scalacheck" % "1.13.4" % Test,
       "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % "1.1.6" % Test
@@ -81,7 +82,7 @@ lazy val lambdaCommon = (project in file("lambda/common")).
       "com.typesafe.scala-logging" %% "scala-logging" % "3.5.0",
       "ch.qos.logback" %  "logback-classic" % "1.2.3",
       "com.amazonaws" % "aws-java-sdk-config" % "1.11.246",
-      "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.10"
+      "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.11"
     )
   )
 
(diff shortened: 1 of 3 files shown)


sbt
```diff
@@ -26,7 +26,11 @@ do
...
 
 java $DEBUG_PARAMS \
-    -Xms1024M -Xmx2048M -Xss1M -XX:+CMSClassUnloadingEnabled -XX:MaxPermSize=1024M \
+    -Xms1024M -Xmx2048M \
+    -Xss1M \
+    -XX:+CMSClassUnloadingEnabled \
+    -XX:MaxPermSize=1024M \
     $CONF_PARAMS \
     $TC_PARAMS \
+    -Dsbt.log.noformat=true \
```
adamnfish: I think we only want this flag on TeamCity? Coloured output on our own shells is extremely useful!

build.sbt
```diff
@@ -91,3 +92,5 @@ lazy val root = (project in file(".")).
   settings(
     name := """security-hq"""
   )
+
+addCommandAlias("dependency-tree", "dependencyTree")
```
adamnfish: Is this required because of Snyk assuming an older version of sbt?

sbt
aug24: I don't know any way I can pass this in to the sbt invocation generated by snyk inside npm.
I could check for a TC-specific env var?  That's getting a bit ugly though.

build.sbt
aug24: The command alias is required because sbt 1.0+ changes the way command line arguments are accepted.  Previously, a cli dash-delimited string was automatically converted into a camel-case string.  Apparently, sbt 1.0 drops that.

sbt
```diff
@@ -18,15 +19,27 @@ do
       shift
     fi
     if [ "$arg" == "--team-city" ]; then
-      f=$(grep '^teamcity.configuration.properties.file' "$TEAMCITY_BUILD_PROPERTIES_FILE" | cut -d'=' -f 2)
-      TC_PARAMS="-Dteamcity.configuration.properties.file=$f"
-      echo "Adding team city property: $TC_PARAMS"
+      echo "Ignoring '--team-city' in favour of checking for TEAMCITY_BUILD_PROPERTIES_FILE env var; please remove this switch"
```
adamnfish: Given we control both ends of this, do we need this deprecation step? Why not just drop it?

aug24: Just being careful/considerate, really.  We need builds to continue until this is merged.  Then I'd remove the switch from TC, then remove from here.

PR #89, opened 2018-02-08
Add AWS Inspector to launch config

aug24:
## What does this change?

Installs the AWS Inspector package on instances as part of launch config

## What is the value of this?

The AWS Inspector service is available to run reports against.

Yes.

Changes are included but not run.

## Any additional notes?

No.

### DIFF
--- a/cloudformation/security-hq.template.yaml
+++ b/cloudformation/security-hq.template.yaml
@@ -256,6 +256,94 @@ Resources:
           # Start application
           start security-hq
 
+  InspectorResourceGroup:
+    Type: "AWS::Inspector::ResourceGroup"
+    Properties:
+      ResourceGroupTags:
+        - Key: Stage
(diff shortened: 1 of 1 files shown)


cloudformation/security-hq.template.yaml
```diff
@@ -253,6 +256,94 @@ Resources:
...
+          Value:
+            Ref: Stage
+        - Key: Stack
+          Value:
+            Fn::FindInMap: [ Constants, Stack, Value ]
+        - Key: App
+          Value:
+            Fn::FindInMap: [ Constants, App, Value ]
+
+  InspectorAssessmentTarget:
```
adamnfish: Does this type not accept tags?

adamnfish: (and elsewhere)

PR #90, opened 2018-02-08
Display more helpful warnings, and confirm when there are no open security groups

katebee:
#### 👉 Show confirmation card when there are no open security groups

This is nicer and more unambiguous than the previously empty container. Teams should be pleased with their big green check! 👍 👍 

picture 212

#### 👉 More details on the warnings for account pages 

This significantly changes the page for an accounts security groups. Previously if the data was not available in the cache, the user would hit the standard error page. However, passing the error to the view allows some bonus styling, and keeps the user on their account hub, so they can still use the tabs on the sub nav.

picture 214

#### 👉 Add warnings about exposed keys to the all accounts page

We currently only show exposed keys on the individual account pages, which makes the warnings less likely to be encountered if someone is using the all accounts page. This PR adds a check for any accounts with exposed keys, and will add a red alert to the top of the page for each account. The alert links to the IAM account page which will have the complete details on any and all exposed keys.

picture 208

Nope

## Any additional notes?

For highlighting exposed keys on the all accounts page, arguably we should map an account to `true` if the exposedKeys cache data is missing. However, the basic alerting introduced is still better than no alerting; happy to iterate!

There are some bonus stylistic improvements and opinionated template restructuring.

### DIFF
--- a/hq/app/logic/ReportDisplay.scala
+++ b/hq/app/logic/ReportDisplay.scala
@@ -93,14 +93,12 @@ object ReportDisplay {
   }
 
   def exposedKeysSummary(allReports: Map[AwsAccount, Either[FailedAttempt, List[ExposedIAMKeyDetail]]]): Map[AwsAccount, Boolean] = {
-    def checkForExposedKeys(temp: Either[FailedAttempt, List[ExposedIAMKeyDetail]]): Boolean = temp match {
-      case Right(keys) if keys.nonEmpty => true
-      case _ => false
-    }
-
     for {
-      report <- allReports
-    } yield (report._1, checkForExposedKeys(report._2))
+      (account, report) <- allReports
+    } yield (account, report match {
+        case Right(keys) if keys.nonEmpty => true
+        case _ => false
+    })
   }
 
   def sortAccountsByReportSummary[L](reports: List[(AwsAccount, Either[L, CredentialReportDisplay])]): List[(AwsAccount, Either[L, CredentialReportDisplay])] = {
--- a/hq/app/logic/ReportDisplay.scala
+++ b/hq/app/logic/ReportDisplay.scala
@@ -3,6 +3,7 @@ package logic
 import model._
 import DateUtils.dayDiff
 import org.joda.time.{DateTime, DateTimeZone, Days}
+import utils.attempt.FailedAttempt
 
 
 object ReportDisplay {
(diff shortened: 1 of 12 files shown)


hq/app/logic/ReportDisplay.scala
```diff
@@ -91,6 +92,17 @@ object ReportDisplay {
...
   }
 
+  def exposedKeysSummary(allReports: Map[AwsAccount, Either[FailedAttempt, List[ExposedIAMKeyDetail]]]): Map[AwsAccount, Boolean] = {
+    def checkForExposedKeys(temp: Either[FailedAttempt, List[ExposedIAMKeyDetail]]): Boolean = temp match {
+      case Right(keys) if keys.nonEmpty => true
+      case _ => false
+    }
+
+    for {
+      report <- allReports
```
adamnfish: You should be able to extract the tuple parts with good names here:

```scala
for {
  (account, report) <- allReports
} yield (account, checkForExposedKeys(report))
```

hq/app/logic/ReportDisplay.scala
```diff
@@ -91,6 +92,17 @@ object ReportDisplay {
     ReportSummary(warnings, errors, other)
   }
 
+  def exposedKeysSummary(allReports: Map[AwsAccount, Either[FailedAttempt, List[ExposedIAMKeyDetail]]]): Map[AwsAccount, Boolean] = {
+    def checkForExposedKeys(temp: Either[FailedAttempt, List[ExposedIAMKeyDetail]]): Boolean = temp match {
```
adamnfish: I don't think you meant to leave this named as temp?

Also, does this need to be a def? It's only called once so probably better inline.

hq/app/views/iam/iam.scala.html
```diff
@@ -35,6 +36,28 @@ <h4 class="header light grey-text text-lighten-5">All accounts</h4>
 
 } { @* Main content *@
     <div class="container">
+
+        <div class="row iam-key-alert__wrapper">
```
adamnfish: This is a good idea

hq/app/views/sgs/sgsAccount.scala.html
```diff
@@ -20,32 +22,48 @@ <h4 class="header light grey-text text-lighten-5">For @awsAccount.name</h4>
...
-        <div class="row">
-            @views.html.fragments.openSecurityGroups(flaggedSgs, shadow=true)
-        </div>
+                <div class="row">
+                    @if(flaggedSgs.nonEmpty) {
+                        @views.html.fragments.openSecurityGroups(flaggedSgs, shadow=true)
+                    } else {
+                    <div class="card-panel valign-wrapper">
+                        <i class="material-icons medium green-text left">verified_user</i>
+                        <h5>No open Security Groups found</h5>
```
adamnfish: Is `<h5>` part of the hierarchy here, or is it just chosen for its appearance?

katebee: Chosen for appearance, to circumvent adding more css :)

If there is a preferred convention, happy to follow

hq/app/logic/ReportDisplay.scala
katebee: Much nicer code, thank you 🎉

PR #92, opened 2018-02-14
Revert add indication of related CloudFormation stack to IAM users

katebee:
Reverts guardian/security-hq#91

Since I probably will not get the chance to update the StackSets until tomorrow... 😖


shtukas:
+1

PR #107, opened 2018-03-07
Reconsider snyk scripting

aug24:
## What does this change?

Updates the scripting to invoke snyk

## What is the value of this?

Ensuring testing only runs on master, clarifying how it works, documenting.

No


shtukas:
Nice, but get a proper +1 from @katebee :)

PR #109, opened 2018-03-13
New snyk api

aug24:
## What does this change?

Calls the Snyk API and correctly parses the new form of response (groups).

## What is the value of this?

We have re-organised our Snyk accounts.  This change handles the resulting change in API responses.
It also handles error cases much better.

No.

## Any additional notes?

The token in ssm needs to be changed, as does the the organisation, which now takes the parent group id.

### DIFF
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -26,12 +26,6 @@ class AppComponents(context: Context)
   with AhcWSComponents with AssetsComponents {
 
   implicit val impWsClient: WSClient = wsClient
-//  val snykWsClient: WSClient = {
-//    val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(3))
-//    val asyncHttpClient = new AsyncHttpClientProvider(environment, configuration, applicationLifecycle)(ec).get
-//    new AhcWSClientProvider(asyncHttpClient).get
-//  }
-
   implicit val impPlayBodyParser: BodyParser[AnyContent] = playBodyParsers.default
   implicit val impControllerComponents: ControllerComponents = controllerComponents
   implicit val impAssetsFinder: AssetsFinder = assetsFinder
@@ -71,7 +65,6 @@ class AppComponents(context: Context)
     new HQController(configuration, cacheService, googleAuthConfig),
     new SecurityGroupsController(configuration, cacheService, googleAuthConfig),
     new SnykController(configuration, configraun, googleAuthConfig),
-//    new SnykController(configuration, configraun, googleAuthConfig, snykWsClient),
     new AuthController(environment, configuration, googleAuthConfig),
     new UtilityController(),
     assets
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -1,5 +1,3 @@
-import java.util.concurrent.Executors
-
 import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagement
 import com.gu.configraun.Configraun
 import com.gu.configraun.aws.AWSSimpleSystemsManagementFactory
@@ -10,15 +8,12 @@ import filters.HstsFilter
 import play.api.ApplicationLoader.Context
 import play.api.{BuiltInComponentsFromContext, Logger}
 import play.api.libs.ws.WSClient
-import play.api.libs.ws.ahc.{AhcWSClientProvider, AhcWSComponents, AsyncHttpClientProvider}
+import play.api.libs.ws.ahc.AhcWSComponents
 import play.api.mvc.{AnyContent, BodyParser, ControllerComponents}
 import play.api.routing.Router
 import play.filters.csrf.CSRFComponents
 import router.Routes
 import services.CacheService
-
-import scala.concurrent.ExecutionContext
-
 
 class AppComponents(context: Context)
   extends BuiltInComponentsFromContext(context)
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -26,11 +26,11 @@ class AppComponents(context: Context)
   with AhcWSComponents with AssetsComponents {
 
   implicit val impWsClient: WSClient = wsClient
-  val snykWsClient: WSClient = {
-    val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(3))
-    val asyncHttpClient = new AsyncHttpClientProvider(environment, configuration, applicationLifecycle)(ec).get
-    new AhcWSClientProvider(asyncHttpClient).get
-  }
+//  val snykWsClient: WSClient = {
+//    val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(3))
+//    val asyncHttpClient = new AsyncHttpClientProvider(environment, configuration, applicationLifecycle)(ec).get
+//    new AhcWSClientProvider(asyncHttpClient).get
+//  }
 
   implicit val impPlayBodyParser: BodyParser[AnyContent] = playBodyParsers.default
   implicit val impControllerComponents: ControllerComponents = controllerComponents
@@ -70,7 +70,8 @@ class AppComponents(context: Context)
     httpErrorHandler,
     new HQController(configuration, cacheService, googleAuthConfig),
     new SecurityGroupsController(configuration, cacheService, googleAuthConfig),
-    new SnykController(configuration, configraun, googleAuthConfig, snykWsClient),
+    new SnykController(configuration, configraun, googleAuthConfig),
+//    new SnykController(configuration, configraun, googleAuthConfig, snykWsClient),
     new AuthController(environment, configuration, googleAuthConfig),
     new UtilityController(),
     assets
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -1,3 +1,5 @@
+import java.util.concurrent.Executors
+
 import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagement
 import com.gu.configraun.Configraun
 import com.gu.configraun.aws.AWSSimpleSystemsManagementFactory
@@ -8,12 +10,14 @@ import filters.HstsFilter
 import play.api.ApplicationLoader.Context
 import play.api.{BuiltInComponentsFromContext, Logger}
 import play.api.libs.ws.WSClient
-import play.api.libs.ws.ahc.AhcWSComponents
+import play.api.libs.ws.ahc.{AhcWSClientProvider, AhcWSComponents, AsyncHttpClientProvider}
 import play.api.mvc.{AnyContent, BodyParser, ControllerComponents}
 import play.api.routing.Router
 import play.filters.csrf.CSRFComponents
 import router.Routes
 import services.CacheService
+
+import scala.concurrent.ExecutionContext
 
 
 class AppComponents(context: Context)
@@ -22,6 +26,12 @@ class AppComponents(context: Context)
   with AhcWSComponents with AssetsComponents {
 
   implicit val impWsClient: WSClient = wsClient
+  val snykWsClient: WSClient = {
+    val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(3))
+    val asyncHttpClient = new AsyncHttpClientProvider(environment, configuration, applicationLifecycle)(ec).get
+    new AhcWSClientProvider(asyncHttpClient).get
+  }
+
   implicit val impPlayBodyParser: BodyParser[AnyContent] = playBodyParsers.default
   implicit val impControllerComponents: ControllerComponents = controllerComponents
   implicit val impAssetsFinder: AssetsFinder = assetsFinder
@@ -60,7 +70,7 @@ class AppComponents(context: Context)
     httpErrorHandler,
     new HQController(configuration, cacheService, googleAuthConfig),
     new SecurityGroupsController(configuration, cacheService, googleAuthConfig),
-    new SnykController(configuration, configraun, googleAuthConfig),
+    new SnykController(configuration, configraun, googleAuthConfig, snykWsClient),
     new AuthController(environment, configuration, googleAuthConfig),
     new UtilityController(),
     assets
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -21,13 +21,13 @@ object Snyk {
     }
   }
 
-  def getProjects(token: SnykToken, organisations: List[SnykOrganisation], wsClient: WSClient)(implicit ec: ExecutionContext): Attempt[List[(SnykOrganisation, WSResponse)]] = {
+  def getProjects(token: SnykToken, organisations: List[SnykOrganisation], wsClient: WSClient)(implicit ec: ExecutionContext): Attempt[List[(SnykOrganisation, String)]] = {
     Attempt.traverse(organisations) { organisation =>
       val snykProjectsUrl = s"https://snyk.io/api/v1/org/${organisation.id}/projects"
       val futureResponse = wsClient.url(snykProjectsUrl)
         .addHttpHeaders("Authorization" -> s"token ${token.value}")
         .get()
-        .transform(response => (organisation, response), f => f )
+        .transform(response => (organisation, response.body), f => f )
       Attempt.fromFuture(futureResponse) {
         case NonFatal(e) =>
           val failure = Failure(e.getMessage, "Could not read projects from Snyk", 502, None, Some(e))
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -25,7 +25,6 @@ object Snyk {
 
   def getProjects(token: SnykToken, organisations: List[SnykOrganisation], wsClient: WSClient)(implicit ec: ExecutionContext): Attempt[List[(SnykOrganisation, WSResponse)]] = {
     Attempt.traverse(organisations) { organisation =>
-      println(organisation)
       val snykProjectsUrl = s"https://snyk.io/api/v1/org/${organisation.id}/projects"
       val b = wsClient.url(snykProjectsUrl)
         .addHttpHeaders("Authorization" -> s"token ${token.value}")
@@ -44,7 +43,6 @@ object Snyk {
     val projectVulnerabilityResponses = projects
       .map(project => {
         val snykProjectUrl = s"https://snyk.io/api/v1/org/${project.organisation.get.id}/project/${project.id}/issues"
-        println(snykProjectUrl)
         val projectIssuesFilter = Json.obj(
           "filters" -> Json.obj(
             "severity" -> JsArray(List(
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -20,10 +20,10 @@ object Snyk {
   }
 
   def getProjects(token: SnykToken, organisations: List[SnykOrganisation], wsClient: WSClient)(implicit ec: ExecutionContext): Attempt[List[(SnykOrganisation, String)]] = {
-    Attempt.traverse(organisations) { organisation =>
+    Attempt.labelledTraverse(organisations) { organisation =>
       val snykProjectsUrl = s"https://snyk.io/api/v1/org/${organisation.id}/projects"
       val futureResponse = snykRequest(token, snykProjectsUrl, wsClient).get
-        .transform(response => (organisation, response.body), f => f )
+        .map(response => response.body)
       handleFuture(futureResponse, "project")
     }
   }
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -10,7 +10,8 @@ import scala.util.control.NonFatal
 
 object Snyk {
 
-  def snykRequest(token:SnykToken, url:String, wsClient: WSClient) = wsClient.url(url).addHttpHeaders("Authorization" -> s"token ${token.value}")
+  def snykRequest(token:SnykToken, url:String, wsClient: WSClient) = 
+    wsClient.url(url).addHttpHeaders("Authorization" -> s"token ${token.value}")
 
   def getSnykOrganisations(token: SnykToken, wsClient: WSClient)(implicit ec: ExecutionContext): Attempt[WSResponse] = {
     val snykOrgUrl = "https://snyk.io/api/v1/orgs"
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -5,7 +5,7 @@ import play.api.libs.json._
 import play.api.libs.ws.{WSClient, WSRequest, WSResponse}
 import utils.attempt.{Attempt, FailedAttempt, Failure}
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 import scala.util.control.NonFatal
 
 object Snyk {
@@ -52,7 +52,7 @@ object Snyk {
     }
   }
 
-  def handleFuture[A](future: scala.concurrent.Future[A], label: String)(implicit ec: ExecutionContext) = Attempt.fromFuture(future) {
+  def handleFuture[A](future: Future[A], label: String)(implicit ec: ExecutionContext) = Attempt.fromFuture(future) {
     case NonFatal(e) =>
       val failure = Failure(e.getMessage, s"Could not read ${label} from Snyk", 502, None, Some(e))
       FailedAttempt(failure)
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -10,11 +10,11 @@ import scala.util.control.NonFatal
 
 object Snyk {
 
+  def snykRequest(token:SnykToken, url:String, wsClient: WSClient) = wsClient.url(url).addHttpHeaders("Authorization" -> s"token ${token.value}")
+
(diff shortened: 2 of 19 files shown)


shtukas:
@aug24 conflicts with master branch

hq/app/AppComponents.scala
```diff
@@ -8,20 +10,28 @@ import filters.HstsFilter
...
   with CSRFComponents
   with AhcWSComponents with AssetsComponents {
 
   implicit val impWsClient: WSClient = wsClient
+//  val snykWsClient: WSClient = {
+//    val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(3))
+//    val asyncHttpClient = new AsyncHttpClientProvider(environment, configuration, applicationLifecycle)(ec).get
+//    new AhcWSClientProvider(asyncHttpClient).get
+//  }
+
```
katebee: Can we delete this comment? (and the commented out code on line 74)

hq/app/controllers/SnykController.scala
```diff
@@ -31,18 +32,22 @@ class SnykController(val config: Configuration, val configraun: com.gu.configrau
     Action.async {
       attempt {
         for {
+
```
adamnfish: Do we need this whitespace?

hq/app/controllers/SnykController.scala
```diff
@@ -60,14 +65,11 @@ class SnykController(val config: Configuration, val configraun: com.gu.configrau
     case Right(a:String) => Attempt.Right(SnykToken(a))
   }
 
-  def getSnykOrganisation: Attempt[SnykOrganisationName] = configraun.getAsString("/snyk/organisation") match {
+  def getSnykOrganisation: Attempt[SnykGroupId] = configraun.getAsString("/snyk/organisation") match {
     case Left(a:ConfigraunError) =>
       val failure = Failure(a.message, "Could not read Snyk organisation from aws parameter store", 500, None, Some(a.e))
       Attempt.Left(failure)
-    case Right(a:String) => Attempt.Right(SnykOrganisationName(a))
+    case Right(a:String) => Attempt.Right(SnykGroupId(a))
```
adamnfish: I doubt you need to annotate `a`, but if you do space after the `:`. Also, `a` is normally used for things of type `A`. This is a string so it probably has a better name? `id`?

hq/app/logic/SnykDisplay.scala
```diff
@@ -2,63 +2,60 @@ package logic
...
-      org <- orglist.find(_.name == requiredOrganisation.value)
-    } yield org
-    val error = parseJsonToError(s)
-    val f = Failure(s"Unable to find organisation from $s", s"Could not read Snyk response (${error.error})", 502, None, None)
-    Attempt.fromOption(organisation, FailedAttempt(f))
+  def parseOrganisations(s: String, snykGroupId: SnykGroupId)(implicit ec: ExecutionContext): Attempt[List[SnykOrganisation]] = {
+    for {
+      organisationList <- parseJsonToOrganisationList(s)
+      guardianOrganisationList = organisationList filter {
+        case SnykOrganisation(_, id, Some(group)) => group.id==snykGroupId.value
```
adamnfish: spaces around the `==`

hq/app/logic/SnykDisplay.scala
```diff
@@ -2,63 +2,60 @@ package logic
...
 
-  def getOrganisation(s: String, requiredOrganisation: SnykOrganisationName): Attempt[SnykOrganisation] = {
-    val organisation = for {
-      orglist <- parseJsonToOrganisationList(s)
-      org <- orglist.find(_.name == requiredOrganisation.value)
-    } yield org
-    val error = parseJsonToError(s)
-    val f = Failure(s"Unable to find organisation from $s", s"Could not read Snyk response (${error.error})", 502, None, None)
-    Attempt.fromOption(organisation, FailedAttempt(f))
+  def parseOrganisations(s: String, snykGroupId: SnykGroupId)(implicit ec: ExecutionContext): Attempt[List[SnykOrganisation]] = {
```
adamnfish: Is there a better name for `s`? Not a big deal, just hard to see what it is here from the name `s`.

adamnfish: After study, perhaps something that mentions `json`?

hq/app/logic/SnykDisplay.scala
```diff
@@ -2,63 +2,60 @@ package logic
...
   private def parseJsonToError(s: String): SnykError = try {
     Json.parse(s).asOpt[SnykError].getOrElse(SnykError(s))
   } catch {
-    case _: JsonParseException => SnykError(s)
+    case e: JsonParseException => SnykError(e.getMessage)
   }
 
-  private def parseJsonToOrganisationList(s: String): Option[List[SnykOrganisation]] = try {
-    (Json.parse(s) \ "orgs").asOpt[List[SnykOrganisation]]
+  def parseJsonToObject[A](label: String, s: String, f: (String => A)): Attempt[A] = try {
```
adamnfish: What is happening here?! You're using exceptions for flow handling. If `f` can fail it should return an `Attempt`.

hq/app/logic/SnykDisplay.scala
```diff
@@ -2,63 +2,60 @@ package logic
...
+    Attempt.Right(f(s))
   } catch {
-    case _: JsonParseException => None
+    case e: JsonParseException =>
+      val f = Failure(s"Unable to find $label from $s", s"Could not read Snyk response ($s)", 502, None, Some(e))
+      Attempt.Left(f)
+    case e: Exception =>
```
adamnfish: `NonFatal` is better than `Exception`, but if you're checking for expected exceptions those are known ahead of time.

The Play JSON library has a `validate` method that returns an either. You'll probably find that easier than using exception handling for the conversion from `JsValue -> A`.
https://www.playframework.com/documentation/2.6.x/ScalaJson#Using-validation

hq/app/logic/SnykDisplay.scala
```diff
@@ -2,63 +2,60 @@ package logic
...
   }
 
-  def getProjectIdList(s: String): Attempt[List[SnykProject]] = {
-    val projects = parseJsonToProjectIdList(s)
-    val error = parseJsonToError(s)
-    val f = Failure(s"Unable to find project ids from $s", s"Could not read Snyk response (${error.error})", 502, None, None)
-    Attempt.fromOption(projects, FailedAttempt(f))
-  }
+  def parseJsonToOrganisationList(s: String): Attempt[List[SnykOrganisation]] =
+    parseJsonToObject("organisations", s, body => {(Json.parse(body) \ "orgs").as[List[SnykOrganisation]]} )
```
adamnfish: `validate` is better than `as` (see above).

hq/app/logic/SnykDisplay.scala
```diff
@@ -2,63 +2,60 @@ package logic
...
 
-  private def parseJsonToProjectIdList(s: String) = try {
-    (Json.parse(s) \ "projects").asOpt[List[SnykProject]]
-  } catch {
-    case _: JsonParseException => None
+  def getProjectIdList(ss: List[(SnykOrganisation, String)])(implicit ec: ExecutionContext): Attempt[List[((SnykOrganisation, String), List[SnykProject])]] = {
```
adamnfish: Is `ss` a good name for something of type `(SnykOrganisation, String)`?

hq/app/logic/SnykDisplay.scala
```diff
@@ -2,63 +2,60 @@ package logic
...
-  }
+  def parseJsonToProjectIdList(s: String): Attempt[List[SnykProject]] =
+    parseJsonToObject("project ids", s, body => (Json.parse(body) \ "projects").as[List[SnykProject]] )
 
-  private def parseJsonToProjectVulnerabilities(s: String) = try {
-    Json.parse(s).asOpt[SnykProjectIssues]
-  } catch {
-    case _: JsonParseException => None
+  def parseProjectVulnerabilitiesBody(s: String): Attempt[SnykProjectIssues] =
+  parseJsonToObject("project vulnerabilities", s, body => Json.parse(body).as[SnykProjectIssues])
```
adamnfish: Indentation is off here (and also, this parse `f` thing with `as` is probably not what we're after see above)

hq/app/logic/SnykDisplay.scala
adamnfish: Something like this jumps to mind if you want to do it generally for re-use (not compiled)

```scala
def extractJson[A](json: Json)(implicit decoder: Reads[A]): Attempt[A] = {
  Attempt.fromEither(json.validate[A].left.map { jsError =>
    FailedAttempt {
      jsError.errors.map { case (path, errs) =>
        Failure(  ..  do something helpful with the path and errors (use context for path)  ..  )
      }
    }
  })
}
```

There is some relevant stuff in Amiable:
https://github.com/guardian/amiable/blob/master/app/prism/JsonUtils.scala

hq/app/controllers/SnykController.scala
```diff
@@ -31,18 +32,22 @@ class SnykController(val config: Configuration, val configraun: com.gu.configrau
     Action.async {
       attempt {
         for {
+
           token <- getSnykToken
           requiredOrganisation <- getSnykOrganisation
 
           organisationResponse <- Snyk.getSnykOrganisations(token, wsClient)
-          organisation <- SnykDisplay.getOrganisation(organisationResponse.body, requiredOrganisation)
+          organisations <- SnykDisplay.parseOrganisations(organisationResponse.body, requiredOrganisation)
```
adamnfish: WSClient responses have a `.json` method that returns the body as a JsValue, which might save you some hassle?

aug24: Decided to stick with parsing the json body and coping with non-json data, just in case.  It would change very little code now.

package.json
```diff
@@ -25,9 +25,9 @@
     "prettier": "prettier -o --write './**/*.{js,css}'",
     "sass-lint": "sass-lint -v -q hq/public/stylesheets/*.css",
     "not-master": "test \"$BRANCH_NAME\" != \"master\" && echo \"BRANCH_NAME variable is '$BRANCH_NAME', not master\" ",
-    "snyk-test": "PATH=$(pwd):$PATH snyk test --debug --org=guardian --json --file=build.sbt ",
+    "snyk-test": "PATH=$(pwd):$PATH snyk test --debug --org=guardian-cuu --json --file=build.sbt ",
     "snyk-test-master": "npm run not-master --silent || npm run snyk-test --silent ",
-    "snyk-monitor": "PATH=$(pwd):$PATH snyk monitor --debug --org=guardian --file=build.sbt ",
+    "snyk-monitor": "PATH=$(pwd):$PATH snyk monitor --debug --org=guardian-cuu --file=build.sbt ",
```
katebee: Why `--org=guardian-cuu`? That is not the org that Security HQ sits in...

We probably want `--org=guardian-security` 😄

shtukas: +1, thanks Kate. When I saw it the first time I thought that "cuu" had a secret meaning I didn't know about :)

adamnfish: Is it ok to hard code that in the repo?

hq/app/api/Snyk.scala
```diff
@@ -2,40 +2,36 @@ package api
...
-        .addHttpHeaders("Authorization" -> s"token ${token.value}")
-        .get()
-    Attempt.fromFuture(a) { case NonFatal(e) =>
-      val failure = Failure(e.getMessage, "Could not read projects from Snyk", 502, None, Some(e))
-      FailedAttempt(failure)
+  def getProjects(token: SnykToken, organisations: List[SnykOrganisation], wsClient: WSClient)(implicit ec: ExecutionContext): Attempt[List[(SnykOrganisation, String)]] = {
+    Attempt.traverse(organisations) { organisation =>
+      val snykProjectsUrl = s"https://snyk.io/api/v1/org/${organisation.id}/projects"
+      val futureResponse = snykRequest(token, snykProjectsUrl, wsClient).get
+        .transform(response => (organisation, response.body), f => f )
```
adamnfish: `transform` is a method in Future for mapping the value in the Future and providing a handler for errors in one go. Since you are using identity for the error handling this would be better written as `map`. At that point though, I would suggest it's better to map the resulting Attempt rather than the intermediate Future. It would reduce the dissonance of mapping over different effects in different places.

This is also a step towards splitting the logic of processing the WSResponse from the code that performs the request. A common pattern for the AWS calls is the following (simplified for our purposes here).

```scala
def doSomethingWithAResponse(response: ServiceResponse): MyType = ???

def awsCall(arg: Something, client: ServiceClient): MyType = {
  makeAnAttempt(client.call(arg)).map(doSomethingWithAResponse)
}
```
This means we have a function that fetches the response separate to a function that does things with that response. I don't mean to say you have to do exactly this here (they are different!) but it'd be great to see minimal logic/code within a traverse call, in particular.

hq/app/api/Snyk.scala
```diff
@@ -48,18 +44,16 @@ object Snyk {
...
+        snykRequest(token, snykProjectUrl, wsClient).post(projectIssuesFilter)
       })
     Attempt.traverse(projectVulnerabilityResponses) {
-      projectVulnerabilityResponse =>
-        Attempt.fromFuture(projectVulnerabilityResponse) {
-          case NonFatal(e) =>
-            val failure = Failure(e.getMessage, "Could not read project vulnerabilities from Snyk", 502, None, Some(e))
-            FailedAttempt(failure)
-        }
+      projectVulnerabilityResponse => handleFuture(projectVulnerabilityResponse, "project vulnerabilities")
```
adamnfish: This is a good amount of complexity for a traverse call (very little)

hq/app/api/Snyk.scala
```diff
@@ -48,18 +44,16 @@ object Snyk {
...
     }
   }
+
+  def handleFuture[A](future: scala.concurrent.Future[A], label: String)(implicit ec: ExecutionContext) = Attempt.fromFuture(future) {
```
adamnfish: The file should have `import scala.concurrent.Future` in it, no need to fully qualify Future on an argument's type annotation.

hq/app/logic/SnykDisplay.scala
```diff
@@ -1,81 +1,63 @@
...
-  }
+  def getProjectIdList(organisationAndRequestList: List[(SnykOrganisation, String)])(implicit ec: ExecutionContext): Attempt[List[((SnykOrganisation, String), List[SnykProject])]] =
+    Attempt.labelledTraverse(organisationAndRequestList) { s => parseJsonToProjectIdList(s._2) }
 
-  private def parseJsonToProjectIdList(s: String) = try {
-    (Json.parse(s) \ "projects").asOpt[List[SnykProject]]
-  } catch {
-    case _: JsonParseException => None
-  }
+  def parseJsonToOrganisationList(s: String): Attempt[List[SnykOrganisation]] = parseJsonToObject("organisations", s, body => {(Json.parse(body) \ "orgs").validate[List[SnykOrganisation]]} )
```
adamnfish: Very long line (implementation on the next line again). there's also redundant `{  ...  }` around the `(Json.parse ... )` bit. Space before the closing `)` looks out of place as well.

hq/app/logic/SnykDisplay.scala
```diff
@@ -1,81 +1,63 @@
...
-  def parseProjectVulnerabilities(projects: List[String])(implicit ec:ExecutionContext): Attempt[List[SnykProjectIssues]] = {
-    val projectVulnerabilitiesList = projects.map( s => {
-      val projectVulnerabilities = parseJsonToProjectVulnerabilities(s)
-      val error = parseJsonToError(s)
-      val f = Failure(s"Unable to find project vulnerabilities from $s", s"Could not read Snyk response (${error.error})", 502, None, None)
-      Attempt.fromOption(projectVulnerabilities, FailedAttempt(f))
-    })
-    Attempt.traverse(projectVulnerabilitiesList)(a => a)
-  }
+  def parseJsonToProjectIdList(s: String): Attempt[List[SnykProject]]         = parseJsonToObject("project ids", s, body => (Json.parse(body) \ "projects").validate[List[SnykProject]] )
```
adamnfish: Very long line, but no redundant `{ ... }` here :-)

hq/app/logic/SnykDisplay.scala
```diff
@@ -1,81 +1,63 @@
...
+  private def parseJsonToError(s: String): SnykError = Try(Json.parse(s).validateOpt[SnykError]) match {
+    case Success(JsSuccess(Some(error), _)) => error
+    case _ => SnykError(s)
   }
 
-  private def parseJsonToOrganisationList(s: String): Option[List[SnykOrganisation]] = try {
-    (Json.parse(s) \ "orgs").asOpt[List[SnykOrganisation]]
-  } catch {
-    case _: JsonParseException => None
+  def parseJsonToObject[A](label: String, rawJson: String, f: (String => JsResult[A])): Attempt[A] = Try(f(rawJson)) match {
```
adamnfish: You don't need `( ... )` around `f`'s type.

hq/app/logic/SnykDisplay.scala
adamnfish: *May* be clearer with `_` instead of `body => ... body ...`, see what you think!

hq/app/logic/SnykDisplay.scala
```diff
@@ -1,81 +1,63 @@
...
+  def parseProjectVulnerabilitiesBody(s: String): Attempt[SnykProjectIssues]  = parseJsonToObject("project vulnerabilities", s, body => Json.parse(body).validate[SnykProjectIssues])
 
-  def linkToSnykProject(snykProjectIssues: SnykProjectIssues, queryString: Option[String]): String = {
-    snykProjectIssues.project match {
-      case Some(project: SnykProject) if project.organisation.nonEmpty =>
-        s"https://snyk.io/org/${project.organisation.get.name}/project/${project.id}/${queryString.getOrElse("")}"
-      case _ => ""
-    }
-  }
+  def parseProjectVulnerabilities(bodies: List[String])(implicit ec: ExecutionContext): Attempt[List[SnykProjectIssues]] = Attempt.traverse(bodies)(parseProjectVulnerabilitiesBody)
```
adamnfish: Nice traverse :-)

hq/app/logic/SnykDisplay.scala
```diff
@@ -1,81 +1,63 @@
...
+    case Some(project) if project.organisation.nonEmpty =>
+      s"https://snyk.io/org/${project.organisation.get.name}/project/${project.id}/${queryString.getOrElse("")}"
+    case _ => ""
   }
 
-  def labelProjects(projects: List[SnykProject], responses: List[SnykProjectIssues]): List[SnykProjectIssues] = {
-    projects.zip(responses).map(a => a._2.copy(project = Some(a._1)))
-  }
+  def labelOrganisations(orgAndProjects: List[((SnykOrganisation, String), List[SnykProject])]): List[SnykProject] =
+    orgAndProjects.flatMap{ case ((organisation, _), projects) => projects.map(project => project.copy(organisation = Some(organisation)))}
```
adamnfish: Long line! Space after the `flatMap` too. It's either a simple inline fn tight in with `()`, or a more complex one spaced out with `{}`

```scala
orgAndProjects.flatMap { case ((organisation, _), projects) =>
  projects.map(project => project.copy(organisation = Some(organisation)))
}
```
`project` could be underscore here now if you like, since we're right next to `projects.map` so we know it's a project.
```scala
orgAndProjects.flatMap { case ((organisation, _), projects) =>
  projects.map(_.copy(organisation = Some(organisation)))
}
```

hq/app/logic/SnykDisplay.scala
```diff
@@ -1,81 +1,63 @@
...
+
+  def labelProjects(projects: List[SnykProject], responses: List[SnykProjectIssues]): List[SnykProjectIssues] = projects.zip(responses).map(a => a._2.copy(project = Some(a._1)))
```
adamnfish: I'd be tempted to pattern match the argument here rather than use `a` and `_n`. `a` is a value of type `a` so certainly not the right name here, but easier to give the parts a proper name. I worked it out from the type signature, but it was a little harder than it needed to be.

```scala
projects.zip(responses).map { case (issues, project) =>
  issues.copy(project = Some(project))
}
```

Are you sure these lists have the same length? I don't see any tests for this which makes me wonder what happens when the lengths differ.

hq/app/model/Serializers.scala
```diff
@@ -7,7 +7,19 @@ object Serializers {
 
   implicit val snykErrorFormat: Reads[SnykError] = Json.reads[SnykError]
 
-  implicit val snykOrgFormat: Reads[SnykOrganisation] = Json.reads[SnykOrganisation]
+  implicit val snykGroupFormat: Reads[SnykGroup] = (
```
adamnfish: I think these can be automatically derived for you. (Less code is better code!)

```scala
implicit val snykGroupFormat: Reads[SnykGroup] = Json.reads[SnykGroup]
implicit val snykOrgFormat: Reads[SnykOrganisation] = Json.reads[SnykOrganisation]
```

This is true iff the case class's field names exactly match the JSON field names.

package.json
aug24: My mistake, I thought that was correct.  I'll fix.

hq/app/logic/SnykDisplay.scala
aug24: Done, although I reckon it's (project, issues)...

hq/app/model/Serializers.scala
aug24: SnykGroup can, SnykOrganisation can't.  I have tests to prove it :)

Done.

hq/app/api/Snyk.scala
aug24: Done.

hq/app/aws/ec2/EC2.scala
```diff
@@ -93,7 +93,7 @@ object EC2 {
   private [ec2] def sortSecurityGroupsByInUse(sgsPortFlags : List[SGOpenPortsDetail], sgUsage: Map[String, Set[SGInUse]]) = {
     sgsPortFlags
       .map(sgOpenPortsDetail => sgOpenPortsDetail -> sgUsage.getOrElse(sgOpenPortsDetail.id, Set.empty))
-      .sortWith{ case  ((_, s1), (_, s2)) => s1.size > s2.size }
+      .sortWith { case  ((_, s1), (_, s2)) => s1.size > s2.size }
```
adamnfish: two spaces after case instead of one

hq/app/logic/SnykDisplay.scala
```diff
@@ -1,81 +1,74 @@
...
-      org <- orglist.find(_.name == requiredOrganisation.value)
-    } yield org
-    val error = parseJsonToError(s)
-    val f = Failure(s"Unable to find organisation from $s", s"Could not read Snyk response (${error.error})", 502, None, None)
-    Attempt.fromOption(organisation, FailedAttempt(f))
+  def parseOrganisations(rawJson: String, snykGroupId: SnykGroupId)(implicit ec: ExecutionContext): Attempt[List[SnykOrganisation]] = {
+    for {
+      organisationList <- parseJsonToOrganisationList(rawJson)
+      guardianOrganisationList = organisationList filter {
+        case SnykOrganisation(_, _, Some(group)) => group.id==snykGroupId.value
```
adamnfish: spaces around `==` would be nice

hq/app/logic/SnykDisplay.scala
```diff
@@ -1,81 +1,74 @@
...
-    (Json.parse(s) \ "orgs").asOpt[List[SnykOrganisation]]
-  } catch {
-    case _: JsonParseException => None
+  def parseJsonToObject[A](label: String, rawJson: String, f: String => JsResult[A]): Attempt[A] = Try(f(rawJson)) match {
+    case Success(JsSuccess(parsedObject, _)) => Attempt.Right(parsedObject)
+    case Success(JsError(e)) =>
+      val error = parseJsonToError(rawJson)
+      val failure = Failure(s"Unable to find $label from $rawJson: $e", s"Could not read Snyk response (${error.error})", 502, None, None)
+      Attempt.Left(failure)
+    case scala.util.Failure(e) =>
```
adamnfish: ~Better to import Failure, no need to full qualify it in the case statement.~

Ah, I see it is because it clashes name.

hq/app/model/Serializers.scala
```diff
@@ -7,7 +7,15 @@ object Serializers {
...
-  implicit val snykOrgFormat: Reads[SnykOrganisation] = Json.reads[SnykOrganisation]
+  implicit val snykGroupFormat: Reads[SnykGroup] = Json.reads[SnykGroup]
+
+  implicit val snykOrgFormat: Reads[SnykOrganisation] = (
+    (JsPath \ "name").read[String]
+      and
+      (JsPath \ "id").read[String]
+      and
+      (JsPath \ "group").readNullable[SnykGroup]
+    )(SnykOrganisation.apply _)
```
adamnfish: This one can probably be derived, but you'd need to call the property `group` (instead of `groupOpt`) so no worries if you prefer it this way

hq/app/controllers/SecurityGroupsController.scala
```diff
@@ -26,7 +26,7 @@ class SecurityGroupsController(val config: Configuration, cacheService: CacheSer
     Ok(views.html.sgs.sgs(sortedFlaggedSgs))
   }
 
-  def securityGroupsAccount(accountId: String) = authAction.async {
+  def securityGroupsAccount(accountId: String): Action[AnyContent] = authAction.async {
```
adamnfish: I know IntelliJ moans about annotating these but it's wrong. It's an action, we know what it does, fine to omit the annotation (particularly since it isn't very).

PR #110, opened 2018-03-13
Display AWS Inspector results

adamnfish:
## What does this change?

Adds AWS Inspector results for each account to the app.

### Homepage
homepage

### All accounts
all-inspector-results

### Single account
account-inspector-results

Yes, there is a change to `watched-account.template.yaml` to enable AWS Inspector calls.

**StackSet update required prior to deploy**

### DIFF
--- a/hq/app/logic/InspectorResults.scala
+++ b/hq/app/logic/InspectorResults.scala
@@ -10,7 +10,7 @@ import scala.collection.JavaConverters._
 
 object InspectorResults {
   private val tagMatch = "[\\w\\-_\\.]"
-  val RunNameMatch = s"AWSInspection-($tagMatch+)-($tagMatch+)-($tagMatch+)-[\\d]+".r
+  val RunNameMatch = s"AWSInspection-($tagMatch+)--($tagMatch+)--($tagMatch+)--[\\d]+".r
 
   def appId(assessmentRunName: String): (String, String, String) = {
     assessmentRunName match {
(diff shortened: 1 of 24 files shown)


hq/app/logic/InspectorResults.scala
```diff
@@ -0,0 +1,104 @@
...
+import model.InspectorAssessmentRun
+import org.joda.time.DateTime
+import utils.attempt.FailedAttempt
+
+import scala.collection.JavaConverters._
+
+
+object InspectorResults {
+  private val tagMatch = "[\\w\\-_\\.]"
+  val RunNameMatch = s"AWSInspection-($tagMatch+)-($tagMatch+)-($tagMatch+)-[\\d]+".r
```
adamnfish: This will change if the Lambda changes its format

shtukas: +1

adamnfish: We can change this separately and exclude non-Lambda runs at the same time

hq/app/logic/InspectorResults.scala
```diff
@@ -0,0 +1,108 @@
...
+import model.InspectorAssessmentRun
+import org.joda.time.DateTime
+import utils.attempt.FailedAttempt
+
+import scala.collection.JavaConverters._
+
+
+object InspectorResults {
+  private val tagMatch = "[\\w\\-_\\.]"
+  val RunNameMatch = s"AWSInspection-($tagMatch+)--($tagMatch+)--($tagMatch+)--[\\d]+".r
```
shtukas: Updated this with a new commit @adamnfish @aug24

aug24:
+1 with latest changes.

adamnfish:
Watched account update is out, so let's YOLO this out.

PR #112, opened 2018-03-16
Update AWS SDK version

adamnfish:
## What does this change?

Updates out of date AWS SDK.

## What is the value of this?

Hopefully this updates the vulnerable Jackson dependency and fixes the Snyk check.

Nope.


adamnfish:
^- looks like this didn't fix it.

Snyk says the AWS SDK is bringing the dependency in and this is the latest version of the AWS SDK. We could manually include the patched version to evict the version the AWS SDK depends on.

Let's chat IRL

adamnfish:
This is already out of date. We can raise another when we're ready.

PR #114, opened 2018-03-16
Properly parse our assessment runs

adamnfish:
## What does this change?

Changes the way we parse assessment runs for the Inspector data.

## What is the value of this?

We now exclude assessment runs that are in progress. We also only show assessments that were created and run by our automated system. Manual invocations will not appear here.

Nope.

## Any additional notes?

Nope.

### DIFF
--- a/hq/app/logic/InspectorResults.scala
+++ b/hq/app/logic/InspectorResults.scala
@@ -10,14 +10,14 @@ import scala.collection.JavaConverters._
 
 object InspectorResults {
   private val tagMatch = "[\\w\\-_\\.]"
-  val RunNameMatch = s"AWSInspection-($tagMatch+)-($tagMatch+)-($tagMatch+)-[\\d]+".r
+  val RunNameMatch = s"AWSInspection--($tagMatch+)--($tagMatch+)--($tagMatch+)--[\\d]+".r
 
-  def appId(assessmentRunName: String): (String, String, String) = {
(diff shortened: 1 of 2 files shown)


shtukas:
LGTM: +1

hq/app/logic/InspectorResults.scala
```diff
@@ -45,26 +45,37 @@ object InspectorResults {
...
-      completedAt = new DateTime(assessmentRun.getCompletedAt),
-      stateChangedAt = new DateTime(assessmentRun.getStateChangedAt),
-      dataCollected = assessmentRun.getDataCollected,
-      findingCounts = assessmentRun.getFindingCounts.asScala.toMap.mapValues(_.toInt)
-    )
+  /**
+    * Parses a *completed* assessment, if it matches the format used by our automatic inspection service.
+    */
+  private[logic] def parseCompletedAssessmentRun(assessmentRun: AssessmentRun): Option[InspectorAssessmentRun] = {
+    if (assessmentRun.getState != "COMPLETED" || assessmentRun.getDataCollected != true) {
```
aug24: if (!= || !=) A else B

Personally I'd rewrite to the opposite clause if (== && ==) B else A

aug24:
LGTM

hq/app/logic/InspectorResults.scala
adamnfish: Great shout, thanks

PR #115, opened 2018-03-16
Inspector parsing support empty tags

adamnfish:
## What does this change?

Support empty Stack/App/Stage tags when looking for inspector runs.

## What is the value of this?

The lambda runner will run *all combos* of stack/app/stage, including ones with missing tags. This will now pick all of them up.

Nope.

## Any additional notes?

Based on https://github.com/guardian/security-hq/pull/114/files <- merge that first


adamnfish:
I need to tidy this change up against the latest version of master.

PR #116, opened 2018-03-19
Filter on assessment runs

aug24:
## What does this change?

Filters on assessment run instead of assessment target when opening the aws console for an inspector result.

## What is the value of this?

It's what is expected.

No.


adamnfish:
:+1:  Excellent, thank you!

PR #118, opened 2018-03-21
Much reduced policy for SSM run-command only

aug24:
## What does this change?

Removes over-generous managed policy for SSM
Adds new, much reduced policy for ssm run-command only
Restructures the CFN Policy resources for clarity

## What is the value of this?

Worked example for other teams to follow

Yes

### Has the CloudFormation or StackSet update been completed? 

No

## Any additional notes?

No

### DIFF
--- a/cloudformation/security-hq.template.yaml
+++ b/cloudformation/security-hq.template.yaml
@@ -83,36 +83,81 @@ Resources:
           Action:
           - sts:AssumeRole
       Path: /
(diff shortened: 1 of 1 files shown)


adamnfish:
Change looks good.

What's the thinking behind splitting this up into loads of separate policies? As well as being lots more verbose, it means a user in the console needs to look in several places to see the permissions an instance has? I don't mind either way, just interested if there's a reason :-)

PR #121, opened 2018-03-21
Add completed-at to Inspector cards

aug24:
## What does this change?

Adds 'completed at' field to Inspector cards.

## What is the value of this?

Users can see how up to date the scan is

No

## Any additional notes?

No


aug24:
What else do we think we want to display?  I'd quite like a 'X days ago' field?

PR #125, opened 2018-04-18
Improved message merging

aug24:
## What does this change?

Improved message with account *name* in it.

## What is the value of this?

This is a precursor to linking directly to the SG in the AWS console via janus.

No

## Any additional notes?

No


lambda/common/src/main/scala/com/gu/hq/AWS.scala
```diff
@@ -27,31 +28,37 @@ object AWS {
...
     elbClient.describeLoadBalancers(request)
   }
 
-  def accountNumber(stsClient: AWSSecurityTokenServiceAsync) = {
+  def accountNumber(stsClient: AWSSecurityTokenServiceAsync): String = {
     stsClient.getCallerIdentity(new GetCallerIdentityRequest()).getAccount
   }
+
+  def accountName(s3Client: AmazonS3, accountNumber: String):String = {
+    def accountsJson = scala.io.Source.fromInputStream(s3Client.getObject("guardian-dist", "guardian/PROD/accounts").getObjectContent).mkString
```
adamnfish: We're mixing logic and IO here. It'd be great if we could test this code?

Fetching the AWS account name config should be a separate step to the logic of getting the account name. This function should *just* do the IO. All the rest can be pure tested code.

lambda/common/src/main/scala/com/gu/hq/AWS.scala
```diff
@@ -27,31 +28,37 @@ object AWS {
...
     val request = new DescribeLoadBalancersRequest()
     elbClient.describeLoadBalancers(request)
   }
 
-  def accountNumber(stsClient: AWSSecurityTokenServiceAsync) = {
+  def accountNumber(stsClient: AWSSecurityTokenServiceAsync): String = {
     stsClient.getCallerIdentity(new GetCallerIdentityRequest()).getAccount
   }
+
+  def accountName(s3Client: AmazonS3, accountNumber: String):String = {
```
adamnfish: Lint: Space between `:` and `String`

lambda/common/src/main/scala/com/gu/hq/lambda/model/JSON.scala
```diff
@@ -5,6 +5,7 @@ import org.joda.time.format.DateTimeFormat
 import play.api.libs.json.{Json, Reads}
 
 
+//noinspection TypeAnnotation
```
adamnfish: What's this comment for?

lambda/common/src/main/scala/com/gu/hq/lambda/model/events.scala
```diff
@@ -44,3 +44,8 @@ case class Relationship(
   resourceName: Option[String],
   name: String
 )
+
+case class AccountMap(
+  accountNumber: String,
+  accountName:String,
```
adamnfish: Lint: trailing comma

lambda/security-groups/src/main/scala/com/gu/hq/Lambda.scala
```diff
@@ -23,9 +24,10 @@ class Lambda extends RequestHandler[ConfigEvent, Unit] with StrictLogging {
       sgConfiguration <- ConfigEventLogic.sgConfiguration(configurationItem.configuration)
       loadBalancers = AWS.describeLoadBalancers(elbClient)
       account = AWS.accountNumber(stsClient)
+      accountName = AWS.accountName(s3Client, account)
```
adamnfish: As before, this is two steps.

1. Fetch the Account name config
2. Apply the logic to lookup the account name from that config

lambda/security-groups/src/main/scala/com/gu/hq/Notifier.scala
```diff
@@ -33,11 +33,11 @@ object Notifier extends StrictLogging {
     }
   }
 
-  def createNotification(groupId: String, targetTags: List[Tag], accountId: String, regionName: String): Notification = {
+  def createNotification(groupId: String, targetTags: List[Tag], accountId: String, accountName:String, regionName: String): Notification = {
     val actions = List(
       Action("View in AWS Console", s"https://$regionName.console.aws.amazon.com/ec2/v2/home?region=$regionName#SecurityGroups:search=$groupId")
     )
-    val message = s"Warning: Security group '$groupId' is open to the world"
+    val message = s"Warning: Security group '$groupId' in account '$accountName' is open to the world"
```
adamnfish: Minor point / bikeshed

Maybe check the formatting Anghammarad supports? Can we do better than quoted?

Email supports markdown, so bold might be nice. I can't remember if Google Hangouts doesn't support much formatting in cards, but IIRC bold is fine. The Anghammarad code shows the transformations it does.

lambda/security-groups/src/test/scala/com/gu/hq/NotifierTest.scala
```diff
@@ -58,28 +58,39 @@ class NotifierTest extends FreeSpec with Matchers with OptionValues {
...
       }
     }
 
     "adds Account ID to target" in {
-      val notification = createNotification(groupId, targetTags, accountId, regionName)
+      val notification = createNotification(groupId, targetTags, accountId, accountName, regionName)
       notification.target should contain(AwsAccount(accountId))
     }
 
+    "adds Account Name to message" in {
```
adamnfish: Good additions!

lambda/security-groups/src/main/scala/com/gu/hq/Notifier.scala
adamnfish: Summary: I am basically saying I think `*string*` will display as bold on both emails and chat messages..

adamnfish: https://github.com/guardian/anghammarad/blob/master/anghammarad/src/main/scala/com/gu/anghammarad/messages/Messages.scala#L51

The message is markdown, these are the fixes Anghammarad does because Google Hangouts Chat only supports a weird subset of tags.

lambda/common/src/main/scala/com/gu/hq/lambda/model/events.scala
```diff
@@ -44,3 +44,8 @@ case class Relationship(
   resourceName: Option[String],
   name: String
 )
+
+case class AccountMap(
```
adamnfish: Not sure this is in the right place?

This is nothing to do with events, so why not here?

https://github.com/guardian/security-hq/blob/master/lambda/common/src/main/scala/com/gu/hq/lambda/model/aws.scala

lambda/common/src/main/scala/com/gu/hq/AWS.scala
aug24: Done.

lambda/common/src/main/scala/com/gu/hq/AWS.scala
aug24: Done.

lambda/common/src/main/scala/com/gu/hq/lambda/model/JSON.scala
aug24: Suppress IntelliJ warnings.

lambda/common/src/main/scala/com/gu/hq/lambda/model/events.scala
aug24: Moved to new file accounts.scala.  It didn't really fit with anything else either.

lambda/common/src/main/scala/com/gu/hq/lambda/model/events.scala
aug24: Done.

lambda/security-groups/src/main/scala/com/gu/hq/Lambda.scala
aug24: Done.

lambda/security-groups/src/main/scala/com/gu/hq/Notifier.scala
aug24: Done.

lambda/security-groups/src/main/scala/com/gu/hq/Lambda.scala
```diff
@@ -3,7 +3,7 @@ package com.gu.hq
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.lambda.runtime.events.ConfigEvent
 import com.amazonaws.services.lambda.runtime.{Context, RequestHandler}
-import com.gu.hq.lambda.ConfigEventLogic
+import com.gu.hq.lambda.JsonParsing
```
adamnfish: I don't see a corresponding change for the tests, have I missed something from a little computer? At least the test string label and probably the class, if not the tests themselves as well?

aug24: Yes, the test was renamed to JsonParsingTest.scala.

aug24: And, of course, the tests still pass.

PR #127, opened 2018-04-24
Move to Quartz scheduler

aug24:
## What does this change?

Update various caches by Quartz Scheduler instead of Play.

## What is the value of this?

Play threads were hanging around and taking up memory, leading to crashes, leading to instance cycling.  This change does not appear to do that.

No

## Any additional notes?

No


shtukas:
This explains the "sealed trait EnumVal" question this morning ;)

PR #130, opened 2018-04-27
Re-use all AWS clients

aug24:
## What does this change?

Never create AWS Clients on demand, but re-use pre-created clients instead.

## What is the value of this?

AWS Clients wrap thread pools and have their own root threads, and thus are not garbage collected.  If a new client is created, it must be explicitly shutdown to avoid a leak.  Alternatively, create one per account per region at the start, and re-use it forevery.  This is what we have done.

No

## Any additional notes?

No

### DIFF
--- a/hq/app/aws/AWS.scala
+++ b/hq/app/aws/AWS.scala
@@ -30,14 +30,13 @@ object AWS {
     )
   }
 
-  private def client[T, S <: AwsClientBuilder[S, T]](
+  private def clientMapping[T, S <: AwsClientBuilder[S, T]](
     account: AwsAccount,
     region: Regions,
     builder: AwsClientBuilder[S, T]
   ): ((String, Regions), T) = {
-    val auth = credentialsProvider(account)
     val client = builder
-      .withCredentials(auth)
+      .withCredentials(credentialsProvider(account))
       .withRegion(region)
       .build()
     (account.id, region) -> client
@@ -51,7 +50,7 @@ object AWS {
     val list = for {
       account <- Config.getAwsAccounts(configuration)
       region <- regionList
-    } yield client[T, S](account, region, builder)
+    } yield clientMapping[T, S](account, region, builder)
     Map(list: _*)
   }
 
--- a/hq/app/aws/AWS.scala
+++ b/hq/app/aws/AWS.scala
@@ -18,8 +18,6 @@ import utils.attempt.{Attempt, Failure}
 
 object AWS {
 
-  //  availableRegions <- EC2.getAvailableRegions(regionClient)
-
   private def credentialsProvider(account: AwsAccount): AWSCredentialsProviderChain = {
     new AWSCredentialsProviderChain(
       new STSAssumeRoleSessionCredentialsProvider.Builder(account.roleArn, "security-hq").build(),
--- a/hq/app/aws/AWS.scala
+++ b/hq/app/aws/AWS.scala
@@ -32,7 +32,7 @@ object AWS {
     )
   }
 
-  private def clients[A, B <: AwsClientBuilder[B, A]](
+  private[aws] def clients[A, B <: AwsClientBuilder[B, A]](
     builder: AwsClientBuilder[B, A],
     configuration: Configuration,
     regionList: Regions*
--- a/hq/app/aws/AWS.scala
+++ b/hq/app/aws/AWS.scala
@@ -2,6 +2,7 @@ package aws
 
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
 import com.amazonaws.auth.{AWSCredentialsProviderChain, STSAssumeRoleSessionCredentialsProvider}
+import com.amazonaws.client.builder.AwsClientBuilder
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.cloudformation.{AmazonCloudFormationAsync, AmazonCloudFormationAsyncClientBuilder}
 import com.amazonaws.services.ec2.{AmazonEC2Async, AmazonEC2AsyncClientBuilder}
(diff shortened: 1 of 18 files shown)


hq/app/aws/ec2/EC2.scala
```diff
@@ -187,10 +192,18 @@ object EC2 {
...
   }
 
-  private[ec2] def getVpcs(account: AwsAccount, flaggedSgs: List[SGOpenPortsDetail])(vpcsDetailsF: AmazonEC2Async => Attempt[Map[String, Vpc]])(implicit ec: ExecutionContext): Attempt[Map[String, Vpc]] = {
+  private[ec2] def getVpcs(account: AwsAccount, flaggedSgs: List[SGOpenPortsDetail], ec2Clients: Map[(String, Regions), AmazonEC2Async])(vpcsDetailsF: AmazonEC2Async => Attempt[Map[String, Vpc]])(implicit ec: ExecutionContext): Attempt[Map[String, Vpc]] = {
     Attempt.traverse(flaggedSgs.map(_.region).distinct) { region =>
-      val ec2Client = client(account, Regions.fromName(region))
-      vpcsDetailsF(ec2Client)
+      val awsRegion = Regions.fromName(region)
+      for {
+        ec2Client <- Attempt.fromOption(ec2Clients.get((account.id, awsRegion)), FailedAttempt(Failure(
```
adamnfish: Why is this failure inlined here? I thought there was an `EC2.client` method that takes care of this?

hq/app/aws/iam/IAMClient.scala
```diff
@@ -1,28 +1,32 @@
...
-    val auth = AWS.credentialsProvider(account)
-    AmazonIdentityManagementAsyncClientBuilder.standard()
-      .withCredentials(auth)
-      .withRegion(Option(Regions.getCurrentRegion).getOrElse(region).getName)
-      .build()
+
+  def client(iamClients: Map[(String, Regions),  AmazonIdentityManagementAsync], awsAccount: AwsAccount): Attempt[ AmazonIdentityManagementAsync] = {
+    val region = Regions.US_EAST_1
+    Attempt.fromOption(iamClients.get((awsAccount.id, region)), FailedAttempt(Failure(
+      s"No AWS Trusted Advisor Client exists for ${awsAccount.id} and $region",
```
adamnfish: :-/

This is IAM, not TrustedAdvisor.

hq/app/controllers/InspectorController.scala
```diff
@@ -16,7 +16,8 @@ import utils.attempt.PlayIntegration.attempt
 
 import scala.concurrent.ExecutionContext
 
-class InspectorController(val config: Configuration,
+class InspectorController(val inspectorClients: Map[(String, Regions), AmazonInspectorAsync],
```
adamnfish: Why are you adding the clients here? They aren't used AFAICT.

hq/app/utils/attempt/Attempt.scala
```diff
@@ -107,7 +107,7 @@ object Attempt {
     *
     * Combines the behaviours of labelledTraverse and traverseWithFailures.
     */
-  def labelledTaverseWithFailures[A, B](as: List[A])(f: A => Attempt[B])(implicit ec: ExecutionContext): Attempt[List[(A, Either[FailedAttempt, B])]] = {
+  def labelledTraverseWithFailures[A, B](as: List[A])(f: A => Attempt[B])(implicit ec: ExecutionContext): Attempt[List[(A, Either[FailedAttempt, B])]] = {
```
adamnfish: Good catch

hq/app/aws/AWS.scala
```diff
@@ -20,4 +28,61 @@ object AWS {
...
+      val inspectorClient = AmazonInspectorAsyncClientBuilder.standard()
+        .withCredentials(auth)
+        .withRegion(region)
+        .build()
+      (account.id, region) -> inspectorClient
+    }
+  )).toMap
+
+
+  def ec2Clients(configuration: Configuration): Map[(String, Regions), AmazonEC2Async] = Config.getAwsAccounts(configuration).flatMap(account => Regions.values().map(
```
adamnfish: I'm not crazy about the duplication here (on top of the codestyle, as mentioned above).

Some options I have considered follow. I think 3 is my favourite for the ones with all regions. For the ones that only take a single region there's no need to do the inner map. A simpler 1 or 2 would be best for them.

```scala

  def inspectorClients(configuration: Configuration): Map[(String, Regions), AmazonInspectorAsync] = {
    Config.getAwsAccounts(configuration).flatMap { account =>
      Regions.values().map { region =>
        val inspectorClient = AmazonInspectorAsyncClientBuilder.standard()
          .withCredentials(AWS.credentialsProvider(account))
          .withRegion(region)
          .build()
        (account.id, region) -> inspectorClient
      }
    }.toMap
  }

  def inspectorClients2(configuration: Configuration): Map[(String, Regions), AmazonInspectorAsync] = {
    (for {
      account <- Config.getAwsAccounts(configuration)
      region <- Regions.values()
    } yield {
      val inspectorClient = AmazonInspectorAsyncClientBuilder.standard()
        .withCredentials(AWS.credentialsProvider(account))
        .withRegion(region)
        .build()
      (account.id, region) -> inspectorClient
    }).toMap
  }

  def inspectorClients3(configuration: Configuration): Map[(String, Regions), AmazonInspectorAsync] = {
    clientsForAllRegions(configuration) { (account, region) =>
      AmazonInspectorAsyncClientBuilder.standard()
        .withCredentials(AWS.credentialsProvider(account))
        .withRegion(region)
        .build()
    }
  }

  def clientsForAllRegions[A](configuration: Configuration)(makeClient: (AwsAccount, Regions) => A): Map[(String, Regions), A] = {
    val clients = for {
      account <- Config.getAwsAccounts(configuration)
      region <- Regions.values()
    } yield {
      (account.id -> region) -> makeClient(account, region)
    }
    clients.toMap
  }

```

hq/app/aws/AWS.scala
```diff
@@ -20,4 +28,61 @@ object AWS {
       Failure.awsAccountNotFound(accountId).attempt
     )
   }
+
+  def inspectorClients(configuration: Configuration): Map[(String, Regions), AmazonInspectorAsync] = Config.getAwsAccounts(configuration).flatMap(account => Regions.values().map(
+    region => {
+      val auth = AWS.credentialsProvider(account)
```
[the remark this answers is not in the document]
aug24: Done.

hq/app/aws/cloudformation/CloudFormation.scala
```diff
@@ -1,52 +1,52 @@
...
+    cfnClients: Map[(String, Regions), AmazonCloudFormationAsync],
+    ec2Clients: Map[(String, Regions), AmazonEC2Async]
+  )(implicit ec: ExecutionContext): Attempt[List[AwsStack]] = {
     for {
+      regionClient <- EC2.client(ec2Clients, account, Regions.EU_WEST_1)
       availableRegions <- EC2.getAvailableRegions(regionClient)
-      regions = availableRegions.map(region => Region.getRegion(Regions.fromName(region.getRegionName)))
-      stacks <- Attempt.flatTraverse(regions)(region => getStacks(account, region))
+      regions = availableRegions.map(region => Regions.fromName(region.getRegionName))
+      stacks <- Attempt.flatTraverse(regions)(region => getStacks(account, region, cfnClients))
```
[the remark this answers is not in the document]
katebee: The AWS SDK `Regions.values()` will get you all the regions but these are not necessarily ones that you or the application will have access to. 😅 

This was fun to discover, since `availableRegions <- EC2.getAvailableRegions(regionClient)` was added to work around the existence of `GovCloud("us-gov-west-1")` which returns **access denied** errors for some strange reason (what could that be?) when trying to `describeResources` for the region. 
🙈 🙉 🙊

So you could make clients for all regions except the US GovCloud, then hopefully all of the clients will have the required permissions to make the API calls the application wants to.

I feel like I have notes on this somewhere (which I will try to find) so this may not or not longer apply. 🙃

hq/app/aws/AWS.scala
```diff
@@ -1,13 +1,26 @@
...
+import config.Config
 import model.AwsAccount
+import play.api.Configuration
 import utils.attempt.{Attempt, Failure}
 
 
 object AWS {
-  def credentialsProvider(account: AwsAccount): AWSCredentialsProviderChain = {
+
+  //  availableRegions <- EC2.getAvailableRegions(regionClient)
```
adamnfish: What's this for?

hq/app/aws/AWS.scala
```diff
@@ -20,4 +33,39 @@ object AWS {
       Failure.awsAccountNotFound(accountId).attempt
     )
   }
+
+  private def clients[T, S <: AwsClientBuilder[S, T]](
```
adamnfish: `A` and `B` is a little more conventional in modern Scala but that's a super minor point and I don't mind at all here (just an FYI)

hq/app/aws/inspector/Inspector.scala
```diff
@@ -44,14 +31,19 @@ object Inspector {
...
+    Attempt.labelledTraverseWithFailures(accounts)(Inspector.inspectorRuns(inspectorClients))
   }
 
-  def inspectorRuns(account: AwsAccount)(implicit ec: ExecutionContext): Attempt[List[InspectorAssessmentRun]] = {
+  def inspectorRuns(inspectorClients: Map[(String, Regions), AmazonInspectorAsync])(account: AwsAccount)(implicit ec: ExecutionContext): Attempt[List[InspectorAssessmentRun]] = {
     val region = Regions.EU_WEST_1  // we only automatically run inspections in Ireland
-    val inspectorClient = client(account, region)
+
     for {
+      inspectorClient <- Attempt.fromOption(inspectorClients.get((account.id, region)), FailedAttempt(Failure(
```
adamnfish: Why not a `clients` function like we have for the other services?

hq/app/aws/AWS.scala
adamnfish: Maybe be good to have a test for this? Make sure it actually creates clients for the provided regions.

hq/app/aws/inspector/Inspector.scala
adamnfish: (I generally try to avoid implementation in a for-comp, which is a good rule of thumb for leading to code that separates logic from IO)

hq/app/aws/AWS.scala
aug24: Damn, that was a reminder to look at available regions.  I'll remove.

hq/app/aws/AWS.scala
aug24: Done anyway.  T is conventional in Java.

hq/app/aws/inspector/Inspector.scala
aug24: Done.

hq/app/aws/AWS.scala
aug24: Done.

hq/app/logic/InspectorResults.scala
```diff
@@ -117,7 +117,7 @@ object InspectorResults {
     assessmentRuns.flatMap(_.findingCounts.values).sum
   }
 
-  def completedDaysAgo(assessmentRun: InspectorAssessmentRun): Int = Days.daysBetween(new DateTime(), assessmentRun.completedAt).getDays
+  def completedDaysAgo(assessmentRun: InspectorAssessmentRun): Int = Days.daysBetween(assessmentRun.completedAt, new DateTime()).getDays
```
adamnfish: Ah yes. Irrelevant change, but a good fix.

hq/app/services/CacheService.scala
```diff
@@ -112,7 +126,7 @@ class CacheService(config: Configuration, lifecycle: ApplicationLifecycle, envir
     setUpQuartzScheduleJob(Cache.AWSInspector, "45 20 1,7,13,19 * * ?")
   }
 
-  private def setUpQuartzScheduleJob(cache: Cache.EnumVal, cronString: String) = {
+  private def setUpQuartzScheduleJob(cache: Cache.EnumVal, cronString: String): Unit = {
```
adamnfish: Is this required? I thought we were keeping the obvservables and just adding the client caches?

PR #133, opened 2018-05-01
Better name for the security group notification lambda

shtukas:


adamnfish:
Much better!

shtukas:
I was mostly interested in the change hitting `master`. It is going to propagate next time I update the lambda everywhere with StackSets. No rush.

PR #135, opened 2018-05-02
Stop lambda spamming on non CIDR changes

aug24:
## What does this change?

The lambda will only notify on changes which include an IpConfiguration change.

## What is the value of this?

Stop notifying on SG changes which do not changes the CIDR (only the Relationship)

We will bump the version and redeploy

## Any additional notes?

No


shtukas:
Nice one. The move to `sealed trait Relevance` feels much cleaner

PR #137, opened 2018-05-03
Increase instance size

aug24:
## What does this change?

Reverts my change back to micro, so we go to t2.small permanently.

## What is the value of this?

Previously we had upscaled to small because we suspected memory issues.  Having resolved those, we believed we could downscale again, back to micro.  However it appears that SHQ has grown, and it is now killed on a micro instance.  upscaling to small permanently fixes the issue.

Yes

## Have you committed your changes to the CloudFormation templates?

Yes

## Has the CloudFormation or StackSet update been completed? 

No

## Any additional notes?

No


shtukas:
+1, in a case like this feel free to self approve and merge :)

PR #144, opened 2018-08-06
Request and display public buckets report from Trusted Advisor

katebee:
### 👉 Fetch and display S3 buckets report from Trusted Advisor

The all accounts and individual account pages now have another panel that will display details from the S3 buckets report:

screen shot 2018-08-06 at 15 10 46

### 👉 Help modal for the S3 buckets

This contains basic guidance for any security issues that are identified, which can be displayed by clicking the help icon:

screen shot 2018-08-06 at 15 13 22

### 👉 Updates to the navigation and homepage

Consistent design of icons and links, and now the bucket report will be discoverable:

screen shot 2018-08-06 at 15 09 37

## What is the value of this?

- Visibility of S3 buckets that have open access 🔓 

- Centralised reporting on the security status of multiple AWS accounts 🔍 

- Early detection of misconfigured buckets 🔐 

- Transparency across Digital of which buckets are where 😈 

- Encourage best security practices 🚀 

Nope

## Any additional notes?

It would be great to expand this to use the `get-bucket-policy` API call so that each policy could be searched to detect global read or global write...

Hopefully this will not hit the API rate limit, however testing on 5 accounts locally was fine and this PR is only adding one request per account.


shtukas:
Thank for that Kate!
Merge at will :)

PR #148, opened 2018-12-19
Fix region assumptions, increase instance size and overhaul AWS SDK failure logging

sihil:
## What does this change?
Security HQ has been down / flaky for a while and was tricky to troubleshoot.

The initial failure was down to AWS launching a new region which didn't exist in the SDK version we were using:
```scala
  describeAwsRegions.map(Regions.fromName)
```

It turns out this was done for good reasons - we want to process all regions but not Chinese or the US GovCloud as both of these need special AWS accounts. The `describeAvailableRegions` command will only include regions you can access, but will also include regions that are not in your SDK.

In initially fixing this we accidentally tried to run security HQ against China/GovCloud and this was hard to troubleshoot as there was no context of the account or region at the call site.

To solve this we introduced context via an `AwsClient` case class which wraps every AWS client and includes the `AwsAccount` and `Region`. When an error occurs, this context is now also included in logs and failures.

Finally the instance was running out of heap at startup. This seem to be transient as the size settles down after it has initialised. Further profiling should be done to determine the root cause but in the interim this problem has been solved by moving Security HQ to a `t2.medium`.

## What is the value of this?
Security HQ wasn't running and now it is.

No, although it should now be possible to remove permission to call DescribeAvailableRegion

### DIFF
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -83,7 +83,7 @@ class AppComponents(context: Context)
 
   Logger.info(s"Polling in the following regions: ${availableRegions.map(_.getName).mkString(", ")}")
 
-  val regionsNotInSdk: Set[String] = Regions.values().map(_.getName).toSet -- availableRegions.map(_.getName).toSet
+  val regionsNotInSdk: Set[String] = availableRegions.map(_.getName).toSet -- Regions.values().map(_.getName).toSet
   if (regionsNotInSdk.nonEmpty) {
     Logger.warn(s"Regions exist that are not in the current SDK (${regionsNotInSdk.mkString(", ")}), update your SDK!")
   }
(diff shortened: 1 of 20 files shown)


philmcmahon:
Looks good!

sihil:
I'm afraid that this isn't quite working. Relied too much on the tests. Working on fixing properly now.

sihil:
OK - this is actually working now. I'd screwed up the region logic.

hq/app/AppComponents.scala
```diff
@@ -53,12 +63,37 @@ class AppComponents(context: Context)
     }
   }
 
+  // the aim of this is to get a list of available regions that we are able to access
+  // note that:
+  //  - Regions.values() returns Chinese and US government regions that are not accessible with the same AWS account
+  //  - available regions can return regions that are not in the SDK and so Regions.findName will fail
+  // to solve these we return the intersection of available regions and regions.values()
```
katebee: 👍 This is a really helpful write up of the issue - thanks!

sihil:
Happy to merge. Has been running on PROD for ages now anyway....

PR #151, opened 2019-07-23
Add bucket encryption status to Security HQ

katebee:
## What is the value of this?

Adds bucket encryption status to Security HQ so that there is a single place we can view all buckets that do not have encryption-at-rest enabled.

> Amazon S3 default encryption provides a way to set the default encryption behavior for an S3 bucket. You can set default encryption on a bucket so that all objects are encrypted when they are stored in the bucket.

https://docs.aws.amazon.com/AmazonS3/latest/dev/bucket-encryption.html

#### 👉 Add S3 clients and fetch S3 bucket default encryption status

Unfortunately, the encryption status for a bucket requires yet another set of AWS clients. This PR adds S3 clients so that we can call `s3:GetEncryptionConfiguration` on each bucket.

If there is no bucket encryption, AWS will return  an error. The current code assumes that the bucket is not encrypted if `getBucketEncryption` throws an `AmazonS3Exception` and this seems to be the only way to check if the bucket has default encryption.

#### 👉 Reflect encryption in bucket report status and update tests

The TA report actually includes all buckets, even if they are status `Green` in Trusted Advisor and do not have any public access. However, we only want to display buckets that may have configuration issues or overly permissive access policies. 

To ensure we can still display the buckets that do not have encryption-at-rest enabled, these buckets are being given a reportStatus of `Blue`

#### 👉 Add encryption status to the buckets view

Since it is likely many buckets will not have default encryption, and this may clutter the view, this adds a toggle to show and hide buckets where the only issue is the encryption status.

Buckets with no issues are now excluded from the page entirely. SHQ will only display buckets with configuration issues or overly permissive access policies.

Screenshot 2019-07-23 at 15 57 31

#### 👉 Add help information to the buckets modal

Clicking the help icon on the bucket card will bring up the help modal. 

Screenshot 2019-07-23 at 17 26 55

The modal explains that: 

> Server-side encryption is about data encryption at rest. When you use server-side encryption, Amazon S3 encrypts an object before saving it to disk and decrypts it when you download the object.

https://docs.aws.amazon.com/AmazonS3/latest/dev/security-best-practices.html#server-side

Yes, the watched-account role now needs `s3:GetEncryptionConfiguration`

- [x] update the StackSets

## Any additional notes?

When enabling default encryption, there is **no change** to the encryption of the objects that existed in the bucket **before** default encryption was enabled.

#### Useful CloudFormation snippets

Proactively block attempts to make a bucket public, or to specify a public ACL for objects:

```yaml
    Type: AWS::S3::Bucket
    Properties:
      PublicAccessBlockConfiguration:
        BlockPublicAcls: true
        BlockPublicPolicy: true
        IgnorePublicAcls: true
        RestrictPublicBuckets: true
```

Describe the default server-side encryption to apply to new objects in the bucket:

```yaml
    Type: AWS::S3::Bucket
    Properties:
      BucketEncryption:
      BucketEncryption:
        ServerSideEncryptionConfiguration:
        - ServerSideEncryptionByDefault:
            SSEAlgorithm: AES256 | aws:kms   # chose one
            KMSMasterKeyID:  <ARN for key>   # allowed if SSEAlgorithm is aws:kms
```

### DIFF
--- a/hq/app/logic/PublicBucketsDisplay.scala
+++ b/hq/app/logic/PublicBucketsDisplay.scala
@@ -28,12 +28,7 @@ object PublicBucketsDisplay {
     s"https://console.aws.amazon.com/s3/home?bucket=${URLEncoder.encode(bucket.bucketName, "utf-8")}"
   }
 
-  def isOnlyMissingEncryption(bucket: BucketDetail): Boolean = {
-    if (bucket.reportStatus.getOrElse(None) == Blue) {
-      return !bucket.isEncrypted
-    }
-    false
-  }
+  def isOnlyMissingEncryption(bucket: BucketDetail): Boolean = bucket.reportStatus.getOrElse(None) == Blue && !bucket.isEncrypted
 
   private[logic] def bucketReportStatus(bucket: BucketDetail): ReportStatus = {
     // permission properties that grant global access
--- a/hq/app/logic/PublicBucketsDisplay.scala
+++ b/hq/app/logic/PublicBucketsDisplay.scala
@@ -28,6 +28,13 @@ object PublicBucketsDisplay {
     s"https://console.aws.amazon.com/s3/home?bucket=${URLEncoder.encode(bucket.bucketName, "utf-8")}"
   }
 
+  def isOnlyMissingEncryption(bucket: BucketDetail): Boolean = {
+    if (bucket.reportStatus.getOrElse(None) == Blue) {
+      return !bucket.isEncrypted
+    }
+    false
+  }
+
   private[logic] def bucketReportStatus(bucket: BucketDetail): ReportStatus = {
     // permission properties that grant global access
     // The bucket ACL allows Upload/Delete access to anyone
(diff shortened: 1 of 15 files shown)


hq/app/logic/PublicBucketsDisplay.scala
```diff
@@ -27,6 +28,13 @@ object PublicBucketsDisplay {
     s"https://console.aws.amazon.com/s3/home?bucket=${URLEncoder.encode(bucket.bucketName, "utf-8")}"
   }
 
+  def isOnlyMissingEncryption(bucket: BucketDetail): Boolean = {
+    if (bucket.reportStatus.getOrElse(None) == Blue) {
```
philmcmahon: This looks like a comparison between an option and an int to me?

philmcmahon: I guess this function could be simplified to just return `bucket.reportStatus.getOrElse(None) == Blue && !bucket.isEncrypted`

philmcmahon:
Also, I think this is the best PR description I've ever seen! 

gold star

PR #154, opened 2020-02-21
Play 2.8 upgrade, bump dependencies and fix use of deprecated Logger

katebee:
#### 👉 Upgrade to Play 2.7.4

Minor upgrade on version from `2.6.7` to `2.7.4` 

App runs locally okay and only issue seems to have been the logging.

#### 👉 Fix use of deprecated Logger

Switching to Play 2.7 causes issues with how logging was done previously, so this PR also changes up the logging.

## What is the value of this?

Updated Play  ⭐️ 🚀 ✨


hq/app/aws/AwsAsyncHandler.scala
```diff
@@ -11,7 +11,7 @@ import scala.concurrent.{ExecutionContext, Future, Promise}
 class AwsAsyncPromiseHandler[R <: AmazonWebServiceRequest, T](promise: Promise[T], clientContext: AwsClient[_]) extends AsyncHandler[R, T] {
   def onError(e: Exception): Unit = {
     val context = Failure.contextString(clientContext)
-    Logger.warn(s"Failed to execute AWS SDK operation (${clientContext.client.getClass.getSimpleName}), $context", e)
+    Logger("aws").warn(s"Failed to execute AWS SDK operation (${clientContext.client.getClass.getSimpleName}), $context", e)
```
adamnfish: Arguably this is an AWSPromiseHandler logger, rather than an aws one? I think using the name would help when someone's reading the logs later? This would be handled for us if we use the mixin trait (see my main comment as well)

hq/app/utils/attempt/PlayIntegration.scala
```diff
@@ -10,7 +10,7 @@ object PlayIntegration extends Results {
   def attempt[A](action: => Attempt[Result])(implicit ec: ExecutionContext, assetsFinder: AssetsFinder): Future[Result] = {
     action.fold(
       { failures =>
-        Logger.error(failures.logMessage, failures.firstException.orNull)
+        Logger("attempt").error(failures.logMessage, failures.firstException.orNull)
```
adamnfish: I think this would be a `PlayIntegration` error, rather than an `attempt` one (as with my other comments)

project/plugins.sbt
```diff
@@ -5,7 +5,7 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.6")
 libraryDependencies += "org.vafer" % "jdeb" % "1.3" artifacts Artifact("jdeb", "jar", "jar")
 
 // The Play plugin
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.6.7")
+addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.7.3")
```
adamnfish: I think 2.7.4 is out now as well? Arguably we could look at making the jump straight to 2.8.x?

build.sbt
```diff
@@ -6,14 +6,14 @@ import sbt.Keys.libraryDependencies
 // common settings (apply to all projects)
 organization in ThisBuild := "com.gu"
 version in ThisBuild := "0.2.0"
-scalaVersion in ThisBuild := "2.12.3"
+scalaVersion in ThisBuild := "2.12.8"
```
adamnfish: I think Scala 2.12.10 is out now? Arguably we could look at bumping to the 2.13.x series as well, although that'll depend on the libraries we depend on being ready.

build.sbt
```diff
@@ -6,14 +6,14 @@ import sbt.Keys.libraryDependencies
...
 scalacOptions in ThisBuild ++= Seq("-deprecation", "-feature", "-unchecked", "-target:jvm-1.8", "-Xfatal-warnings")
 
 // resolvers += "guardian-bintray" at "https://dl.bintray.com/guardian/sbt-plugins/"
 resolvers += DefaultMavenRepository
 
 val awsSdkVersion = "1.11.596"
-val playVersion = "2.6.7"
+val playVersion = "2.7.4"
```
adamnfish: This is used to pull in Play Json, but looks like this is out of sync with the version we're using for Play itself (see my other comment in the plugin file)

katebee:
Great, thanks for the review @adamnfish ! 
I will take another look at this next week and tackle the comments 😄

build.sbt
katebee: Have bumped to `2.12.10` might see if we can get to `2.13.x` and update the AWS SDK versions in a separate PR 🚀

hq/app/aws/AwsAsyncHandler.scala
katebee: Have followed suggestion in main comment to use mixin trait instead 😸

katebee:
> One thing to double check with a big Play update is that the assets still load the way you expect (especially if they are versioned)

Thanks that is really good to know for this and for other updates 👍 🥇
Have checked and versioned assets still loading.

PR #161, opened 2020-07-20
Create DNS validated certificate within the cloudformation template.

akash1810:
## What does this change?
With https://github.com/guardian/dns-validation-lambda it is now possible to create DNS validated certificates that require no human interaction to be verified.

DNS validate certificates are preferred over email validated ones as they auto-renew; email validated certificates require a manual step of clicking a link in an email.

This changes the template to accept the domain name as a parameter, which is used to create the certificate.

## What is the value of this?
DNS validate certificates are preferred over email validated ones as they auto-renew; email validated certificates require a manual step of clicking a link in an email.

CloudFormation changes. Note: creating certificates in CFN is a blocking request.

## Any additional notes?
Added the domain as a parameter to keep the template as generic as possible for external collaborators.

### DIFF
--- a/cloudformation/security-hq.template.yaml
+++ b/cloudformation/security-hq.template.yaml
@@ -19,7 +19,7 @@ Metadata:
       Parameters:
       - VpcId
       - Subnets
-      - TLSCertArn
+      - TLSCertDomain
 
 Parameters:
   SecurityHQSourceBundleBucket:
@@ -34,8 +34,8 @@ Parameters:
   AMI:
     Description: Base AMI for Security HQ instances
     Type: AWS::EC2::Image::Id
-  TLSCertArn:
-    Description: ARN of a TLS certificate to install on the load balancer
+  TLSCertDomain:
+    Description: Domain name for a TLS certificate to install on the load balancer
     Type: String
   AccessRestrictionCidr:
     Description: CIDR block from which access to Security HQ should be allowed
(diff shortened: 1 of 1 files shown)


cloudformation/security-hq.template.yaml
```diff
@@ -331,7 +331,20 @@ Resources:
         Value:
           Fn::FindInMap: [ Constants, App, Value ]
         PropagateAtLaunch: true
-
+  TLSCertArn:
```
katebee, line 334: I am uncertain if having the certificate created as part of the CloudFormation is 100% desirable? This presumes the existence of the DNS validation lambda and that the string passed in the `TLSCertDomain` parameter will be one that the lambda will be able to validate.

If this certificate fails to validate, what happens to the stack? Does it remain in create/update status? OR will it eventually timeout?

akash1810: > If this certificate fails to validate, what happens to the stack? Does it remain in create/update status? OR will it eventually timeout?

ACM in CloudFormation is blocking; the stack will be in create/update whilst the certificate is not validated. IIRC a certificate request times out after 72 hours and will cause the stack to fail.

I'm uncertain if we would ever change the domain of out instance of Security HQ. That is, our estate is setup to be happy with this change.

It sounds like we'd like to keep this repository approachable for external parties? If so, I think I agree, this change isn't needed and we should create the domain validated certificate manually and just update this parameter in the stack.

akash1810:
Closing this as we're creating the certificate outside of this stack to remove a dependency on the dns validation lambdas.

PR #162, opened 2020-07-21
log shipping

akash1810:
## What does this change?
Adds log shipping in two different flavours:
- Remote shipping to the central-ELK stack via Kinesis and the [`kinesis-logback-appender`](https://github.com/guardian/kinesis-logback-appender).
- Local log shipping to a [`local-elk`](https://github.com/guardian/local-elk) stack. This is just to provide a simpler interface for log interrogation compared to `grep` and others.

## What is the value of this?
Shipping logs to an ELK stack provides a simpler interface for interrogation.

Shipping logs to a remote ELK stack will mean one no longer needs to ssh onto an instance to observe logs generated by the application.

No.

## Any additional notes?
This is heavily inspired by the Editorial Tools, for example [Workflow Frontend](https://github.com/guardian/workflow-frontend/blob/ca3789371031d52caaced42103740fbfd57c96b1/common-lib/src/main/scala/com/gu/workflow/lib/LogConfig.scala) and [Grid](https://github.com/guardian/grid/blob/a42d4d1acf1de7e4f06af003ddfe39dc25105371/common-lib/src/main/scala/com/gu/mediaservice/lib/logging/LogConfig.scala).

This uses the [latest 1.X.X `kinesis-logback-appender` available](https://mvnrepository.com/artifact/com.gu/kinesis-logback-appender). Version 2.X.X has a different API that I couldn't get to compile, maybe we bump a little later?

## Screenshots
**local logs**
image

TODO:
- [x] test it!
- [x] create a start script with a flag to abstract away the env var, for example [this in Workflow Frontend](https://github.com/guardian/workflow-frontend/blob/ca3789371031d52caaced42103740fbfd57c96b1/scripts/start.sh)

### DIFF
--- a/hq/app/config/Config.scala
+++ b/hq/app/config/Config.scala
@@ -2,6 +2,7 @@ package config
 
 import java.io.FileInputStream
 
+import com.amazonaws.regions.Regions
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
 import com.gu.googleauth.{AntiForgeryChecker, GoogleAuthConfig, GoogleGroupChecker, GoogleServiceAccount}
 import model.{AwsAccount, DEV, PROD, Stage}
@@ -13,6 +14,8 @@ import scala.util.Try
 
 
 object Config {
+  val region: Regions = Regions.EU_WEST_1
+
   def getStage(config: Configuration): Stage = {
     config.getAndValidate("stage", Set("DEV", "PROD")) match {
       case "DEV" => DEV
--- a/hq/app/config/Config.scala
+++ b/hq/app/config/Config.scala
@@ -14,6 +14,7 @@ import scala.util.Try
 
 
 object Config {
+  // TODO fetch the region dynamically from the instance
   val region: Regions = Regions.EU_WEST_1
 
   def getStage(config: Configuration): Stage = {
--- a/hq/app/config/LoggingConfig.scala
+++ b/hq/app/config/LoggingConfig.scala
@@ -1,0 +1,38 @@
+package config
+
+import com.amazonaws.regions.Regions
+import com.amazonaws.util.EC2MetadataUtils
+import model.{DEV, Stage}
+import play.api.Configuration
+
+case class LoggingConfig (
+  stack: String,
+  stage: Stage,
+  app: String,
+  instanceId: String,
+  region: Regions,
+  streamName: Option[String],
+  stsRole: Option[String]
+) {
+  val isDev: Boolean = stage == DEV
+}
+
+object LoggingConfig {
+  def apply(configuration: Configuration): LoggingConfig = {
+    val stack: String = configuration.get[String]("stack")
+    val app: String = configuration.getOptional[String]("app").getOrElse("security-hq")
+    val instanceId: String = Option(EC2MetadataUtils.getInstanceId).getOrElse("unknown")
+    val loggingStreamName: Option[String] = configuration.getOptional[String]("aws.kinesis.logging.streamName")
+    val loggingRole: Option[String] = configuration.getOptional[String]("aws.kinesis.logging.stsRoleToAssume")
+
+    LoggingConfig(
+      stack,
+      Config.getStage(configuration),
+      app,
+      instanceId,
+      Config.region,
+      loggingStreamName,
+      loggingRole
+    )
+  }
+}
--- a/hq/app/config/LoggingConfig.scala
+++ b/hq/app/config/LoggingConfig.scala
@@ -1,10 +1,8 @@
 package config
 
-import java.net.InetSocketAddress
-
 import com.amazonaws.regions.Regions
 import com.amazonaws.util.EC2MetadataUtils
-import model.{DEV, Stage}
+import model.Stage
 import play.api.Configuration
 
 case class LoggingConfig (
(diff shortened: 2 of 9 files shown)


philmcmahon:
Code LGTM 👍

hq/app/config/Config.scala
```diff
@@ -13,6 +14,8 @@ import scala.util.Try
 
 
 object Config {
+  val region: Regions = Regions.EU_WEST_1
```
adamnfish, line 17: So this change is an interesting one!

It was originally hard coded in the AppComponents, which is the "edge" of the program that starts our webserver. In truth, this shouldn't be hard-coded at all, we should be reading it from the AWS metadata somehow. Moving it here is a good choice while it is static, but is a step backwards in terms of us being able to un-hard-code it in future. As I mention elsewhere in this review, if we're doing a lookup we'll want to keep the IO "at the edge" rather than have the instantiation of our Config object at load time cause trouble when it starts running off to the network to fill in the vals.

I don't feel strongly one way or the other with this change here - it is hard coded now so this looks great. I just wanted to share the thinking about why it was where it was :-)

hq/app/logging/LogConfig.scala
```diff
@@ -0,0 +1,113 @@
...
+  private def buildCredentialsProvider(stsRole: String) = {
+    val random = new SecureRandom()
+    val sessionId = s"session${random.nextDouble()}"
+
+    val instanceProvider = InstanceProfileCredentialsProvider.getInstance
+    val stsClient = AWSSecurityTokenServiceClientBuilder.standard.withCredentials(instanceProvider).build
+    new STSAssumeRoleSessionCredentialsProvider.Builder(stsRole, sessionId).withStsClient(stsClient).build
+  }
+
+  def initLocalLogShipping(config: LoggingConfig): Unit = {
```
adamnfish, line 43: Is this ever called?

hq/app/config/LoggingConfig.scala
```diff
@@ -0,0 +1,42 @@
...
+
+case class LoggingConfig (
+  stack: String,
+  stage: Stage,
+  app: String,
+  instanceId: String,
+  region: Regions,
+  streamName: Option[String],
+  stsRole: Option[String]
+) {
```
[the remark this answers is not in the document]
akash1810: Interesting view point. The intention here is to create a logical grouping, `localLogShippingEnabled` doesn't make sense in any context other than logging configuration. Maybe a better pattern would be for these to be properties of the case class and then instantiated at the edge, rather than within the case class.

hq/app/config/LoggingConfig.scala
```diff
@@ -0,0 +1,42 @@
...
+  val isDev: Boolean = stage == DEV
```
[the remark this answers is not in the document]
akash1810: This is to maintain DRYness and should arguably be in `Config`.

hq/app/config/LoggingConfig.scala
```diff
@@ -0,0 +1,42 @@
...
+  val localLogShippingEnabled: Boolean = sys.env.getOrElse("LOCAL_LOG_SHIPPING", "false").toBoolean
+  lazy val localLogShippingDestination: InetSocketAddress = new InetSocketAddress("localhost", 5000)
```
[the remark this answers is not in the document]
akash1810: This is deserving of a comment; `local-elk` accepts TCP logs on `localhost:5000` only ([link](https://github.com/guardian/local-elk#tcp)).

hq/app/logging/LogConfig.scala
akash1810: Oops, it should be!

hq/app/logging/LogConfig.scala
```diff
@@ -31,24 +32,25 @@ object LogConfig {
...
+    val sessionId = s"${config.app}-session"
 
     val instanceProvider = InstanceProfileCredentialsProvider.getInstance
     val stsClient = AWSSecurityTokenServiceClientBuilder.standard.withCredentials(instanceProvider).build
     new STSAssumeRoleSessionCredentialsProvider.Builder(stsRole, sessionId).withStsClient(stsClient).build
   }
 
   def initLocalLogShipping(config: LoggingConfig): Unit = {
-    if(config.isDev && config.localLogShippingEnabled) {
+    if(config.stage == DEV && config.localLogShippingEnabled) {
```
adamnfish, line 44: This is a super minor point, but would it be helpful to log something in an `else` statement for this `if`? The remote logging says words to the effect of "I am disabled in DEV", in DEV. Should we be logging "this is only enabled in dev" and/or "local log shipping is disabled" to help people that come to this later?

adamnfish (approved):
Love these!

As before, the main thing is that it works, which you have already checked I think! Thanks heaps for contributing this

🙇

akash1810:
Reported Snyk issue should be resolved in https://github.com/guardian/play-googleauth/pull/77.

johnduffell:
can we call this PR "send logs directly to elk stack" or something as it wasn't clear what it was when I saw it linked from play-googleauth

Anyway, I just dropped by to say if you have trouble with guava/Function types with WebDriver, you need to explicitly pull in the "jre" version of guava, as somehow google-api-client doesn't depend on a specific version and sbt picks up the android one and evicts all the rest.  @Jpsi  and I separately spent a while diagnosing that one on support-frontend and membership-frontend.

akash1810:
Updated config in S3. Will merge this and address the Snyk issue in future PRs.

PR #163, opened 2020-08-28
Bump play-googleauth to fix Snyk warning

amyhughes:
This should address the [Snyk warning introduced through com.google.oauth-client:google-oauth-client](https://snyk.io/vuln/SNYK-JAVA-COMGOOGLEOAUTHCLIENT-575276), which is a transitive dependancy of play-googleauth 0.7.6.

A later version of google-oauth-client was brought in to play-googleauth in [this PR](https://github.com/guardian/play-googleauth/pull/77), and [this PR](https://github.com/guardian/play-googleauth/pull/79) built on this to support Play 2.8. We can now bring in the latest version of play-googleauth to address the Snyk warning.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ import sbt.Keys.libraryDependencies
 organization in ThisBuild := "com.gu"
 version in ThisBuild := "0.2.0"
 scalaVersion in ThisBuild := "2.12.10"
-scalacOptions in ThisBuild ++= Seq("-deprecation", "-feature", "-unchecked", "-target:jvm-1.8", "-Xfatal-warnings")
+scalacOptions in ThisBuild ++= Seq("-deprecation", "-feature", "-unchecked", "-target:jvm-1.8")
 
 // resolvers += "guardian-bintray" at "https://dl.bintray.com/guardian/sbt-plugins/"
 resolvers += DefaultMavenRepository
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -7,7 +7,7 @@ import sbt.Keys.libraryDependencies
 organization in ThisBuild := "com.gu"
 version in ThisBuild := "0.2.0"
 scalaVersion in ThisBuild := "2.12.10"
-scalacOptions in ThisBuild ++= Seq("-deprecation", "-feature", "-unchecked", "-target:jvm-1.8", "-Xfatal-warnings")
+scalacOptions in ThisBuild ++= Seq("-deprecation", "-feature", "-unchecked", "-target:jvm-1.8")
```
katebee, line 10: Why are we removing `"-Xfatal-warnings"` ?

akash1810: Good question! It's complicated!

`play-googleauth` has [deprecated](https://github.com/guardian/play-googleauth/blob/edcd52cf5dc541b7b01772c670e96bd82088096d/play-v27/src/main/scala/com/gu/googleauth/auth.scala#L144) `AntiForgeryChecker.borrowSettingsFromPlay`, which we're using, in favour of [`play-secret-rotation`](https://github.com/guardian/play-secret-rotation). In order to add `play-secret-rotation` we'd also need to add parameter store support to this project, making this PR quite involved!

An alternative would be to `@nowarn` annotation [as Janus did](https://github.com/guardian/janus-app/pull/12/files), however that's a Scala 2.13 feature. Security HQ is on Scala 2.12.10, bumping it to 2.13.3 is non-trivial as not all of the dependencies have a Scala 2.13 release.

The cheapest solution was to remove the `-Xfatal-warnings` compiler option.

adamnfish: I'd definitely prefer not to see this flag removed from SHQ. If this PR is important as-is then I think we should consider restoring the flag as debt that needs addressing in a follow-up.

Alternatively, we might re-consider whether deprecating that method in play-googleauth is appropriate. It seems like all our projects are choosing to use it, in which case the friction that the deprecation creates might be outweighing the benefit of encouraging teams to adopt rotation.

adamnfish (approved):
I've pitched in on the chat about the deprecation warning, but I don't think that necessarily needs to block this PR. Looks good otherwise, thank you!

jfsoul:
Closing this PR as it has gotten quite stale. We can always re-open once this vulnerability/dependency is reviewed again

PR #165, opened 2020-12-01
Package as debian

jamie-lynch:
## What does this change?

This PR updates the build to package as a debian. It also changes some other config as part of this:

* Copy the configuration file to `/etc/gu` and import in the `application.conf` instead of the other way around.
* Include java options in `build.sbt` rather than a start script.

As well as this, it bumps the instance to be `t3.large` as recent work has identified that the instances are somewhat under-powered.

## What is the value of this?

This change brings the deployment more in line with other projects. It will also allow us to continue investigating recent issues. 

Yes. Both the user data and instance type have been changed. The changes will be applied before this PR is merged.

It does not require changes to the StackSet.


akash1810:
I always get confused about this: why is debian preferred?

jamie-lynch:
> I always get confused about this: why is debian preferred?

That's a good question which I don't really know the answer to. Maybe one for @sihil?

philmcmahon:
I think it's because the debian package includes all the systemd config so you don't have to provide it separateely. you can also install it in a single line rather than having to unzip and then do something else

katebee (approved):
👍

PR #167, opened 2020-12-11
Remove SSH access, add egress for wazuh agents

katebee:
#### 👉 Remove SSH access to instances

For a long time at The Guardian it has been possible to SSH without opening up port 22, thanks to [SSM-Scala!](https://github.com/guardian/ssm-scala)

Along with removing the requirement to have port 22 open, SSM-Scala also provides visibility into which instances have been accessed and removes the requirement for bastions (thanks to ssm-tunnelling!) The latest version of SSM-Scala - 2.0.0 - sets SSM tunnelling as the default behaviour.

SSM-Scala should therefore be the default way we access instances.

#### 👉  Enable Wazuh agent communication 

This changes the security groups of several instances so that the Wazuh agents can register with the Wazuh manager and send events to the Wazuh cluster.

- Port 1514 is used for agent event logging
- Port 1515 is used for agent registration

#### 👉  Add permissions to describe region and ASG

Describing region, tags and ASG is required so that the instance can configure labels during Wazuh agent installation. The labels can provide additional information about the instance the agent is running on, such as  the stack, app and stage.

Yes, the Cloudformation template will be applied after these changes are approved

TODO:
- [ ] Apply Cloudformation changes

### DIFF
--- a/cloudformation/security-hq.template.yaml
+++ b/cloudformation/security-hq.template.yaml
@@ -129,6 +129,9 @@ Resources:
         - Effect: Allow
           Resource: "*"
           Action:
+          - autoscaling:DescribeAutoScalingInstances
+          - autoscaling:DescribeAutoScalingGroups
+          - ec2:DescribeRegions
           - ec2:DescribeTags
       Roles:
       - !Ref SecurityHQInstanceRole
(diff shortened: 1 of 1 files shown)


jamie-lynch (approved):
🔒

PR #169, opened 2021-01-04
Update node and packages, switch to stylelint

katebee:
#### 👉 Update node and package versions

This will give us the (almost) latest and greatest package features and also fix some security warnings.

The current stable version of node is 14 so this has also been updated in the nvmrc.

#### 👉  Remove sasslint and replace it with stylelint

sass-lint has been unmaintained for a while and is the source of several warnings when using `yarn audit`. 
After looking around at alternatives stylelint appeared to a good drop-in replacement.

## Any additional notes?

We might be able to switch to using stylelint for the JS too therefore I have included the scripts required.
Linting has never been formalised in SHQ, but having some linting options available keeps me happy. 🧙‍♀️ ✨


Nirvikalpa108 (approved):
This looks good to me @katebee 👍🏽
I had a look through the failing build errors caused by snyk vulnerabilities and it looks like we can upgrade to AWS SDK 2 and upgrade our google-auth client to fix them. I would be happy to pair on this!💕

PR #170, opened 2021-01-11
Move readme

Nirvikalpa108:
## What does this change?

Here we move the README to the root of the project, so that it shows on https://github.com/guardian/security-hq. 

## What is the value of this?

I think this is a good, because when people navigate to the project, they will immediately see the README which contains a lot of  useful information.


katebee (approved):
👍  and agree this is a good move.

Guessing this is a straight copy-paste from the deleted readme. Some of the information is outdated/incorrect (e.g. reference to EBS) but that is outside the scope of the PR. I am mentioning here for record and will hope we can improve the README this quarter 😄

PR #174, opened 2021-01-18
Update readme

Nirvikalpa108:
## What does this change?
We add some instructions to the README about adding AWS accounts for local development and the nginx process that is required for running the app locally. 

We also remove the deployment section (which I'm not 100% sure is the correct thing to do! @katebee I would love your advice here, thank you!💕💕💕)

## What is the value of this?
I'm hoping to incrementally improve the README with new things that I learn as I am working on the app. I hope this will improve the developer experience for developers who work on the app after me, as well as my future self!😂


README.md
```diff
@@ -64,11 +50,33 @@ The next file is `~/.gu/security-hq.local.conf`, which should be given by one of
...
+The value of `id` should be the same as the AWS Profile name, which you can see when you copy your credentials from janus. You can add a `roleArn` if you want to generate an iAM report, otherwise you don't need it.
+
 ### AWS Security Policies
 See `watched-account` template under `cloudformation` folder for the security policies needed to run security-hq.
 
 ### nginx setup
 
+When running security-hq locally you will always need nginx running in the background. You only need to configure dev-nginx once (as described in step 2), but if you restart your computer, you will need to run `sudo nginx` to restart the nginx process.
+
+To check nginx is running, you can run `$ ps -e | grep nginx` and you should receive a response that includes `nginx: master process nginx` and `nginx: worker process`.
```
katebee, line 78: Might be worth removing the `$` to avoid confusion? The `$` is a stand-in for the end of the prompt and should not be entered before the `ps`. I realise the `$` is sometimes included elsewhere in the README so you might have been keeping with the existing style. 😊 

```suggestion
To check nginx is running, you can run `ps -e | grep nginx` and you should receive a response that includes `nginx: master process nginx` and `nginx: worker process`.
```

README.md
```diff
@@ -64,11 +50,33 @@ The next file is `~/.gu/security-hq.local.conf`, which should be given by one of
 
 Another file you need will need (here again ask for a team member) is `~/.gu/security-hq-service-account-cert.json`.
 
+### Adding additional AWS accounts for local development
+
+As well as the `security-test` and `security` accounts, you can add additional AWS accounts to your local security-hq. This may be helpful for debugging purposes for example. You will need janus credentials for the account you want to add.
```
katebee, line 55: (minor) If we are mentioning Janus, we should capitalize the J. Although, another option is to avoid mentioning Janus entirely. Some reports require IAM user credentials (which is why we create a security-test user) rather than Janus credentials.

```suggestion
When running security HQ locally, you can modify the list of AWS accounts to include additional account. For example, you may want to add a specific account for debugging purposes. You will need valid AWS credentials for any accounts you wish to include.
```

README.md
```diff
@@ -64,11 +50,33 @@ The next file is `~/.gu/security-hq.local.conf`, which should be given by one of
...
+AWS_ACCOUNTS = [
+  {
+   name = "Deploy Tools"
+   id = "deployTools" 
+   roleArn = ""
+  }
+]
+```
+
+The value of `id` should be the same as the AWS Profile name, which you can see when you copy your credentials from janus. You can add a `roleArn` if you want to generate an iAM report, otherwise you don't need it.
```
katebee, line 69: Couple of capitalisation fixes 😸 

```suggestion
The value of `id` should be the same as the AWS Profile name, which you can see when you copy your credentials from Janus. You can add a `roleArn` if you want to generate an IAM report, otherwise you don't need it.
```

katebee (approved):
👍  I have added a few minor comments that you can selectively address if you want 😊
Thanks for the README improvements!! 🌈 🌟

Issue #168, opened 2020-12-24
Indicate the age of each access key in the IAM report


katebee:
AWS [recommends](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html?icmpid=docs_iam_console) rotating access keys frequently.

Displaying the key age in the IAM report would help identify keys that are overdue for rotation.

I have experimented with added the access key age to the credentials report in this branch: https://github.com/guardian/security-hq/compare/kw-20201224-key-age

This still needs some decisions around UX and design, then the twirl template needs to be updated to include the ages

PR #175, opened 2021-01-21
Add access key age to the credential report

katebee:
#### 👉 Include access key rotation dates in IAM report

Switch from `KeyStatus` to a new case class `AccessKey`, which includes both the `KeyStatus` and the last rotation date.

#### 👉 Add access key age to the displayed credential report

Along with the key icon, the last rotation date is used to calculate and display the age of the key. 
The text will be black if the key is enabled and grey if the key is disabled:

Screenshot 2021-01-21 at 11 49 35

#### 👉 Delete the security-group-testing template

Small bonus clean up, as this template is no longer useful or applied anywhere. 
Removing it from the repository to declutter!! 🧙‍♀️ 🧹 ✨ 

## What is the value of this?

Closes: https://github.com/guardian/security-hq/issues/168 🚀 

As a member of the InfoSec team, I want a single place where I can review the age of all permanent AWS access keys. Displaying the key age in the IAM report would help identify keys that are overdue for rotation.

As an administration of an AWS account, I want to identify any keys over a certain age in my account so that they can be rotated as [recommended by AWS](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html?icmpid=docs_iam_console).

Now that we have the last rotation date of the keys included, subsequent PRs could flag accounts with keys over a certain age, include the key ages in the ranking of accounts, or provide downloadable reports that can be shared with stakeholders.

## Any additional notes?

Boris Johnson standing behind a podium emblazoned with the words review, reduce, rotate


Nirvikalpa108 (approved):
Thank you for this great feature @katebee 🎉🎉🎉  

This makes it so much easier for us to audit permanent credentials across the Guardian now that we have the age of the access key!

Additional features that would build on this work, could include:
- creating a new status that takes into account the age of the access key, so that it's easier to see in the UI the highest priority actions for each key holder.
- It would be great if this information could be downloaded into a CSV file, so that we can show this to stakeholders and easily track our progress over time.
- Finally, as a stretch goal, it would be great if we could send users alerts when their key is a certain number of days old and then perhaps delete it automatically after a certain period of time.

Amazing work @katebee 👍🏽 😃 💯👏🏽

PR #176, opened 2021-02-03
Add tooltips for menu items

philmcmahon:
## What does this change?
Adds some tooltips to the menu to identify what each icon does.

## What is the value of this?
Usability. A better option would arguably be labels but they probably wouldn't fit, and this is a super quick and easy option.

No

Screenshot 2021-02-03 at 15 47 15


katebee (approved):
👍  good idea

Issue #172, opened 2021-01-12
Consider removing AWS Inspector


katebee:
The inspector results on the dashboard are no longer updating. This may indicate that the inspector is no longer running, or be due to another issue. [AWS inspector](https://aws.amazon.com/inspector/) is a paid for service, so if we are not obtaining actionable information we should remove it.

Additionally, the InfoSec program has led to the adoption of [Wazuh](https://wazuh.com/) as our security monitoring solution for threat detection, integrity monitoring, incident response and compliance. 

There may be significant overlap between what is reported via a [Wazuh agent](https://documentation.wazuh.com/3.13/amazon/instances.html) and covered by AWS Inspector therefore we may want to focus on rollout of the Wazuh agents and hardening systems unless we are clear on the additional value that AWS Inspector would provide.

 Wazuh agents can now be installed and configured via [the AMIgo role](https://amigo.gutools.co.uk/roles#wazuh-agent) and provide:
- Log data collection
- File integrity monitoring
- Anomaly and malware detection
- Security policy monitoring
- System inventory
- Vulnerability detection

@aug24 and @adamnfish, it has been a while since we worked on this but I would value your input! 😁

PR #178, opened 2021-02-04
Remove AWS Inspector results from Security HQ

katebee:
## What does this change?

Remove the AWS Inspector service, cache documentation and views.

## What is the value of this?

Closes https://github.com/guardian/security-hq/issues/172

The InfoSec program has led to the adoption of [Wazuh](https://wazuh.com/) as our security monitoring solution for threat detection, integrity monitoring, incident response and compliance. 

There may be significant overlap between what is reported via a [Wazuh agent](https://documentation.wazuh.com/3.13/amazon/instances.html) and covered by AWS Inspector therefore we may want to focus on rollout of the Wazuh agents and hardening systems unless we are clear on the additional value that AWS Inspector would provide.

Additionally, AWS Inspector results on the dashboard are no longer updating. This may indicate that the inspector is no longer running, or be due to another issue. [AWS inspector](https://aws.amazon.com/inspector/) is a paid for service, so if we are not obtaining actionable information we should remove it.

There are a couple of read permissions on the watched account template I have decided to leave these for now to avoid needing to update the StackSet.

There are a few fixes and improvements that can be made to the template which might be a better time to review all the permissions Security HQ has and needs.

## Any additional notes?

Tests pass locally and I can still run the application locally.

## TODO:

- [ ] Find any Inspector Lambdas that are running and decommission them


Nirvikalpa108 (approved):
Amazing, thank you so much for this @katebee 🎉  As well as the business benefits of removing this and no longer paying for AWS Inspector, there will also be benefits of a lighter weight Security HQ when developers are running locally. Great!💯

PR #179, opened 2021-02-05
Adds Elastic File System (EFS)

Nirvikalpa108:
## What is the problem?

Security HQ does not recognise AWS Elastic File Systems. This means that when an EFS security group is flagged by Trusted Advisor (because for example it has ports open to the world and is therefore a security risk), the `/security-groups` endpoint does not give us the file system id and corresponding link to it in the AWS console. 

## What does this change?

We've now added the capability for Security HQ to recognise EFSs.

| Before      | After* |
| ----------- | ----------- |
| Screenshot 2021-02-05 at 13 24 00 | Screenshot 2021-02-05 at 13 23 52 |
_*This test File System is now deleted_

## What is the value of this?

This change means that when a Guardian account has an EFS with security groups flagged by Trusted Advisor, we will see the file system id at https://security-hq.gutools.co.uk/security-groups and, if we have credentials, can click through to it in the AWS console. 

This is helpful, because it means we can more easily remediate EFS security problems with relevant teams☺️.

### DIFF
--- a/hq/app/aws/ec2/EFS.scala
+++ b/hq/app/aws/ec2/EFS.scala
@@ -53,12 +53,11 @@ object EFS {
   }
 
   // these functions remove the UnknownUsage EFS values, because we already have the correct EfsVolume value
-  def removeAllEfsUnknownUsages(allSecGrps: Attempt[Map[String, Set[SGInUse]]])(implicit ec: ExecutionContext): Attempt[Map[String, Set[SGInUse]]] = allSecGrps.map(filterUnknownEfsSecGrps)
-  def filterUnknownEfsSecGrps(secGrpToResources: Map[String, Set[SGInUse]]): Map[String, Set[SGInUse]] = secGrpToResources.mapValues(filterResources)
-  def filterResources(resources: Set[SGInUse]): Set[SGInUse] = resources.filter(filterResource)
-  def filterResource(resource: SGInUse): Boolean = resource match {
-    case resource:UnknownUsage => if (resource.description.contains("EFS mount target")) false else true
-    case _ => true
+  def filterOutEfsUnknownUsages(secGrpToResources: Map[String, Set[SGInUse]]): Map[String, Set[SGInUse]] = secGrpToResources.mapValues(filterOutEfsUnknownUsagesFromSet)
+  def filterOutEfsUnknownUsagesFromSet(resources: Set[SGInUse]): Set[SGInUse] = resources.filterNot(efsUnknownUsages)
+  def efsUnknownUsages(resource: SGInUse): Boolean = resource match {
+    case resource:UnknownUsage => resource.description.contains("EFS mount target")
+    case _ => false
   }
 }
 
--- a/hq/app/aws/ec2/EFS.scala
+++ b/hq/app/aws/ec2/EFS.scala
@@ -1,0 +1,64 @@
+package aws.ec2
+
+import aws.AwsAsyncHandler.{awsToScala, handleAWSErrs}
+import aws.AwsClient
+import com.amazonaws.services.elasticfilesystem.AmazonElasticFileSystemAsync
+import com.amazonaws.services.elasticfilesystem.model._
+import model.{SGInUse, UnknownUsage}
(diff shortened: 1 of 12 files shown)


hq/app/aws/ec2/EFS.scala
```diff
@@ -0,0 +1,64 @@
...
+      sgIds.contains(scGrp)
+    }
+  }
+
+  // these functions remove the UnknownUsage EFS values, because we already have the correct EfsVolume value
+  def removeAllEfsUnknownUsages(allSecGrps: Attempt[Map[String, Set[SGInUse]]])(implicit ec: ExecutionContext): Attempt[Map[String, Set[SGInUse]]] = allSecGrps.map(filterUnknownEfsSecGrps)
+  def filterUnknownEfsSecGrps(secGrpToResources: Map[String, Set[SGInUse]]): Map[String, Set[SGInUse]] = secGrpToResources.mapValues(filterResources)
+  def filterResources(resources: Set[SGInUse]): Set[SGInUse] = resources.filter(filterResource)
+  def filterResource(resource: SGInUse): Boolean = resource match {
+    case resource:UnknownUsage => if (resource.description.contains("EFS mount target")) false else true
```
philmcmahon, line 60:
```suggestion
    case resource:UnknownUsage => resource.description.contains("EFS mount target"))
```

hq/app/aws/ec2/EFS.scala
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  def getFlaggedSecGrps(sgIds: List[String], efsSecGrps: List[(String, SGInUse)]):List[(String, SGInUse)] = {
+    efsSecGrps.filter{ case (scGrp, _) =>
+      sgIds.contains(scGrp)
+    }
+  }
+
+  // these functions remove the UnknownUsage EFS values, because we already have the correct EfsVolume value
+  def removeAllEfsUnknownUsages(allSecGrps: Attempt[Map[String, Set[SGInUse]]])(implicit ec: ExecutionContext): Attempt[Map[String, Set[SGInUse]]] = allSecGrps.map(filterUnknownEfsSecGrps)
```
philmcmahon, line 56: I think it's better to try and only pass information to functions that they actually need - so in this case if the Attempt is a fail then we don't need to bother doing any of the filtering - so this function should just take the Map[String, Set[SgInUse]] as params. This also makes testing easier

hq/test/aws/ec2/EFSTest.scala
```diff
@@ -0,0 +1,35 @@
+package aws.ec2
```
philmcmahon, line 1: great work writing some tests!

hq/app/aws/ec2/EFS.scala
Nirvikalpa108: I really like this suggestion, thank you! ☺️ 

This works if I also use a `filterNot` in the function that calls this one and change the final case to `false`. It looks like this:
```
  def filterResources(resources: Set[SGInUse]): Set[SGInUse] = resources.filterNot(filterResource)
  def filterResource(resource: SGInUse): Boolean = resource match {
    case resource:UnknownUsage => resource.description.contains("EFS mount target")
    case _ => false
  }
```

Thank you for suggesting! ❤ ❤️ ❤️

hq/app/aws/ec2/EFS.scala
Nirvikalpa108: Thank you for this suggestion! 💯 👍🏽 😀

To implement this, I've changed the usage in `EC2.scala` so that we apply the function as part of the yield statement, like this:

```
    for {
      efsSg <- efsSgs
      ec2Sg <- ec2Sgs
    } yield filterOutEfsUnknownUsages(List(ec2Sg, efsSg).fold(Map.empty)(_ |+| _))
```

Thank you 💖  💖  💖

philmcmahon (approved):
👍🏻

Nirvikalpa108:
Before we can implement this change, we need to add efs permissions to Security HQ's stackset - https://github.com/guardian/security-hq/pull/181.

PR #198, opened 2021-03-09
Add GCP security vulnerabilities

Nirvikalpa108:
## What does this change?

This PR adds a new route, `/gcp`, to show Google Cloud Platform (GCP) security vulnerabilities for each project in the Guardian organisation.

Screenshot 2021-03-09 at 15 01 26

## What is the value of this?

We at the Guardian generally use AWS for our cloud infrastructure needs, but have recently increased our usage of GCP. Given this, it's vital that we include GCP assets in our security monitoring. 

This work is the Guardian's first major project to expose GCP security vulnerabilities so that engineering teams can easily see the remediation steps required for them to run their GCP infrastructure securely. 

## TODO

This PR should be seen as a minimum viable product, because there is still lots of work to do before this is ready to present to the department. Here are some of the major things on our TODO list after this PR is merged:

- add a /gcp link to the other routes, so that users can navigate to /gcp with ease!
- solve the missing severity values problem
- add a filter by severity
- colour code findings by severity
- change the green tick image next to the project names 
- add some logic to only show findings for one project when running in DEV mode (to make the developer experience nicer)
- more helpful error message when the GCP cache is not ready yet

### DIFF
--- a/hq/public/stylesheets/main.css
+++ b/hq/public/stylesheets/main.css
@@ -338,6 +338,30 @@ table.iam-report__table th {
   box-sizing: initial;
 }
 
+table.gcp-report__table td,
+table.gcp-report__table th {
+  min-height: 21px;
+  box-sizing: initial;
+  word-wrap: break-word;
+}
+
+table.gcp-report__table {
+  table-layout: fixed;
+  word-break: break-all;
+}
+
+.gcp-table-row-severity {
+  width: 55px;
+}
+
+.gcp-table-row-category {
+  width: 130px;
+}
+
+.gcp-table-row-date {
+  width: 65px;
+}
+
 .iam-modal__trigger {
   height: 24px;
 }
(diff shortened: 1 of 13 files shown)


hq/public/stylesheets/main.css
```diff
@@ -338,6 +338,30 @@ table.iam-report__table th {
...
 }
 
+table.gcp-report__table td,
+table.gcp-report__table th {
+  min-height: 21px;
+  box-sizing: initial;
+  word-wrap: break-word;
+}
+
+table.gcp-report__table {
```
katebee, line 348: Minor: I am not sure we need the `table` selector here. Security HQ currently tries to use [BEM](https://en.bem.info/methodology/quick-start/) (Block, Element, Modifier) to prevent the need for combined selectors.

You have a perfectly lovely BEM classname there, `gcp-report` is the block and `table` is the Element.

philmcmahon (approved):
🚀 
cloud

PR #201, opened 2021-03-16
Add GCP link to all pages

Nirvikalpa108:
## What does this change?
This PR adds the GCP logo and link to the homepage and to the headers of the other pages.
Screenshot 2021-03-16 at 10 02 08

## What is the value of this?
Users can now navigate to /gcp from other pages without manually typing it into the browser!


philmcmahon (approved):
☁️

PR #202, opened 2021-03-16
Post metrics to cloudwatch

jorgeazevedo:
## What does this change?

This makes Security HQ post the results of its data collection to cloudwatch as a metric. It's powering this [grafana dashboard](https://metrics.gutools.co.uk/d/RHAvkh8Gz/securityhq?orgId=1&from=now-2d&to=now). The data collected is based [on this original proposal](https://docs.google.com/document/d/1PIZC67-RymEYrtpOWLy18vAV8hRqjh67yBy4UezQoTQ/edit), but the key difference is in this PR I've only implemented 4 data types to reflect "totals". Namely

* Total number of S3 buckets `s3/total`
* Total number of IAM credentials marked red `iam/credentials/critical`
* Total number of IAM credentials marked amber `iam/credentials/warning`
* Total number of IAM credentials red or amber `iam/credentials/total`
* Total number of exposed keys `iam/keys/total`
* Total number of public security groups `securitygroup/total`

I'd be very keen on feedback for this, both on my dodgy scala code and the metrics themselves.

## What is the value of this?

To be able to answer questions like:

- Is the number of vulnerabilities trending up or down?
- Which teams have started fixing their issues?
- Is the number of public S3 buckets trending up or down?

I completely forgot about this, but yes. It probably requires a CFN change to allow the server to post metrics to cloudwatch. No change on the StackSet.

## Any additional notes?

- Still need to think about increasing retention
- The total vulnerabilities per account chart is kind of busted. Need to understand if this is an issue with cloudwatch metrics or just a general challenge of adding timeseries data, which might require that total to be calculate at the source.

@katebee @philmcmahon

### DIFF
--- a/hq/app/logging/Cloudwatch.scala
+++ b/hq/app/logging/Cloudwatch.scala
@@ -2,10 +2,11 @@ package logging
 
 import com.amazonaws.AmazonServiceException
 import com.amazonaws.services.cloudwatch.AmazonCloudWatchClientBuilder
-import com.amazonaws.services.cloudwatch.model.{Dimension, MetricDatum, PutMetricDataRequest, PutMetricDataResult, StandardUnit}
+import com.amazonaws.services.cloudwatch.model.{Dimension, MetricDatum, PutMetricDataRequest, StandardUnit}
+import logic.CredentialsReportDisplay
+import logic.CredentialsReportDisplay.reportStatusSummary
 import model.{AwsAccount, CredentialReportDisplay}
 import play.api.Logging
-
 import collection.JavaConverters._
 import scala.util.{Failure, Success, Try}
 
@@ -15,6 +16,8 @@ object Cloudwatch extends Logging {
   object DataType extends Enumeration {
     val s3Total = Value("s3/total")
     val iamCredentialsTotal = Value("iam/credentials/total")
+    val iamCredentialsCritical = Value("iam/credentials/critical")
+    val iamCredentialsWarning = Value("iam/credentials/warning")
     val iamKeysTotal = Value("iam/keys/total")
     val sgTotal = Value("securitygroup/total")
   }
@@ -22,7 +25,12 @@ object Cloudwatch extends Logging {
   def logAsMetric[T](data: Seq[T], dataType: DataType.Value ) : Unit = {
     data.foreach {
       case (account: AwsAccount, Right(details: CredentialReportDisplay)) =>
-        putMetric(account, dataType, details.humanUsers.length + details.machineUsers.length)
+
+        val reportSummary: CredentialsReportDisplay.ReportSummary = reportStatusSummary(details)
+        putMetric(account, DataType.iamCredentialsCritical, reportSummary.errors)
+        putMetric(account, DataType.iamCredentialsWarning, reportSummary.warnings)
+        putMetric(account, dataType, reportSummary.errors + reportSummary.warnings)
+
       case (account: AwsAccount, Right(details: List[Any])) =>
         putMetric(account, dataType, details.length)
       case (account: AwsAccount, Left(_)) =>
--- a/hq/app/logging/Cloudwatch.scala
+++ b/hq/app/logging/Cloudwatch.scala
@@ -33,7 +33,7 @@ object Cloudwatch extends logging {
     }
   }
 
-  def putMetric(account: AwsAccount, dataType: DataType.Value , value: Int): String = {
+  def putMetric(account: AwsAccount, dataType: DataType.Value , value: Int): Unit = {
     logger.info(s"METRIC:  Account=${account.name},DataType=${dataType},Value=${value}")
     val cw = AmazonCloudWatchClientBuilder.defaultClient
 
@@ -43,7 +43,6 @@ object Cloudwatch extends logging {
     )
     val datum = new MetricDatum().withMetricName("Vulnerabilities").withUnit(StandardUnit.Count).withValue(value.toDouble).withDimensions(dimension.asJava)
     val request = new PutMetricDataRequest().withNamespace("SecurityHQ").withMetricData(datum)
-    val response: PutMetricDataResult = cw.putMetricData(request)
-    response.toString
+    cw.putMetricData(request)
   }
 }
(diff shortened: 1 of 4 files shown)


hq/app/logging/Cloudwatch.scala
```diff
@@ -0,0 +1,48 @@
...
+    for ((account: AwsAccount, result) <- data) {
+      result match {
+        case Right(details: CredentialReportDisplay) => {
+          putMetric(account, dataType, details.humanUsers.length + details.machineUsers.length)
+        }
+        case Right(details: List[Any]) => {
+          putMetric(account, dataType, details.length)
+        }
+        case Left(_) => {
+          println(s"Attempt to log cloudwatch metric failed. Data of type ${dataType} is missing for account ${account.name}.")
```
philmcmahon, line 29: don't use println use `logger`  - there should be examples elsewhere

hq/app/logging/Cloudwatch.scala
```diff
@@ -0,0 +1,48 @@
...
+        }
+      }
+    }
+  }
+
+  def putMetric(account: AwsAccount, dataType: DataType.Value , value: Int): String = {
+    println(s"METRIC:  Account=${account.name},DataType=${dataType},Value=${value}")
```
philmcmahon, line 36: as above

hq/app/logging/Cloudwatch.scala
```diff
@@ -0,0 +1,48 @@
...
+    val cw = AmazonCloudWatchClientBuilder.defaultClient
+
+    val dimension = List(
+      (new Dimension).withName("Account").withValue(account.name),
+      (new Dimension).withName("DataType").withValue(dataType.toString)
+    )
+    val datum = new MetricDatum().withMetricName("Vulnerabilities").withUnit(StandardUnit.Count).withValue(value.toDouble).withDimensions(dimension.asJava)
+    val request = new PutMetricDataRequest().withNamespace("SecurityHQ").withMetricData(datum)
+    val response: PutMetricDataResult = cw.putMetricData(request)
+    response.toString
```
philmcmahon, line 46: what's the motivation of this function returning a string when the calling function returns unit?

hq/app/logging/Cloudwatch.scala
```diff
@@ -0,0 +1,48 @@
...
+    println(s"METRIC:  Account=${account.name},DataType=${dataType},Value=${value}")
+    val cw = AmazonCloudWatchClientBuilder.defaultClient
+
+    val dimension = List(
+      (new Dimension).withName("Account").withValue(account.name),
+      (new Dimension).withName("DataType").withValue(dataType.toString)
+    )
+    val datum = new MetricDatum().withMetricName("Vulnerabilities").withUnit(StandardUnit.Count).withValue(value.toDouble).withDimensions(dimension.asJava)
+    val request = new PutMetricDataRequest().withNamespace("SecurityHQ").withMetricData(datum)
+    val response: PutMetricDataResult = cw.putMetricData(request)
```
philmcmahon, line 45: if you e.g. have invalid credentials what happens here? Might be worth wrapping in a try catch and logging the error on failure

hq/app/services/CacheService.scala
```diff
@@ -25,6 +25,10 @@ import utils.attempt.{Attempt, FailedAttempt, Failure}
 
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
+import logging.Cloudwatch
+import logging.Cloudwatch.putMetric
```
philmcmahon, line 29: unnecessary imports?

hq/app/logging/Cloudwatch.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  object DataType extends Enumeration {
+    val s3Total = Value("s3/total")
+    val iamCredentialsTotal = Value("iam/credentials/total")
+    val iamKeysTotal = Value("iam/keys/total")
+    val sgTotal = Value("securitygroup/total")
+  }
+
+  def logAsMetric[T](data: Seq[T], dataType: DataType.Value ) : Unit = {
+    for ((account: AwsAccount, result) <- data) {
```
philmcmahon, line 20: I think you could also write this as:

```
data.foreach{
 case(account:AWSAccount, Right(details: CredentialReportDisplay)) => 
             putMetric(account, dataType, details.humanUsers.length + details.machineUsers.length)
```

etc. though I'm not certain it will come out looking any cleaner so ignore me if it becomes a mess!

jorgeazevedo:
From talking with Phil just now, our idea for IAM coverage metrics would be
* human access keys older than 90 days
* machine access older than a year
* red (flagged as such in the UI)
* yellow

hq/app/logging/Cloudwatch.scala
jorgeazevedo: I've tried it and it does look cleaner to me! Thanks

jorgeazevedo:
@philmcmahon I've implemented the red/amber metrics.

Thinking about the metrics for old access keys, I've realised that they are marked as ✅ in the UI. See [this example](https://security-hq.gutools.co.uk/iam/frontend).

I understand our interest in tracking these, but the by being green on the UI we're actually not instructing teams to take action on them. Do we need to track them if that's the case? Should we alternatively redefine old keys to be considered a warning state?

hq/app/logging/Cloudwatch.scala
```diff
@@ -0,0 +1,65 @@
...
+    data.foreach {
+      case (account: AwsAccount, Right(details: List[Any])) =>
+        putMetric(account, dataType, details.length)
+      case (account: AwsAccount, Left(_)) =>
+        logger.error(s"Attempt to log cloudwatch metric failed. Data of type ${dataType} is missing for account ${account.name}.")
+    }
+  }
+
+  def putMetric(account: AwsAccount, dataType: DataType.Value , value: Int): Unit = {
+    val cw = AmazonCloudWatchClientBuilder.defaultClient
```
philmcmahon, line 49: I think this should be pulled up into the object so that it's only  instantiated once, and maybe call it cloudwatchClient or similar

philmcmahon (approved):
Just one small change, otherwise LGTM 

post

hq/app/logging/Cloudwatch.scala
jorgeazevedo: Great point! I've gone ahead and committed that change

PR #203, opened 2021-03-17
chore: open HTTPS access on the load balancer to the AUS office

akash1810:
## What does this change?

Opens HTTPS (443) on the load balancer to the AUS office. Requested from Nick E via @mchv:

> Nick E in Aus seems unable to access security-hq from the Aus.

## What is the value of this?

Colleagues can use the tool.

Yes.

## Any additional notes?

n/a

### DIFF
--- a/cloudformation/security-hq.template.yaml
+++ b/cloudformation/security-hq.template.yaml
@@ -11,7 +11,7 @@ Metadata:
     - Label:
         default: Security
       Parameters:
-      - AccessRestrictionCidr
+      - GuardianIpBlock
     - Label:
         default: Networking
       Parameters:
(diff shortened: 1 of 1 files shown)


cloudformation/security-hq.template.yaml
```diff
@@ -11,7 +11,7 @@ Metadata:
     - Label:
         default: Security
       Parameters:
-      - AccessRestrictionCidr
+      - GuardianIpBlock
```
jacobwinch, line 14: Minor: we might want to add `GuardianAUIpBlock` and `GuardianAUIpBlock2` to the same parameter group so that they are [grouped together in the Console](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudformation-interface.html)?

jacobwinch (approved):

akash1810:
This becomes so much better with the recent changes to add GuCDK and private-infrastructure-config. Great work again @nicl !

Closing and will check with Nick E if this is still needed.

PR #207, opened 2021-03-19
Add filtering on severity

Nirvikalpa108:
## What does this change?
This PR adds checkboxes to filter GCP security findings by severity.

Screenshot 2021-03-19 at 18 05 00

### DIFF
--- a/hq/public/javascripts/app.js
+++ b/hq/public/javascripts/app.js
@@ -36,6 +36,7 @@ $(document).ready(function() {
     }
   );
 
+  // filtering table results
   $('.js-finding-filter').change(function() {
     $('#show-ignored-findings')[0].checked
       ? $('.finding-suppressed--true').show()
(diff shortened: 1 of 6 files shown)


hq/public/javascripts/app.js
```diff
@@ -36,17 +36,34 @@ $(document).ready(function() {
     }
   );
 
+  // filtering table results
```
akash1810, line 39: We can use [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) here to simplify this a little. The benefit of `FormData` is we can won't need to repeat this line `$('.js-finding-filter-for-X').change(function() {` for each form, thus keeping things a DRY.

Here's an example https://jsfiddle.net/ahyr1wpc/.

hq/public/stylesheets/main.css
```diff
@@ -423,3 +397,27 @@ table.gcp-report__table {
...
 
+/* GCP */
+table.gcp-report__table td,
+table.gcp-report__table th {
+  min-height: 21px;
+  box-sizing: initial;
+  word-wrap: break-word;
+}
+
+table.gcp-report__table {
```
philmcmahon, line 408: I think you don't need to bother with the `table.` selector as you already have the class name - think we discussed this on a previous PR (I might be wrong though!)

philmcmahon (approved):
This looks great, @akash1810 's suggestion is a good one though!

hq/public/javascripts/app.js
Nirvikalpa108: Thank you so much for detailing this here! I'm going to write up a ticket for this as it would be good to improve this across all the checkboxes used in SHQ.

akash1810: https://github.com/guardian/security-hq/pull/244

PR #209, opened 2021-03-22
Fetch new iam report every 4 hours

jorgeazevedo:
## What does this change?

Set the refresh rate of the credentials report to 4 hours instead of 90 minutes

## What is the value of this?

We're being rate limited by aws on a regular basis. From what I understand - the intention was for this report to be fetched every 4 hours instead of 90 minutes?

Nope

## Any additional notes?

@katebee right? 4 hours instead of 90 minutes?

### DIFF
--- a/hq/app/services/CacheService.scala
+++ b/hq/app/services/CacheService.scala
@@ -175,7 +175,7 @@ class CacheService(
       refreshSgsBox()
     }
 
-    val credentialsSubscription = Observable.interval(initialDelay + 4000.millis, 90.minutes).subscribe { _ =>
+    val credentialsSubscription = Observable.interval(initialDelay + 4000.millis, 4.hours).subscribe { _ =>
       refreshCredentialsBox()
     }
 


katebee:
The IAM credential report can only be generated once every four hours. Updating the cached version of the report is less valuable if the report has not been regenerated since the last time it was retrieved.

I think we will still hit rate limits as it is the number of accounts that we fetch the report for that is the issue, (although this may need more experimentation to confirm) 😄

jorgeazevedo:
Oh I see what you mean. From [the docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_getting-report.html)

> You can generate a credential report as often as once every four hours. When you request a report, IAM first checks whether a report for the AWS account has been generated within the past four hours. If so, the most recent report is downloaded. If the most recent report for the account is older than four hours, or if there are no previous reports for the account, IAM generates and downloads a new report.

So even if we call it every 90 minutes it's OK, it will just return the last generated report. It would be odd for the API to hit a rate limit because the app gets reports within the space of two hours, as opposed to during the process of quickly iterating over the accounts.

I can close this then, thanks!

adamnfish:
This came up again in a chat today, thought I would add some thoughts here in case people find this in the future.

Alas, I don't think this change would help. The rate limit happens because we slam AWS when we run the update. For each account it has to
- trigger a refresh of the credential report
- poll to check if the report is finished
- fetch the newly updated credential report

The trigger and poll steps for an account will be very quick if the credentials report is less than 4 hours old, and slow if the report is out of date and needs to be regenerated.

If I remember correctly, the rate limit is an outgoing one for SHQ (and the AWS account it runs in), not an incoming one for each of the other accounts. This means that it is the aggregated work of sending out multiple back-to-back API calls to the IAM Management API that causes problems, which also explains why this has gotten worse as we've added more accounts to SHQ.

If we run this refresh only every 4 hours, I suspect it just means that when we're rate limited that error will remain in the UI for 4 hours, instead of (probably) being fixed 90 minutes later when it next runs. It also means that the UI will often be very out of date! If a report was 3 hours 59 minutes old when the job runs, it won't be re-generated until our refresh runs again 4 hours later. Instead of reports being 4 hours old, they could be as much as 8.

Instead, I think we should be a little cleverer about how the refresh happens. Instead of blindly refreshing every account on every run, we can choose to only refresh accounts where the existing report has expired. We'll still slam AWS the first time this runs after boot, but on subsequent runs it is only accounts where the cached credential report is more than 4 hours old that will provoke API calls.

PR #210, opened 2021-03-23
Set loglevel to info

jorgeazevedo:
## What does this change?

This modifies both what gets printed to the console and gets sent to central elk.

## What is the value of this?

Currently, central elk is not getting much of anything sent to it, so the data isn't very useful!

When developing locally, it's useful to see more logs in the console than just errors

No

### DIFF
--- a/hq/app/logging/LogConfig.scala
+++ b/hq/app/logging/LogConfig.scala
@@ -3,7 +3,6 @@ package logging
 import java.net.InetSocketAddress
 
 import ch.qos.logback.classic.spi.ILoggingEvent
-import ch.qos.logback.classic.Level
 import ch.qos.logback.classic.{Logger => LogbackLogger}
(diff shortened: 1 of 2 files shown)


hq/app/logging/LogConfig.scala
```diff
@@ -98,6 +99,7 @@ object LogConfig {
             appender.setCredentialsProvider(buildCredentialsProvider(stsRole, config))
             appender.start()
             rootLogger.addAppender(appender)
+            rootLogger.setLevel(Level.INFO)
```
philmcmahon, line 102: I just had a look at another project and I'm pretty sure this is unnecessary - the change to logback.xml will modify the rootLogger

philmcmahon (approved):

hq/app/logging/LogConfig.scala
jorgeazevedo: Ok, thanks for looking further into this, I'll remove this bit

PR #211, opened 2021-03-23
Move cloudwatch code to its own MetricService

jorgeazevedo:
## What does this change?

This PR changes the data collection approach quite a bit. Instead of piggy backing off the existing CacheService, which has a refresh cadence of 5 minute updates for security groups and S3 and 90 minutes for iam, this PR creates an independent MetricService which runs once every 6 hours and fetches the latest data produced my CacheService.

## What is the value of this?

The value is twofold.

First, posting new metrics every 5 minutes is unnecessary for this data's usecase. It's too much resolution for a dataset where we care about daily and weekly trends. We also hit cloudwatch limits when trying to retrieve weeks of data if there's too many datapoints.

Secondly, this PR avoids posting to cloudwatch if the dataset is incomplete. This is slightly counter intuitive, at first glance that's not good because it means we're going to miss a datapoint. Even though that's true, by guaranteeing that every time we post we post everything we make it meaningful to sum different timeseries streams.

Consider this example. We're interested in tracking the total vulnerabilities that the dotcom team still have to fix

| Time | Account  | Security groups | Exposed keys | Total Vulnerabilities |
| -------------| ------------- | ------------- | ------------- | ------------- | 
| 1pm | Frontend  | 5  | 2  | 7  |
| 2pm | Frontend  | 5  | MISSING  | 5  |
| 3pm | Frontend  | 5  | 1  | 6  |

The fact that exposed keys is missing makes the Total Vulnerabilities number unstable. At 2pm it looks like 2 vulnerabilities were fixed and then 1 new was introduced at 3pm.

But if we reject the datapoint

| Time | Account  | Security groups | Exposed keys | Total Vulnerabilities |
| -------------| ------------- | ------------- | ------------- | ------------- | 
| 1pm | Frontend  | 5  | 2  | 7  |
| 3pm | Frontend  | 5  | 1  | 6  |

We get a stable trend that shows an issue was fixed at 3pm.

I'd propose the solution to get the 2pm datapoint is not to post partial datasets, but to re engineer Security HQ to provide stale data when the data refresh process fails.

Nope

### DIFF
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -19,7 +19,7 @@ import play.api.routing.Router
 import play.api.{BuiltInComponentsFromContext, Logging}
 import play.filters.csrf.CSRFComponents
 import router.Routes
-import services.CacheService
+import services.{CacheService, MetricService}
 import utils.attempt.Attempt
 
 import scala.concurrent.Await
@@ -115,6 +115,13 @@ class AppComponents(context: Context)
     efsClients,
     availableRegions,
     securityCenterClient
+  )
+
+  new MetricService(
+    configuration,
+    applicationLifecycle,
+    environment,
+    cacheService
   )
 
   override def router: Router = new Routes(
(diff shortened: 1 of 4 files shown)


philmcmahon (approved):
🚀

PR #212, opened 2021-03-23
How to integrate Snyk using GitHub Actions

ripecosta:
## What does this change?
This PR changes the instructions on how to integrate Snyk to use GitHubs Actions instead of Snyk's GitHub integration.

## What is the value of this?
Snyk's GitHub integration can lead to number of issues, such as lack of immediate feedback or repo scans failing silently. This often leads developers to opt by using CLI integration alongside it. However, CLI integration also has its shortcomings and limitations.
Furthermore, using one or both of these integrations mean that the Snyk config is spread through different systems (Snyk website, GitHub hooks config, local scripts, TeamCity build steps, etc.), making it difficult for new developers to easily understand the Snyk workflow. 

Using GitHub Action to run Snyk solves these and other issues, making it the superior choice for integration.
It consistently updates snyk.io so the state of the repo can be tracked over time and vulnerabilities handled during the team's Snyk rota, it presents developers with clear and immediate information on whether vulnerabilities are present when they submit code, and keeps all of the Snyk related config alongside the code it's corresponds to.

### DIFF
--- a/hq/markdown/snyk.md
+++ b/hq/markdown/snyk.md
@@ -25,14 +25,15 @@ section of the dashboard).
 
 ## Integrating Snyk with your project(s)
 
-By far the most effective way to integrate Snyk is through GitHub Actions. Here's how to do it:
+By far the most effective way to integrate Snyk is through GitHub Actions. You can also add projects via the Snyk UI, but that method has it's limitation and it's less accurate. Here's how to do it using GH Actions:
 
 1. Make sure your repo has the `SNYK_TOKEN` available under 'Organization secrets' (Go to Settings -> Secrets). If it does not, ask an engineering manager to enable it for your repo.
 1. Create a file called `snyk.yml` in the `/.github/workflows/` folder of your repo. Paste one the appropriate code snippets from below into your new file.
-1. Edit the `org` argument in the code snippet to your org's code. We have several organisations in our Snyk account. The code snippet bellow uses the Reader Revenue's code. To get your organisation's code go to the [Snyk dashboard](https://app.snyk.io/org/), select the org you want then obtain the code from the URL.
+1. Edit the `org` argument in the code snippet to your org's code. We have several organisations in our Snyk account. To get your organisation's code go to the [Snyk dashboard](https://app.snyk.io/org/), select the org you want then obtain the code from the URL.  
+![image](https://user-images.githubusercontent.com/48949546/112194614-f6985880-8c00-11eb-946f-a88fdae57662.jpg)
 1. If your package file does not live in the main folder you will need to add a `--file` argument, such as the one in the Node example below. 
 
-![image](https://user-images.githubusercontent.com/48949546/112194614-f6985880-8c00-11eb-946f-a88fdae57662.jpg)
+
 
 
 Setting up this Action will take care of updating your project's entry in snyk.io daily or whenever a new push to `main` takes place. It will also add feedback to commits and PRs, showing developers if their branch has any security vulnerabilities.
@@ -69,7 +70,7 @@ jobs:
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
-          args: --org=the-guardian-cuu --project-name=${{ github.repository }} --file=./app/yarn.lock
+          args: --org=REPLACE_WITH_YOUR_ORG --project-name=${{ github.repository }} --file=./app/yarn.lock
           command: ${{ env.SNYK_COMMAND }}
 
 ```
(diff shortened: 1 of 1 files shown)


philmcmahon:
This looks great! Do you think it would be suited to using a workflow template organisation wide? https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization

philmcmahon:
Also looks like there's a recommendation to 'pin actions to a full length commit sha' https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions

ripecosta:
> This looks great! Do you think it would be suited to using a workflow template organisation wide? https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization

Most likely. That was the next thing I was going to look at.
I would be keen to understand how easy it is to edit the Action as you're importing it for things like editing the `org` parameter or adding a `--file` argument when necessary. The other thing I'd be curious about is whether or not updating the organisation action would update the repos that imported it.

> Also looks like there's a recommendation to 'pin actions to a full length commit sha' https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions

From what I understand this is an argument against using branch names when referencing actions because those branches can be updated with vulnerable code. But tags are not updated, so is there still a need for full length commit sha?

hq/markdown/snyk.md
```diff
@@ -25,11 +25,90 @@ section of the dashboard).
...
+  security:
+    runs-on: ubuntu-latest
+    env:
+      SNYK_COMMAND: test
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+
+      - name: Set command to monitor
+        if: github.ref == 'refs/heads/main'
```
philmcmahon, line 101: can you use `$default-branch` here?

ripecosta: `$default-branch` is a macro that works if you are importing a workflow template. At the import stage it will replace it with the actual branch name so it can't be used in an action unfortunately.

philmcmahon:
I think it would be good to add a note saying something like "you can also add projects via the Snyk UI, which is slightly easier but less accurate"

philmcmahon (approved):
This is a great improvement - just one comment above to maybe let people know there's another less accurate option

hq/markdown/snyk.md
```diff
@@ -25,11 +25,90 @@ section of the dashboard).
...
+        run: echo "SNYK_COMMAND=monitor" >> $GITHUB_ENV
+
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/scala@0.3.0
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          args: --org=the-guardian-cuu --project-name=${{ github.repository }}
```
katebee, line 109: Small suggestion...
Consider adding a comment above this line or replace the org name with an invalid one (as below) to remind people that they will need to change this value. I know changing the org is mentioned on line 32 but the distance increases the chance that the step will be forgotten or missed 😉 

```suggestion
          args: --org=REPLACE_WITH_YOUR_ORG --project-name=${{ github.repository }}
```

rtyley:
Thanks for these updated instructions!

Just to clarify, for us on the Ophan team (where our [main build](https://teamcity.gutools.co.uk/buildConfiguration/Ophan_OphanAll?branch=%3Cdefault%3E&buildTypeTab=overview&mode=builds) is in **TeamCity**), is GitHub Actions still the preferred approach to get Snyk integration?

ripecosta:
> Thanks for these updated instructions!
> 
> Just to clarify, for us on the Ophan team (where our [main build](https://teamcity.gutools.co.uk/buildConfiguration/Ophan_OphanAll?branch=%3Cdefault%3E&buildTypeTab=overview&mode=builds) is in **TeamCity**), is GitHub Actions still the preferred approach to get Snyk integration?

Hi Roberto! This is the recommended way to integrate Snyk for new projects. However migrating your existing project to this method will bring you a number of benefits (4 minutes pres about it [here](https://drive.google.com/file/d/1wUKC5Go-UcpFgE7Gc1Itk1oSDrUOfsNR/view?t=32m8s)).
If you want to have a chat about it or have any questions feel free to reach out. I've done this for a number of projects in SX.

rtyley:
Thanks @ripecosta ! We've just watched your excellent presentation, and we're trying it out with https://github.com/guardian/ophan/pull/4141 - however, we are running into a small problem, which is that we have a (_human_-friendly) check that enforces Java 11 in our build.sbt (see https://github.com/guardian/ophan/pull/4132). However, `snyk monitor` is kicking off a docker image [`snyk/snyk:scala`](https://hub.docker.com/layers/snyk/snyk/scala/images/sha256-69ef0b1c4f56391af2b277ad779ce7268b819eeead060283ccfbfcacdb57cf16?context=explore) that only has Java 8 available, so the build fails! This isn't really related to your advice to use GitHub Actions - that still seems like a good steer - we're just going to have to work out what to do about enforcing/not-enforcing Java 11.

PR #217, opened 2021-04-01
Add row summary with counter

Nirvikalpa108:
## What does this change?
This PR adds a count of High, Low, Medium and Unknown severity findings per project to each table row.

Screenshot 2021-04-01 at 06 04 03

_Please ignore the fact that I have put my own interpretation on the sorting of the alphabet for the project names! I thought I'd just do 3 projects, but ended up wanting to show more!😂_

## What is the value of this?
Developers can quickly see how severe security vulnerabilities are per project.

## Notes for reviewer
I'm not sure about the icon styling - let me know if you think we can improve them to ensure that users understand their meaning. Here is where I got the icons from: https://material.io/resources/icons/?style=baseline

A potential solution could be to add the icons next to the checkboxes.

Another question is do we want to show the icon if there are no findings with that severity? Perhaps we should follow the pattern set in the credentials table where the icon is only shown if there is 1 or more with that value?

### DIFF
--- a/hq/app/model/models.scala
+++ b/hq/app/model/models.scala
@@ -208,19 +208,4 @@ case class GcpFinding(
   recommendation: Option[String]
 )
 
-//TODO delete after /gcp dev work complete
-object GcpFinding {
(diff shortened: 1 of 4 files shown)


philmcmahon (approved):

philmcmahon:
love the icons

PR #218, opened 2021-04-06
Shows severity value icon on /gcp if greater than zero

Nirvikalpa108:
## What does this change?
This PR changes the showing of the severity value summary icon on each row of the /gcp table. Now the icon is only shown when this value is great than zero. This is an improvement on PR https://github.com/guardian/security-hq/pull/217.

Screenshot 2021-04-06 at 06 39 33

## What is the value of this?
This is so much better than when we were showing each icon everytime, because there is less noise on the UI. It's easier to see where the red high severity findings are. Hopefully this should make it easier for developers to see at a glance where they should prioritise their gcp security efforts.

Also, this change follows the pattern already used in SHQ in the S3 buckets and IAM report tables.

### DIFF
--- a/hq/app/views/gcp/gcp.scala.html
+++ b/hq/app/views/gcp/gcp.scala.html
@@ -68,14 +68,22 @@
                     <div class="collapsible-header" tabindex="22">
                     <i class="material-icons">keyboard_arrow_down</i>
                     <span class="iam-header__name">@projectName</span>
-                        <span class="icon-count">@findings.count(_.severity == "High")</span>
-                        <i class="material-icons red-text">error</i>
-                        <span class="icon-count">@findings.count(_.severity == "Medium")</span>
-                        <i class="material-icons orange-text">warning</i>
-                        <span class="icon-count">@findings.count(_.severity == "Low")</span>
-                        <i class="material-icons yellow-text">priority_high</i>
-                        <span class="icon-count">@findings.count(_.severity == "Unknown")</span>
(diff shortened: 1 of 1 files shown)


hq/app/views/gcp/gcp.scala.html
```diff
@@ -68,14 +68,22 @@ <h3>GCP Security Findings</h3>
...
                     <span class="iam-header__name">@projectName</span>
-                        <span class="icon-count">@findings.count(_.severity == "High")</span>
-                        <i class="material-icons red-text">error</i>
-                        <span class="icon-count">@findings.count(_.severity == "Medium")</span>
-                        <i class="material-icons orange-text">warning</i>
-                        <span class="icon-count">@findings.count(_.severity == "Low")</span>
-                        <i class="material-icons yellow-text">priority_high</i>
-                        <span class="icon-count">@findings.count(_.severity == "Unknown")</span>
-                        <i class="material-icons grey-text">not_listed_location</i>
+                        @if(findings.exists(_.severity == "High")) {
```
philmcmahon, line 71: if these strings (low/high) are used in more than one place we should probably turn them into case objects like this  https://github.com/guardian/frontend/blob/968f803d15847ec4ec60a0818506a9c7fcc075f5/common/app/model/Section.scala#L72

Nirvikalpa108: Great idea, thank you @philmcmahon! 

They're not used more than once right now, but if this changes I'll be sure to add this pattern. Thank you!👍🏽

PR #220, opened 2021-04-07
Adds Read more/less toggle button

Nirvikalpa108:
## What does this change?
This PR adds a read more/less toggle button to text inside the GCP table.

Screenshot 2021-04-07 at 13 50 10

## What is the value of this?
Quite a lot of text is returned in some of the cells, so this toggle button should make it easier for users to select the text they want to see and skim over what they don't need to read.

### DIFF
--- a/hq/app/views/gcp/gcpReportBody.scala.html
+++ b/hq/app/views/gcp/gcpReportBody.scala.html
@@ -28,28 +28,21 @@
                         <td class="gcp-table-row-severity"><span>@finding.severity</span></td>
                         <td class="gcp-table-row-category"><span>@finding.category</span></td>
                         <td class="gcp-table-row-date"><span>@finding.eventTime.toString(GcpDisplay.dateTimePattern)</span></td>
-                        <td><span>@finding.explanation</span></td>
                         <td>
-                        @*if the text is less than 40 chars, show it*@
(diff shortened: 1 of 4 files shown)


hq/app/views/gcp/gcpReportBody.scala.html
```diff
@@ -28,10 +28,27 @@
...
+                                <span class="unaltered-short">@finding.explanation</span>} else {
+                                <span class="gcp-toggle gcp-truncated">@GcpDisplay.preview(finding.explanation.getOrElse(""), 30)</span>
+                                <span class="gcp-toggle gcp-full-text">@finding.explanation</span>
+                                <button class="js-read-more mdc-button">
+                                    <span class="mdc-button__label">Read More/Less</span>
+                                </button>
+                            }
+                        </td>
+                        <td>
+                            @if(finding.recommendation.getOrElse("").length < 30) {
```
philmcmahon, line 42: you can do finding.recommendation.exists(_.length < 30)  I think

philmcmahon: or alternatively wrap this entire eblock in a `@finding.recommendatiion.map{ r +> ` to avoid the other .getOrElse

philmcmahon (approved):
just one style suggestion otherwise 👍🏻

PR #227, opened 2021-04-19
Show stale data on credential report failure

jorgeazevedo:
## What does this change?

This change essentially backfills missing credentials reports with existing ones.

It does so by looking at the result from the fetch and replacing tuples with "left"  with tuples with the data currently in memory.

## What is the value of this?

The idea here is that it's more valuable to display the last known good report rather than an error message like this

Discussion___IAM_report___Security_HQ

The age of the report is displayed up top but it doesn't do anything clever to highlight that a report is older than it should be. Given that it's fetched every 90 minutes, I'd expect that the api rate limiting issue wont' leave us with very old reports.

Baton___IAM_report___Security_HQ

Nope

## Any additional notes?

This is probably a little hacky, but we do intend to solve the root cause of the issue (which is the way we're fetching data causing us to be API rate limited) next week or the week after.

### DIFF
--- a/hq/app/logic/DateUtils.scala
+++ b/hq/app/logic/DateUtils.scala
@@ -12,7 +12,9 @@ object DateUtils {
   }
   def toISOString(dateTime: DateTime) = ISODateTimeFormat.dateTime().print(dateTime)
 
-  def dayDiff(date: Option[DateTime]): Option[Long] =  date.map(new Duration(_, DateTime.now(DateTimeZone.UTC)).getStandardDays)
+  def dayDiff(date: Option[DateTime]): Option[Long] =  date.map(dayDiff)
+
+  def dayDiff(date: DateTime): Long =  new Duration(date, DateTime.now(DateTimeZone.UTC)).getStandardDays
 
   def printTime(date: DateTime): String = date.toString("HH:mm")
 }
--- a/hq/app/views/iam/iamCredReportBody.scala.html
+++ b/hq/app/views/iam/iamCredReportBody.scala.html
@@ -2,10 +2,17 @@
 @import model._
 @import logic.CredentialsReportDisplay
 @(report: CredentialReportDisplay)
-
+@message = @{s"Report generated at ${DateUtils.printTime(report.reportDate)} (refreshed every four hours)"}
+@warning = @{". Warning: this data is over a day old."}
 <div class="row">
     <div class="card-panel">
-        <span>Report generated at @{DateUtils.printTime(report.reportDate)} (refreshed every four hours)</span>
+        <span>
+        @if(DateUtils.dayDiff(report.reportDate) > 0) {
+            @{message + warning}
+        } else {
+            @{message}
+        }
+        </span>
     </div>
 </div>
 <div class="row">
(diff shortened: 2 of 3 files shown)


jorgeazevedo:
By increasing the refresh frequency of the credentials report to 1 minute I've been able to get rate limited locally, and indeed I don't see the error message anymore. So this seems to be working

philmcmahon (approved):
👍🏻  great to get this out, I think if you have time would be good to add a message on the end of this line saying 'warning: stale data' so it's clear to people when the data is a bit old https://github.com/guardian/security-hq/blob/main/hq/app/views/iam/iamCredReportBody.scala.html#L8

jorgeazevedo:
Ah thanks for the link to the line in question, that makes it easier. I'll see what I can do, I think it makes sense to display something if it's over a certain number of hours old

jorgeazevedo:
@philmcmahon is this overkill?

Deploy_Tools___IAM_report___Security_HQ

philmcmahon:
> @philmcmahon is this overkill?
> 
I think a bit overkill but it does the job! I think the big orange suggests a problem with  the actual IAM stuff rather than with the data, I would probably put the message inline with the 'report generated at' time

hq/app/logic/DateUtils.scala
```diff
@@ -12,7 +12,9 @@ object DateUtils {
   }
   def toISOString(dateTime: DateTime) = ISODateTimeFormat.dateTime().print(dateTime)
 
-  def dayDiff(date: Option[DateTime]): Option[Long] =  date.map(new Duration(_, DateTime.now(DateTimeZone.UTC)).getStandardDays)
+  def dayDiff(date: Option[DateTime]): Option[Long] =  date.map(dayDiff)
+
+  def dayDiff(date: DateTime): Long =  new Duration(date, DateTime.now(DateTimeZone.UTC)).getStandardDays
```
jorgeazevedo, line 17: @philmcmahon is method overloading like recommendable in scala or best to avoid?

hq/app/views/iam/iamCredReportBody.scala.html
```diff
@@ -2,10 +2,17 @@
 @import model._
 @import logic.CredentialsReportDisplay
 @(report: CredentialReportDisplay)
-
+@message = @{s"Report generated at ${DateUtils.printTime(report.reportDate)} (refreshed every four hours)"}
```
jorgeazevedo, line 5: @philmcmahon what about declaring constants like this in templates?

hq/app/logic/DateUtils.scala
philmcmahon: looks good to me!

hq/app/views/iam/iamCredReportBody.scala.html
philmcmahon: I think this is a neat solution. The alternative would be to store the strings in the scala codebase and pass through to the controller but since they're just strings related to the view that the app doesn't care about I think it's reasonable to put them in the view

philmcmahon (approved):

jorgeazevedo:
brilliant, thanks @philmcmahon !

PR #228, opened 2021-04-20
Add tests for IamNotifier

Nirvikalpa108:
## What does this change?
This PR adds tests to the IamNotifier logic. Here are the functions that are tested and the bullet points underneath them are the ways these have been tested:

### makeCredentialsNotification:List[Either[FailedAttempt, Notification]]
- "makes a credentials notification with a message including **both old access keys and missing mfa**"
- "makes a credentials notification with a message notifying about **old access keys only**"
- "makes a credentials notification with a message notifying about **missing mfas only**"
- "returns a failure when there are **no old access keys or missing mfas**"

### findOldAccessKeys: CredentialReportDisplay
- "returns CredentialReportDisplays with **access keys greater than 90 days old**"
- "returns empty human user and machine user lists when there are **no access keys greater than 90 days old**"

### findMissingMfa: CredentialReportDisplay
- "returns a list of human users in the credential report displays with **missing Mfas**"
- "returns an empty human user list in the credential report displays when there are **no missing Mfas**"

## This PR also does the following:

-  improves the naming of some of the classes 
- fixes a compile error related to `SnykDisplayTest`being named twice in the codebase
- refactors the createMessage function to avoid multiple createMessage functions with different parameter Seq parameters, which was causing a compile error.


hq/app/schedule/IamAudit.scala
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  def findOldAccessKeys(credsReport: CredentialReportDisplay): CredentialReportDisplay = {
+    val filteredMachines = credsReport.machineUsers.filter(user => hasOutdatedKey(List(user.key1, user.key2)))
+    val filteredHumans = credsReport.humanUsers.filter(user => hasOutdatedKey(List(user.key1, user.key2)))
+    credsReport.copy(machineUsers = filteredMachines, humanUsers = filteredHumans)
+  }
+
+  def hasOutdatedKey(keys: List[AccessKey]): Boolean = keys.exists(key => DateUtils.dayDiff(key.lastRotated).getOrElse(1L) > 90L)
```
Nirvikalpa108, line 28: I'm needing to add `L` here, otherwise I get a compile error that the compiler doesn't see these values as Longs. I wonder if this can be improved 😊

hq/app/schedule/IamJob.scala
```diff
@@ -0,0 +1,47 @@
...
+
+    for {
+      snsClient <- snsClients
+    } yield {
+      makeCredentialsNotification(getCredsReport(cacheService)).foreach{ result: Either[FailedAttempt, Notification] =>
+        result match {
+          case Left(error) =>
+            error.failures.foreach(e =>
+              //TODO: how can i manage the option throwable?
+              logger.error(s"failed to collect credentials report for IAM notifier: ${e.friendlyMessage}", e.throwable.getOrElse(throw new Throwable))
```
Nirvikalpa108, line 36: I wonder how best to manage an `Option[Throwable]` here? 😀

hq/test/logic/SnykDisplayTest.scala
```diff
@@ -8,7 +8,7 @@ import scala.concurrent.ExecutionContext.Implicits.global
 
 class SnykDisplayTest extends FreeSpec with Matchers with AttemptValues {
 
-  private def readFile(filename:String) = Source.fromResource(s"logic/SnykDisplayTest/$filename.json").getLines.mkString
+  private def readFile(filename:String) = Source.fromResource(s"logic/SnykDisplayTestResources/$filename.json").getLines.mkString
```
Nirvikalpa108, line 11: This fixes a compile error. The compiler was unhappy, because both the test and the file with the resources had the same name.

hq/app/schedule/IamAudit.scala
philmcmahon: I think the 90L should go in the Config.scala object as 'recommendedRotationCadence' or similar. The L is fine, though you might not need it if you do `val blah: Long = 90`

Also, we have been recommending to teams to rotate machine credentials once a year and human credentials once every 3 months so there might be more than one concept of outdated

hq/app/schedule/IamAudit.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  def findMissingMfa(credsReport: CredentialReportDisplay): CredentialReportDisplay = {
```
philmcmahon, line 30: this function is called `findMissingMfa` but it seems to also do something to the machine users?

hq/app/schedule/IamAudit.scala
philmcmahon, line 28: this function could also check for whether the key is active or not (maybe update the name accordingly) - I don't think initially we should be sending alerts for inactive keys

hq/app/schedule/IamJob.scala
philmcmahon: argh that's awkward, maybe:

```
val errorMessage = s"failed to collect credentials report for IAM notifier: ${e.friendlyMessage}"
e.throwable.fold(logger.error(errorMessage))(t => logger.error(errorMessage, e)), 
```

hq/app/schedule/IamAudit.scala
Nirvikalpa108: Yes, it's filtering out all machine users as they cannot have Mfa. I've improved the naming to make that clearer. If there's a better way of doing this, please let me know! Thanks😀

hq/app/schedule/IamAudit.scala
Nirvikalpa108: Ah fantastic, I wasn't aware of this, that's great! I'll make that change 👍🏽

Nirvikalpa108: Yes I also thought that initially, but Kate has wants us to alert on disabled keys. They said that keys should either be deleted or rotated, because there is always the chance they will be reenabled 👍🏽

philmcmahon (approved):
Yeah :shipit: ! 👍🏻

PR #230, opened 2021-04-21
GCP Dashboard Improvements

philmcmahon:
## What does this change?
This change:
 - Increases page size of GCP queries to 1000 (the default is 10)
 - Adds an extra filter to only show 'active' findings - GCP Security centre keeps a record of all security events that have happened, but we only care about current ones
 - Switches the 'severity' value being used to the one at the top level of the finddings object.

## What is the value of this?
- The GCP query now loads more quickly as it involves fewer requests thanks to the larger page size. 
- We aren't showing non-active findings to users
 - We have fewer 'unknown severity' findings as the top level severity property seems to be more frequently set.

No


Nirvikalpa108 (approved):
Really great improvements @philmcmahon 🎉

PR #231, opened 2021-04-23
fix: Do not flag port 443 as open to the world

akash1810:
## What does this change?

Do not flag a security group with port specifications as from 443 to 443 as open to the world.

This configuration is used to give HTTPS access to the world, which is totally legit.

## What is the value of this?

The alerts sent via Anghammarad include less false positives.

No.

## Any additional notes?

n/a

### DIFF
--- a/lambda/security-groups/src/main/scala/com/gu/hq/SecurityGroups.scala
+++ b/lambda/security-groups/src/main/scala/com/gu/hq/SecurityGroups.scala
@@ -17,12 +17,19 @@ object SecurityGroups {
     elbSGs.contains(sgConfiguration.groupId)
   }
 
+  private[hq] def isExactlyHttps(sGConfiguration: SGConfiguration): Boolean = {
+    sGConfiguration.ipPermissions.exists(ipPermission => {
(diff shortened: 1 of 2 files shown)


jonathonherbert:
Thanks for following up, @akash1810!

To check – would this mean that we're not spotting SGs that enable HTTPS access to things that aren't ELBs?

Or, phrased another way – does the idea of 'open' not include internet access over HTTPS? Is that clear to users (developers?)

akash1810:
> does the idea of 'open' not include internet access over HTTPS? Is that clear to users (developers?)

Good point, maybe this is better expressed as a separate rule, rather than conflating the existing one.

philmcmahon:
I think I would agree with Jon that non-elb things shouldn't have port 443 open ideally so maybe we only add the logic for line 31? Though that doesn't allow for api gateway, and it would be more work to identify those ones.

akash1810:
I've totally forgotten _why_ this is needed (the PR description needs improving!). As it's been open for a while, I'll close this - we can always reopen with an improved PR description if needed.

PR #232, opened 2021-04-27
Add final quartz components for IAM notifier

Nirvikalpa108:
## What does this change?
This PR adds the final changes required to set up the IAM notifier, which sends emails to AWS accounts notifying them of their old permanent credentials which need rotating or their credentials missing multi-factor authentication.

The emails will be sent on the first Monday of every month. The hope is that at the end of the month we automatically disable credentials posing a security risk. This will be a forthcoming piece of work.

## What is the value of this?
This new feature automates the process of checking for old permanent AWS credentials or those missing mfa, so that developers are no longer required to manually check for this. It also paves the way for the automatic disablement of these credentials, which will greatly improve the Guardian's security stance. 

Furthermore, this work has added a reusable scheduling feature, which means that future development work to send notifications for other reasons will be both easier and quicker.

## How has this been tested?
This feature has been tested by sending emails to a test google group. 

**Here is what the emails look like:**
Screenshot 2021-04-27 at 06 22 55

## Are there any follow up changes required?
SHQ's CFN has been updated to include anghammarad's SNS topic ARN. Disable CD here: https://riffraff.gutools.co.uk/deployment/continuous

This change has required a change in SHQ's local config file, so all developers working on it will need to download the latest version from S3 with `aws s3 cp s3://security-dist/security/PROD/security-hq/security-hq.local.conf ~/.gu --profile security` @jorgeazevedo @philmcmahon @katebee 💕

Once deployed, it might be worth sending an email out to the department letting them know about this change. Not only will this raise awareness about IAM creds, but if anyone has any feedback or picks up on a bug, it would be useful to hear about.

### DIFF
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -11,6 +11,7 @@ import config.Config
 import controllers._
 import filters.HstsFilter
 import model.AwsAccount
+import org.quartz.impl.StdSchedulerFactory
 import play.api.ApplicationLoader.Context
 import play.api.libs.ws.WSClient
 import play.api.libs.ws.ahc.AhcWSComponents
@@ -19,6 +20,7 @@ import play.api.routing.Router
 import play.api.{BuiltInComponentsFromContext, Logging}
 import play.filters.csrf.CSRFComponents
 import router.Routes
+import schedule.JobScheduler
 import services.{CacheService, MetricService}
 import utils.attempt.Attempt
 
@@ -135,4 +137,10 @@ class AppComponents(context: Context)
     assets,
     new GcpController(configuration, googleAuthConfig, cacheService)
   )
+
+  //initialise IAM notification service
+  val quartzScheduler = StdSchedulerFactory.getDefaultScheduler
+  val iamJob = new IamJob(true, cacheService, snsClients, configuration)(executionContext)
+  val jobScheduler = new JobScheduler(quartzScheduler, List(iamJob))
+  jobScheduler.initialise()
 }
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -140,7 +140,7 @@ class AppComponents(context: Context)
 
   //initialise IAM notification service
   val quartzScheduler = StdSchedulerFactory.getDefaultScheduler
-  val iamJob = new IamJob(true, cacheService, snsClients, configuration)(executionContext)
+  val iamJob = new IamJob(enabled = true, cacheService, snsClients, configuration)(executionContext)
   val jobScheduler = new JobScheduler(quartzScheduler, List(iamJob))
   jobScheduler.initialise()
 }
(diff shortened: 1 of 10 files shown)


hq/app/schedule/IamJob.scala
```diff
@@ -11,11 +12,11 @@ import utils.attempt.FailedAttempt
...
 import scala.concurrent.ExecutionContext
 
-class IamJob(enabled: Boolean, cacheService: CacheService, snsClients: AwsClients[AmazonSNSAsync])(executionContext: ExecutionContext) extends JobRunner with Logging {
+class IamJob(enabled: Boolean, cacheService: CacheService, snsClients: AwsClients[AmazonSNSAsync], config: Configuration)(executionContext: ExecutionContext) extends JobRunner with Logging {
   override val id = "credentials report job"
   override val description = "Automated emails for old permanent credentials"
-  override val cronSchedule: CronSchedule = CronSchedules.onceADayAt1am
-  val topicArn: String = ??? //TODO retrieve from config
+  override val cronSchedule: CronSchedule = CronSchedules.firstMondayOfEveryMonth
+  val topicArn: String = getAnghammaradSNSTopicArn(config).getOrElse("") //TODO in the event that it is a None, what should we do?
```
Nirvikalpa108, line 19: How would it be best to handle this?

philmcmahon: generally with features like notifications we would not send notifications locally by default, not having the config there is one way of making this happen

Nirvikalpa108: Great, I've made this change thanks 👍🏽

hq/app/AppComponents.scala
```diff
@@ -135,4 +137,10 @@ class AppComponents(context: Context)
     assets,
     new GcpController(configuration, googleAuthConfig, cacheService)
   )
+
+  //initialise IAM notification service
+  val quartzScheduler = StdSchedulerFactory.getDefaultScheduler
+  val iamJob = new IamJob(true, cacheService, snsClients, configuration)(executionContext)
```
philmcmahon, line 143: the first parameter should be named e.g. `IamJob(something=true, cacheService,`

hq/app/schedule/IamNotifier.scala
```diff
@@ -20,15 +20,20 @@ object IamNotifier extends Logging {
...
-      val id = Await.result(response, 5.seconds)
-      logger.info(s"Sent notification to ${notification.target}: $id")
-    } catch {
-      case NonFatal(err) =>
-        logger.error("Failed to send notification", err)
+    topicArn match {
+      case Some(arn) =>
+        val response = Anghammarad.notify(notification, arn, snsClient)
+        try {
+          val id = Await.result(response, 5.seconds)
```
philmcmahon, line 29: rather than await result can you do `response.foreach(id => logger.info(s"Sent notification to ${notification.target}: $id"))`

philmcmahon:
This is looking great! My only concern is that if something goes wrong ( doesn't look like it will based off the code but, you know this is software!) it could send loads of emails and suprise teams - so maybe worth sending an explanation email out before you enable this or before the first batch of emails are likely to be sent

philmcmahon (approved):

PR #236, opened 2021-05-04
Improve the error messages for caches in their initially empty state

adamnfish:
## What does this change?

We now distinguish the "initially empty" error state from a state where we have subsequently failed to update the contents of a cache.

Screenshot 2021-05-04 at 11 26 44

## What is the value of this?

This makes it easier to debug issues with SHQ. When we know the initial population has failed we can give a much more helpful message to the user.

Nope.


katebee (approved):
👍  💎

Nirvikalpa108 (approved):
This is a great change, because now SHQ will output a helpful error message when the cache is empty and just needs time to populate.

jfsoul (approved):
Spotted this old PR and it's 🏅 so I'll merge it. Thanks @adamnfish

PR #238, opened 2021-05-05
Store tags of IAM users and use them to power anghammarad alerts

philmcmahon:
## What does this change?
This PR does two things:

 - Sets security HQ up to enrich the data it gets from the IAM report with IAM User Tags. The current approach is to fail an entire report if the call to 'describe tags' for any of the users within the report fails. This isn't very resilient, and may need more work in future so that even if the tag fetch fails, the IAM report is still available. This work is all in this commit https://github.com/guardian/security-hq/commit/0d92708246498a7a9ab4b96c8aa0571c0a5cabdf **this feature is currently disabled**
 - Modifies `makeCredentialsNotification` so that it splits notifications into groups based off the  tags of the users with issues. Each group has a list of Targets, which are used by anghammarad to determine where to send the alert to. There's a fairly awkward operation here which happens in `getNotificationTargetGroups`

The motivation for opening this PR with the feature disabled is that  the reaper work @Nirvikalpa108 is doing builds on the same functions modified by this work. To avoid mega merge conflicts it's best to get this in place now before trying to deal with the problematic stack sets.

Tested on PROD - doesn't seem to cause any issues!

## What is the value of this?
For teams that share accounts we can send the right alerts to the right mailing list

yes, see https://github.com/guardian/security-hq/pull/237  (merged)

### DIFF
--- a/hq/app/aws/iam/IAMClient.scala
+++ b/hq/app/aws/iam/IAMClient.scala
@@ -54,7 +54,7 @@ object IAMClient {
   )(implicit ec: ExecutionContext): Attempt[CredentialReportDisplay] = {
     val delay = 3.seconds
     val now = DateTime.now()
-    
+
     if(CredentialsReport.credentialsReportReadyForRefresh(currentData, now))
       for {
         client <- iamClients.get(account, SOLE_REGION)
--- a/hq/app/aws/iam/IAMClient.scala
+++ b/hq/app/aws/iam/IAMClient.scala
@@ -6,14 +6,15 @@ import aws.{AwsClient, AwsClients}
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.cloudformation.AmazonCloudFormationAsync
 import com.amazonaws.services.identitymanagement.AmazonIdentityManagementAsync
-import com.amazonaws.services.identitymanagement.model.{GenerateCredentialReportRequest, GenerateCredentialReportResult, GetCredentialReportRequest}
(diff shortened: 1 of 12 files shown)


hq/app/aws/iam/IAMClient.scala
```diff
@@ -46,10 +61,9 @@ object IAMClient {
         _ <- Retry.until(generateCredentialsReport(client), CredentialsReport.isComplete, "Failed to generate credentials report", delay)
         report <- getCredentialsReport(client)
         stacks <- CloudFormation.getStacksFromAllRegions(account, cfnClients, regions)
-        enrichedReport = CredentialsReport.enrichReportWithStackDetails(report, stacks)
-      } yield {
-        CredentialsReportDisplay.toCredentialReportDisplay(enrichedReport)
-      }
+//        reportWithTags <- getCredentialTags(report, client) NOTE: disabled pending fixing the stack set
```
philmcmahon, line 64: Due to problems with the stack set, the feature is disabled for now so that this PR can be merged (to reduce merge conflicts later)

hq/app/model/models.scala
```diff
@@ -157,23 +158,57 @@ object Green extends ReportStatus
...
+    if (ssaTags.nonEmpty) {
+      ssaTags.sortBy(_.key).map(_.value).mkString("-")
+    } else {
+      EMPTY_SSAID
+    }
+  }
+}
+
+
+trait IAMUser
```
jacobwinch, line 190: Minor: would using a `sealed trait` help here?

philmcmahon: good shout

jacobwinch (approved):
LGTM!

🚢 🏃🏖️

PR #243, opened 2021-05-12
send IAM notifications

Nirvikalpa108:
## Problem
The notification job is in PROD, but judging by the logs, the notifications were not sent. 

## Solution
This PR adds a new route to trigger notifications locally, which means that this feature can be tested more easily. Additional logging has been added for debugging purposes.

## Email
Here is what the email message to flagged AWS accounts looks like:

Screenshot 2021-05-14 at 07 16 45

## Impact
This work surfaced a bug with the credentials used for SNS permissions. Once fixed, the logs confirmed that the notifications were sent successfully.

It also surfaced an inaccuracy in the message text for old keys as the logs showed that notifications are also triggered for old keys, which are disabled. The notifications require that inactive users with old access keys should be deleted.

## Next steps
I would like to improve the formatting of the email by adding a table detailing the flagged keys. Here is the ticket, which has been added to the backlog: https://trello.com/c/Z1SHnejg


hq/app/AppComponents.scala
```diff
@@ -99,7 +104,12 @@ class AppComponents(context: Context)
   private val s3Clients = AWS.s3Clients(configuration)
   private val iamClients = AWS.iamClients(configuration, availableRegions)
   private val efsClients = AWS.efsClients(configuration, availableRegions)
-  private val snsClients = AWS.snsClients(configuration)
+  val credsProvider = new AWSCredentialsProviderChain(DefaultAWSCredentialsProviderChain.getInstance(), new ProfileCredentialsProvider("security"))
```
philmcmahon, line 107: could we call this `securityCredentialsProvider` just as there's more than one credentials provider in this app

hq/app/AppComponents.scala
```diff
@@ -99,7 +104,12 @@ class AppComponents(context: Context)
...
+  private val snsClient = AmazonSNSAsyncClientBuilder.standard()
```
philmcmahon, line 108: likewise - `securitySnsClient`

hq/app/controllers/CredentialsController.scala
```diff
@@ -41,4 +42,9 @@ class CredentialsController(val config: Configuration, cacheService: CacheServic
     cacheService.refreshCredentialsBox()
     Ok("Refreshing IAM credentials reports (may take a minute or so to appear)")
   }
+
+  def sendNotifications(send: Boolean) = authAction {
+    iamJob.run()
```
philmcmahon, line 47: oopps! We forgot to wrap this in
```
if (send) iamJob.run()
```

philmcmahon (approved):
A few tiny tweaks, feel free to merge once those are fixed!

PR #244, opened 2021-05-13
refactor: Use `FormData` to react to filter changes

akash1810:
## What does this change?

In this change we refactor multiple jQuery event handlers (one per checkbox) into a single event handler on the enclosing form and use `FormData` to understand what has been selected.

From https://developer.mozilla.org/en-US/docs/Web/API/FormData:

> The `FormData` interface provides a way to easily construct a set of key/value pairs representing form fields and their values

Currently, we're reaching into the DOM a few times which, whilst not expensive on small app, does result in quite a lot of code repetition. `FormData` allows us to be terser and more idiomatic.

## What is the value of this?

The end result of this change for the user is a no-op, though we do get:
  - a single event listener
  - a simpler DOM structure
  - slightly less jQuery

It should also mean adding additional filters is simpler, though this is just a hunch.

No.

## Any additional notes?

Lastly, my IDE runs Prettier on save, so it might be easier to review on a commit by commit basis or with whitespace ignored.

### DIFF
--- a/hq/public/javascripts/app.js
+++ b/hq/public/javascripts/app.js
@@ -48,9 +48,20 @@ $(document).ready(function () {
     const lowFindings = $('.finding-low');
     const unknownFindings = $('.finding-unknown');
 
+    const toggleElements = (checkboxValue, elements) => {
+      if (checkboxValue === undefined) {
+        // field is not on the current page's form, nothing to do
+        return;
+      }
+
+      // Checkboxes have a value of `on` or `off`. Yay HTML forms!
+      return checkboxValue === 'on' ? elements.show() : elements.hide();
+    };
+
     form.addEventListener('input', function () {
       const formData = new FormData(form);
 
+      // values are `undefined` by default, for example if they are not on the current page's form
       const {
         // S3 and Security Group filters
         showFlaggedFindings,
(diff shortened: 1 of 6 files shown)


hq/public/javascripts/app.js
```diff
@@ -7,68 +7,80 @@ jQuery(function($) {
...
-        ? $('.finding-medium').show()
-        : $('.finding-medium').hide();
-    $('#show-low-findings')[0].checked
-        ? $('.finding-low').show()
-        : $('.finding-low').hide();
-    $('#show-unknown-findings')[0].checked
-        ? $('.finding-unknown').show()
-        : $('.finding-unknown').hide();
-  });
+  const form = document.querySelector('form.finding-filter');
```
akash1810, line 40: The interesting change starts here.

hq/public/javascripts/app.js
```diff
@@ -7,68 +7,80 @@ jQuery(function($) {
...
+    const criticalFindings = $('.finding-critical');
+    const highFindings = $('.finding-high');
+    const mediumFindings = $('.finding-medium');
+    const lowFindings = $('.finding-low');
+    const unknownFindings = $('.finding-unknown');
+
+    form.addEventListener('input', function () {
+      const formData = new FormData(form);
+
+      const {
```
akash1810, line 54: [Destructure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) the form; the field names map to the `name` attribute on the `input` element in the form.

hq/public/javascripts/app.js
```diff
@@ -7,68 +7,80 @@ jQuery(function($) {
...
+        showMediumFindings,
+        showLowFindings,
+        showUnknownFindings,
+      } = Object.fromEntries(formData);
+
+      showIgnoredFindings ? ignoredFindings.show() : ignoredFindings.hide();
+      showFlaggedFindings ? flaggedFindings.show() : flaggedFindings.hide();
+      showUnencryptedFindings
+        ? unencryptedFindings.show()
+        : unencryptedFindings.hide();
```
akash1810, line 74: I agree, this is annoyingly spread over multiple lines! I don't argue with Prettier though.

philmcmahon (approved):
⭐ 🕺🏻

akash1810:
This PR changes behaviour in a really subtle way.

On the S3 page, we initially show:
  - Flagged findings
  - Unencrypted findings 
  - A checkbox to show/hide unencrypted buckets (no checkbox for flagged findings as they should always be shown)

This PR assumes a checkbox to exist per type of finding. That is, this PR:
  - Constructs a `FormData` for form results
  - Toggles elements based on the value of the checkbox
  - If the checkbox is not selected, hide the elements

This means on the S3 page this happens:
  - Initial page loads showing flagged and unencrypted findings
  - Unencrypted filter is unchecked, unencrypted findings are hidden so to are flagged findings
  - Unencrypted filter is checked, unencrypted findings are shown, flagged findings are still hidden

This is a breaking change.

It's working on `main` because we're reacting to specific `input` element changes, rather than the generic `form` change in this PR.

If we included a checkbox filter for each finding type, this PR would work. However there is quite a bit of [conditional logic in the templates](https://github.com/guardian/security-hq/blob/8a93b178dd9cb12cf730bbc88ae9ac7d79798225/hq/app/views/s3/publicBucketsAccount.scala.html#L54-L69), that change would be bigger than what this PR attempts.

This PR was attempting to be a no-op. Sadly it isn't. Closing.

akash1810:
Another way to "fix" this would be to treat each form in a specific way, rather than one super generic form handler. That is, using a specific form handler would mean we can guarantee the checkboxes in those forms. The resulting code will likely be longer than what's on `main`. Though it would be more idiomatic, it's probably not worth it.

PR #245, opened 2021-05-13
Log GCP vulnerabilities to cloudwatch metrics

jorgeazevedo:
## What does this change?

This patch adds logging of GCP high and critical vulnerabilities to cloudwatch alongside the existing AWS vulnerabilities metrics.

The fact that GCP vulnerabilities have a project, as opposed to AWS vulnerabilities which have an account, means that their respective cloudwatch metrics will have different dimensions, which means they will have different schemas.

Accounting for this requires a refactor of putMetric into something
parameterised that can be used for both cases.

## What is the value of this?

We'll be able to add gcp vulnerabilities to the security dashoard

Nein

## Any additional notes?

Unlike for example `cacheService.getAllSgs` which basically returns an `Either` that can be unpackaged using pattern matching, `cacheService.getGcpFindings` returns an `Attempt` which I think needs to be unpackaged using a for comprehension? It seems to work, but it's unclear to me if this needs a try or something to handle errors

The diff is made a bit more complicated than expected because unlike the metrics we have so far, it won't have Account/DataType dimensions but Project/DataType. Accommodating these two needs meant making `putMetric` fully parameterised.

### DIFF
--- a/hq/app/logging/Cloudwatch.scala
+++ b/hq/app/logging/Cloudwatch.scala
@@ -3,9 +3,10 @@ package logging
 import com.amazonaws.AmazonServiceException
 import com.amazonaws.services.cloudwatch.AmazonCloudWatchClientBuilder
 import com.amazonaws.services.cloudwatch.model.{Dimension, MetricDatum, PutMetricDataRequest, StandardUnit}
+import com.google.cloud.securitycenter.v1.Finding
 import logic.CredentialsReportDisplay
 import logic.CredentialsReportDisplay.reportStatusSummary
-import model.{AwsAccount, CredentialReportDisplay}
+import model.{AwsAccount, CredentialReportDisplay, GcpFinding}
 import play.api.Logging
 import utils.attempt.FailedAttempt
 
@@ -24,15 +25,32 @@ object Cloudwatch extends Logging {
     val iamCredentialsWarning = Value("iam/credentials/warning")
     val iamKeysTotal = Value("iam/keys/total")
     val sgTotal = Value("securitygroup/total")
+    val gcpTotal = Value("gcp/total")
+    val gcpCritical = Value("gcp/critical")
+    val gcpHigh = Value("gcp/high")
+  }
+
+  def logMetricsForGCPFindings(allGcpFindings: Seq[GcpFinding]): Unit = {
+    val gcpProjectToFinding = allGcpFindings.groupBy(_.project)
+
+    gcpProjectToFinding.toSeq.foreach {
+      case (project: String, findings: Seq[GcpFinding]) =>
+        val criticalFindings = allGcpFindings.filter(_.severity == Finding.Severity.CRITICAL)
+        val highFindings = allGcpFindings.filter(_.severity == Finding.Severity.HIGH)
+        putGcpMetric(project, Cloudwatch.DataType.gcpCritical, criticalFindings.length)
+        putGcpMetric(project, Cloudwatch.DataType.gcpHigh, highFindings.length)
+        putGcpMetric(project, Cloudwatch.DataType.gcpTotal, criticalFindings.length + highFindings.length)
+    }
+
   }
 
   def logMetricsForCredentialsReport(data: Map[AwsAccount, Either[FailedAttempt, CredentialReportDisplay]] ) : Unit = {
     data.toSeq.foreach {
       case (account: AwsAccount, Right(details: CredentialReportDisplay)) =>
         val reportSummary: CredentialsReportDisplay.ReportSummary = reportStatusSummary(details)
-        putMetric(account, DataType.iamCredentialsCritical, reportSummary.errors)
-        putMetric(account, DataType.iamCredentialsWarning, reportSummary.warnings)
-        putMetric(account, DataType.iamCredentialsTotal, reportSummary.errors + reportSummary.warnings)
+        putAwsMetric(account, DataType.iamCredentialsCritical, reportSummary.errors)
+        putAwsMetric(account, DataType.iamCredentialsWarning, reportSummary.warnings)
+        putAwsMetric(account, DataType.iamCredentialsTotal, reportSummary.errors + reportSummary.warnings)
       case (account: AwsAccount, Left(_)) =>
         logger.error(s"Attempt to log cloudwatch metric failed. IAM data is missing for account ${account.name}.")
     }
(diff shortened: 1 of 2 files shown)


hq/app/services/MetricService.scala
```diff
@@ -35,10 +35,15 @@ class MetricService(
...
-    Cloudwatch.logAsMetric(allExposedKeys, Cloudwatch.DataType.iamKeysTotal)
-    Cloudwatch.logAsMetric(allPublicBuckets, Cloudwatch.DataType.s3Total)
-    Cloudwatch.logMetricsForCredentialsReport(allCredentials)
+    for {
+      allGcpFindings <- cacheService.getGcpFindings
+    } yield {
+      Cloudwatch.logAsMetric(allSgs, Cloudwatch.DataType.sgTotal)
+      Cloudwatch.logAsMetric(allExposedKeys, Cloudwatch.DataType.iamKeysTotal)
+      Cloudwatch.logAsMetric(allPublicBuckets, Cloudwatch.DataType.s3Total)
+      Cloudwatch.logMetricsForCredentialsReport(allCredentials)
```
philmcmahon, line 44: I think by moving these calls inside the for/yield, it now means that if `getGcpFindings` fails then nothing in the yield will run. It might be best to keep them outside (assuming I understand this correctly!)

hq/app/logging/Cloudwatch.scala
```diff
@@ -41,25 +59,28 @@ object Cloudwatch extends Logging {
...
+    putMetric("SecurityHQ", "Vulnerabilities", Seq(("Account", account.name),("DataType", dataType.toString)), value)
+  }
+
+  def putGcpMetric(project: String, dataType: DataType.Value , value: Int): Unit = {
+    putMetric("SecurityHQ", "Vulnerabilities", Seq(("GcpProject", project),("DataType", dataType.toString)), value)
+  }
 
-    val datum = new MetricDatum().withMetricName("Vulnerabilities").withUnit(StandardUnit.Count).withValue(value.toDouble).withDimensions(dimension.asJava)
-    val request = new PutMetricDataRequest().withNamespace("SecurityHQ").withMetricData(datum)
+  def putMetric(namespace: String, metricName: String, metricDimensions: Seq[(String, String)] , value: Int): Unit = {
```
philmcmahon, line 76: it might be a good idea to make this functoin `private` to discourage future devs from calling it directly wiith their own custom dimensions - it might be good to stick to the two AWS/GCP options

philmcmahon (approved):
A couple of suggestions but otherwise this LGTM 👍🏻

hq/app/services/MetricService.scala
jorgeazevedo: So that's actually intended behaviour. By only sending complete datasets, it makes it possible to aggregate them over time in a meaningful way.

I'm not sure what happens if the GCP call fails. Will it throw an exception? Does this need a try block?

hq/app/logging/Cloudwatch.scala
jorgeazevedo: Good point!

hq/app/logging/Cloudwatch.scala
```diff
@@ -24,15 +25,32 @@ object Cloudwatch extends Logging {
...
+    val gcpTotal = Value("gcp/total")
+    val gcpCritical = Value("gcp/critical")
+    val gcpHigh = Value("gcp/high")
+  }
+
+  def logMetricsForGCPFindings(allGcpFindings: Seq[GcpFinding]): Unit = {
+    val gcpProjectToFinding = allGcpFindings.groupBy(_.project)
+
+    gcpProjectToFinding.toSeq.foreach {
+      case (project: String, findings: Seq[GcpFinding]) =>
```
jorgeazevedo, line 37: @philmcmahon Is there a better way to iterate over a list of tuples and get variable names for each side of it?

hq/app/services/MetricService.scala
philmcmahon: Is that still desired behaviour even for GCP where we aren't going to be able to correlate the accounts against one another?

The GCP call returns an either so it won't fail - your for comprehension just ignores the failure case.

hq/app/logging/Cloudwatch.scala
philmcmahon: I think this is a good way of doing it!

hq/app/services/MetricService.scala
jorgeazevedo: `> The GCP call returns an either so it won't fail - your for comprehension just ignores the failure case.`
Ahhh, got it!

`> Is that still desired behaviour even for GCP where we aren't going to be able to correlate the accounts against one another?`

Yeah, you're right, I'm pushing for something here that is quite counter intuitive and maybe I need to abandon it. Here's my thinking on this, you tell me if I'm overcomplicating it

So, to answer your question yes - this makes it easy to calculate a total number of vulnerabilities. Say for example we do a query like `Vulnerabilities where DataType contains '/critical'` and we get

| Time | DataType  | Vulnerabilities |
| -------------| ------------- | ------------- |
| 1pm | iam/credentials/critical  | 5  |
| 1pm | gcp/critical  | 3 |
| 2pm | iam/credentials/critical  | 5  |
| 3pm | iam/credentials/critical  | 5  |
| 3pm | gcp/critical  | 2 |
| 4pm | iam/credentials/critical  | 5  |
| 4pm | gcp/critical  | 3 |

If we do a sum aggregated per time we get

| Time | CriticalVulnerabilities |
| -------------| ------------- |
| 1pm | 8  |
| 2pm | 5  |
| 3pm | 7 |
| 4pm | 8 |

The 2pm value is not meaningful.

There is a workaround for this, which is to aggregate on a longer period. This only works if our data _has_ a period that guarantees we'll find a value for every DataType.

In the example it's pretty easy. We could aggregate on day instead.

| Time | CriticalVulnerabilities |
| -------------| ------------- |
| Today | 8... or maybe 7? |

When doing a roll up like this, we can pick one of 6 statistical operations. We can pick for example the minimum (7), maximum (8), average (~7.5), but [there's no function to pick the most recent value](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Statistic), which is annoying. So you can't aggregate this in a way where cloudwatch will add the 4pm gcp/critical with the 4pm /iam/credentials/critical values.

So I guess my pitch is that by sacrificing resolution we get a dataset that's easier to work with. So far this hasn't been problematic, ever since we fixed the rate limiting issue the data is consistent. Do you anticipate GCP will be flaky?

philmcmahon: You make an excellent case. I think you're right and we should do all or nothing - maybe in the future we might alert off 'critical vulnerabilities >0' and it's important that is based off complete data. Maybe just add a comment here to explain why the other metrics are dependent on GCP - could even link directly to this discussion

PR #247, opened 2021-05-21
add dynamoDB and put notification data

Nirvikalpa108:
## What does this change?
This PR adds dynamoDB so that Security HQ's IAM notification job can record which notifications have been successfully
sent. 

## What is the value of this?
The aim is that this data will be used for two reasons. 

First, so that a new feature, the Credentials Reaper, can be built which automatically disables permanent credentials which are either deemed too old or are missing multi factor authentication. This new feature should only disable a credential if 2 notifications have been sent, hence the need to store data on whether the notification has been sent in a database.

The second reason is that we hope this data will provide a helpful audit trail for Infosec.

## How has this been tested?
This work was tested locally using this https://github.com/aws-samples/aws-sam-java-rest. Here are the steps to use this:

start the server: `docker run -p 8000:8000 amazon/dynamodb-local`

create your dynamoDB table (make sure the same table name is used in your application code):
```
aws dynamodb create-table --table-name security-hq-iam-DEV \
--attribute-definitions AttributeName=id,AttributeType=S \
--key-schema AttributeName=id,KeyType=HASH \
--billing-mode PAY_PER_REQUEST \
--endpoint-url http://localhost:8000 \
--region eu-west-1
```
create a local client in your application code:
```
val dynamoDbClient: AmazonDynamoDB = AmazonDynamoDBClientBuilder.standard()
    .withCredentials(securityCredentialsProvider)
    //.withRegion(Config.region.getName)
    .withEndpointConfiguration(new EndpointConfiguration("http://127.0.0.1:8000", Config.region.getName))
    .build()
```
here's an example request using the AWS SDK within your application code:
```
  def put(item: Map[String, AttributeValue]): Unit = try {
    client.putItem(
      new PutItemRequest().withTableName(tableName).withItem(item.asJava))
  } catch {
    case NonFatal(e) =>
    logger.error(s"unable to put item to dynamoDB table: ${e.getMessage}", e)
  }
```

query your local table:
```
aws dynamodb scan --table-name security-hq-iam-DEV \
--endpoint-url http://127.0.0.1:8000 \
--region eu-west-1
```

### DIFF
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -1,10 +1,11 @@
-import aws.AWS.credentialsProvider
 import aws.ec2.EC2
 import aws.{AWS, AwsClient}
 import com.amazonaws.ClientConfiguration
-import com.amazonaws.auth.{AWSCredentialsProviderChain, DefaultAWSCredentialsProviderChain}
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
+import com.amazonaws.auth.{AWSCredentialsProviderChain, DefaultAWSCredentialsProviderChain}
+import com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration
 import com.amazonaws.regions.Regions
+import com.amazonaws.services.dynamodbv2.{AmazonDynamoDB, AmazonDynamoDBClientBuilder}
 import com.amazonaws.services.ec2.AmazonEC2AsyncClientBuilder
 import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagement
 import com.amazonaws.services.sns.AmazonSNSAsyncClientBuilder
@@ -15,7 +16,8 @@ import com.gu.configraun.models._
 import config.Config
 import controllers._
 import filters.HstsFilter
-import model.AwsAccount
+import model.{AwsAccount, IamAuditAlert, IamAuditUser, Warning}
+import org.joda.time.DateTime
 import org.quartz.impl.StdSchedulerFactory
 import play.api.ApplicationLoader.Context
 import play.api.libs.ws.WSClient
@@ -25,7 +27,7 @@ import play.api.routing.Router
 import play.api.{BuiltInComponentsFromContext, Logging}
 import play.filters.csrf.CSRFComponents
 import router.Routes
-import schedule.{IamJob, JobScheduler}
+import schedule.{Dynamo, IamJob, JobScheduler}
 import services.{CacheService, MetricService}
 import utils.attempt.Attempt
 
@@ -112,6 +114,20 @@ class AppComponents(context: Context)
     .build()
   private val securityCenterSettings = SecurityCenterSettings.newBuilder().setCredentialsProvider(Config.gcpCredentialsProvider(configuration)).build()
   private val securityCenterClient = SecurityCenterClient.create(securityCenterSettings)
+  private val dynamoDbClient: AmazonDynamoDB = {
+    configuration.getOptional[String]("stage") match {
+      case Some("DEV") =>
+        AmazonDynamoDBClientBuilder.standard()
+          .withCredentials(securityCredentialsProvider)
+          .withEndpointConfiguration(new EndpointConfiguration("http://127.0.0.1:8000", Config.region.getName))
+          .build()
+      case _ =>
+        AmazonDynamoDBClientBuilder.standard()
+          .withCredentials(securityCredentialsProvider)
+          .withRegion(Config.region.getName)
+          .build()
+    }
+  }
 
   private val cacheService = new CacheService(
     configuration,
@@ -135,9 +151,13 @@ class AppComponents(context: Context)
     environment,
     cacheService
   )
+
+  val iamDynamoDbTableName = Config.getIamDynamoTableName(configuration)
+  val dynamo = new Dynamo(dynamoDbClient, iamDynamoDbTableName)
+
   //initialise IAM notification service
   val quartzScheduler = StdSchedulerFactory.getDefaultScheduler
-  val iamJob = new IamJob(enabled = true, cacheService, securitySnsClient, configuration)(executionContext)
+  val iamJob = new IamJob(enabled = true, cacheService, securitySnsClient, dynamo, configuration)(executionContext)
   val jobScheduler = new JobScheduler(quartzScheduler, List(iamJob))
   jobScheduler.initialise()
 
@@ -152,5 +172,4 @@ class AppComponents(context: Context)
     assets,
     new GcpController(configuration, googleAuthConfig, cacheService)
   )
-
 }
--- a/hq/app/config/Config.scala
+++ b/hq/app/config/Config.scala
@@ -114,4 +114,5 @@ object Config {
   }
 
   def getAnghammaradSNSTopicArn(config: Configuration): Option[String] = config.getOptional[String]("anghammaradSnsArn")
+  def getIamDynamoTableName(config: Configuration): Option[String] = config.getOptional[String]("iamDynamoTableName")
 }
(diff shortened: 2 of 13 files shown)


hq/app/schedule/IamNotifier.scala
```diff
@@ -3,36 +3,54 @@ package schedule
...
 
-  def createNotification(accountName: AwsAccount, targets: List[Target], message: String): Notification = {
-    Notification(subject(accountName), message, List.empty, targets, channel, sourceSystem)
+  def createNotification(
+    accountName: AwsAccount,
+    targets: List[Target],
+    message: String,
+    awsAccountId: String,
+    username: String,
+    emailType: IamAuditNotificationType = Warning
```
Nirvikalpa108, line 25: The functionality that determines whether a notification is a `Warning` or a `Final` notification will be created later, so this is set as `Warning` for now.

hq/app/AppComponents.scala
```diff
@@ -135,9 +141,18 @@ class AppComponents(context: Context)
     environment,
     cacheService
   )
+
+  val iamDynamoDbTableName: String = Config.getIamDynamoTableName(configuration) match {
+    case Some(tableName) => tableName
+    case None =>
+      logger.error("unable to retrieve Iam Dynamo Table Name from config - check that table name is present in security-hq.conf in S3")
+      "error"
```
Nirvikalpa108, line 149: If there is no table name in SHQ's config, the "error" string will be used to initialise the Dynamo class. Then, a `try catch` is used when we call the `put` method in `Dynamo.scala` which will catch the error and log out the error message. These two log lines combined should be good information if we need to debug this particular problem in the future.

hq/app/model/models.scala
```diff
@@ -261,7 +263,16 @@ trait IAMAlert {
...
 
-case class IAMAlertTargetGroup(targets: List[Target], outdatedKeysUsers: Seq[UserWithOutdatedKeys], noMfaUsers: Seq[UserNoMfa])
\ No newline at end of file
+sealed trait IamAuditNotificationType {def name: String}
+object Warning extends IamAuditNotificationType {val name = "Warning"}
+object Final extends IamAuditNotificationType {val name = "Final"}
+
+case class IamAuditEmail(dateNotificationSent: DateTime, notificationType: IamAuditNotificationType)
+case class IamAuditUser(id: String, awsAccount: String, userName: String, emails: List[IamAuditEmail])
+case class IamNotification(iamUser: IamAuditUser, anghammaradNotification: Notification)
```
Nirvikalpa108, line 277: This new class wraps the Anghammarad Notification class, so that we can pass on information about the Iam User, which is needed for recording in DynamoDB.

hq/app/schedule/IamAudit.scala
```diff
@@ -36,30 +35,38 @@ object IamAudit extends Logging {
...
   }
 
-  def makeCredentialsNotification(allCreds: Map[AwsAccount, Either[FailedAttempt, CredentialReportDisplay]]): List[Notification] = {
-    allCreds.toList.flatMap { case (awsAccount, eFCreds) =>
-      eFCreds match {
-        case Right(credsReport) =>
-          val outdatedKeys = outdatedKeysInfo(findOldAccessKeys(credsReport))
-          val missingMfa = missingMfaInfo(findMissingMfa(credsReport))
-          val targetGroups = getNotificationTargetGroups(outdatedKeys, missingMfa)
+  def getFlaggedCredentialsReports(allCreds: Map[AwsAccount, Either[FailedAttempt, CredentialReportDisplay]]): Map[AwsAccount, CredentialReportDisplay] = {
```
Nirvikalpa108, line 38: This PR splits up the previous function, `makeCredentialsNotification` into two separate functions. The first, filters credentials that are old and/or missing mfa and and the second, creates a notification for these flagged credentials. This is helpful, because the reaper feature will need to check DynamoDB to see which notifications have already been sent in the past before sending another notification. This will be forthcoming.

hq/test/schedule/IamAuditTest.scala
```diff
@@ -2,12 +2,10 @@ package schedule
...
 import model.{UserWithOutdatedKeys, _}
-import com.gu.anghammarad.models.{Email, Notification, Preferred, AwsAccount => AwsAccountTarget}
-import model._
 import org.joda.time.DateTime
 import org.scalatest.{FreeSpec, Matchers}
 import schedule.IamAudit._
-import utils.attempt.FailedAttempt
+import schedule.IamMessages.createMessage
 
 class IamAuditTest extends FreeSpec with Matchers {
```
Nirvikalpa108, line 10: I've changed the tests which were testing `makeCredentialsNotification` to now test the `createMessage` function, because this was the main part of the feature that was not covered by other tests.

hq/app/AppComponents.scala
```diff
@@ -112,6 +114,10 @@ class AppComponents(context: Context)
     .build()
   private val securityCenterSettings = SecurityCenterSettings.newBuilder().setCredentialsProvider(Config.gcpCredentialsProvider(configuration)).build()
   private val securityCenterClient = SecurityCenterClient.create(securityCenterSettings)
+  private val dynamoDbClient: AmazonDynamoDB = AmazonDynamoDBClientBuilder.standard()
+    .withCredentials(securityCredentialsProvider)
+    .withRegion(Config.region.getName)
+    .build()
```
philmcmahon, line 120: I think at this point you have the 'stage' property available (either in `Config` or `configuration`) which you could use to work out whether to build a local dynamo client or this one.

hq/app/AppComponents.scala
philmcmahon: It's generally not the best idea to use a String to contain either an error or the result - probably better to keep the value as an option and deal with the option later

hq/app/config/Config.scala
```diff
@@ -114,4 +114,6 @@ object Config {
   }
 
   def getAnghammaradSNSTopicArn(config: Configuration): Option[String] = config.getOptional[String]("anghammaradSnsArn")
+  def getIamDynamoTableArn(config: Configuration): Option[String] = config.getOptional[String]("iamDynamoTableArn")
```
philmcmahon, line 117: this function appears to be unused at the moment - is it for the local client?

hq/app/schedule/Dynamo.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  def put(item: Map[String, AttributeValue]): Unit = try {
+    client.putItem(
+      new PutItemRequest().withTableName(tableName).withItem(item.asJava))
+  } catch {
+    case NonFatal(e) =>
+    logger.error(s"unable to put item to dynamoDB table: ${e.getMessage}", e)
+  }
+
+  def emailToMap(e: IamAuditEmail): Map[String, AttributeValue] = {
```
philmcmahon, line 33: I think we're using a mix of 'email' and 'alert' to refer to the same thing, it might be good where possible to reduce this to just use 'alert' to simplify naming a bit

hq/app/schedule/IamAudit.scala
```diff
@@ -36,30 +35,38 @@ object IamAudit extends Logging {
...
+   allCreds.map { case (awsAccount, either) => either match {
```
philmcmahon, line 39: `maybeReport` or something would be a better name than `either` here I think

hq/app/schedule/IamAudit.scala
philmcmahon, line 38: I think in a future PR we could try to simplify this function further by getting to to take a `Map[AwsAccount, CredentialReportDisplay]` as parameter and dealing with the eithers somewhere else - might make it easier to test as well. out of scope for this work though!

philmcmahon:
One final thing - this pr description is super useful - you could maybe copy the code snippets into a file called `setup-local-dynamo.sh` which performs the docker command and `aws dynamodb create-table`

philmcmahon (approved):

PR #249, opened 2021-06-03
☠️ Credentials Reaper - check if alert already sent

Nirvikalpa108:
## What does this change?

This PR follows the work started in PR #247 on the credentials reaper feature, which aims to disable vulnerable permanent IAM credentials. 

This particular change gets data from a DynamoDB table to check whether a notification has already been sent to a user about a vulnerable credential.

This is valuable, because the reaper feature will send 2 alerts before disabling a credential and will need to check if this has happened before doing so.

The next piece of work on this feature is to check the type of notification sent, which is stored in dynamoDB and change the alert message based on this.

### DIFF
--- a/hq/app/model/models.scala
+++ b/hq/app/model/models.scala
@@ -2,10 +2,8 @@ package model
 
 import com.amazonaws.regions.Region
 import com.google.cloud.securitycenter.v1.Finding.Severity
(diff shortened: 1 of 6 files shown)


philmcmahon (approved):
Nice work!

PR #250, opened 2021-06-03
Improve resiliency of IAM tag fetching

philmcmahon:
## What does this change?
This PR follows  from https://github.com/guardian/security-hq/pull/238 but changes the behaviour. Previously, if a get tags request for any of the users listed in a credential report for an AWS account failed, then the entire report would fail to generate. This modifies the 'enrich with tags' function so that it just returns the report with no tags added in the event that the get tags request fails

## What is the value of this?
Security HQ polls multiple accounts, there's quite a high risk of failure due to issues with the stack set that gives it the permission to crawl an account, or just general network failure. The dashboard should show as much information as possible rather then failing because the tags are missing. 

No


Nirvikalpa108 (approved):
👍🏽

PR #251, opened 2021-06-04
Make tag fetching more resilient, show stack tag in IAM report table

philmcmahon:
## What does this change?
This change adds a 'recover' step to aws sdk requests to fetch tags for the user for when the request fails. It also simplifies the logic around enriching a credentials report with tags (https://github.com/guardian/security-hq/pull/251/commits/dd99f32d35e4dd1c84a4f4bcc9928c4032b8a4ac).

Also, expose the Stack tag in a column in the IAM report ui to make it easier to distinguish credentials in the same account being managed by different teams, it looks like this:
Screenshot 2021-06-23 at 10 20 43


Nirvikalpa108 (approved):
Great! I love the UI change in particular 💯

PR #252, opened 2021-06-07
Add deadline to new flagged credentials

Nirvikalpa108:
This change adds a deadline field to the flagged users with permanent credentials. The deadline is the date that the credential will be automatically disabled. The email alert will include the deadline, so that users can make the required changes, so that the credential is no longer flagged. 

This change currently impacts users receiving their first notification only. The next piece of work on this feature will be to change the type of email alert sent depending on the deadline.

This PR also removes the tests which were testing the string message, because they were painful to work with given all the changes that are still being made to the message content and format. I will consider using a twirl template.


hq/app/schedule/Dynamo.scala
```diff
@@ -73,7 +74,8 @@ class Dynamo(client: AmazonDynamoDB, tableName: Option[String]) extends Attribut
   def alertToMap(e: IamAuditAlert): Map[String, AttributeValue] = {
     Map(
       "date" -> N(e.dateNotificationSent.getMillis),
-      "notificationType" -> S(e.notificationType.name)
+      "disableDeadline" -> N(e.disableDeadline.getMillis)
+      //"notificationType" -> S(e.notificationType.name)
```
philmcmahon, line 78: should these lines still be removed?

philmcmahon (approved):
Just one comment about commenteed out code

hq/app/schedule/Dynamo.scala
Nirvikalpa108: Great spot, thanks Phil! I've removed these commented out lines now.

PR #255, opened 2021-06-10
Merge two vulnerable user types into one

Nirvikalpa108:
Writing the credentials reaper feature managing two separate user case classes was becoming cumbersome. There was one user type for keys which are old and the other for keys that are missing mfa. Each change to the users required double the code to handle each user type. This change makes it easier to complete the credentials reaper feature.

The downside of this change is that the email alert to users will not include why the user has been flagged. This shouldn't be a major problem as these details are including on the security HQ dashboard.

### DIFF
--- a/hq/app/model/models.scala
+++ b/hq/app/model/models.scala
@@ -263,27 +263,10 @@ trait IAMAlert {
 }
 case class IAMAlertTargetGroup(
   targets: List[Target],
-  outdatedKeysUsers: Seq[UserWithOutdatedKeys],
-  noMfaUsers: Seq[UserNoMfa]
+  users: Seq[VulnerableUser]
 )
 
-case class UserWithOutdatedKeys(
-  username: String,
-  key1LastRotation: Option[DateTime],
-  key2LastRotation: Option[DateTime],
-  userLastActiveDay: Option[Long],
-  tags: List[Tag],
-  disableDeadline: Option[DateTime] = None
-) extends IAMAlert
-
-case class UserNoMfa(
-  username: String,
-  userLastActiveDay: Option[Long],
-  tags: List[Tag],
-  disableDeadline: Option[DateTime] = None
-) extends IAMAlert
-
-case class VulnerableUsers(outdatedKeys: Seq[UserWithOutdatedKeys], noMfa: Seq[UserNoMfa])
+case class VulnerableUser(username: String, tags: List[Tag], disableDeadline: Option[DateTime] = None) extends IAMAlert
 
 sealed trait IamAuditNotificationType {def name: String}
 object Warning extends IamAuditNotificationType {val name = "Warning"}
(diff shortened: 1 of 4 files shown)


hq/app/model/models.scala
```diff
@@ -263,27 +263,11 @@ trait IAMAlert {
...
-case class UserNoMfa(
-  username: String,
-  userLastActiveDay: Option[Long],
-  tags: List[Tag],
-  disableDeadline: Option[DateTime] = None
-) extends IAMAlert
-
-case class VulnerableUsers(outdatedKeys: Seq[UserWithOutdatedKeys], noMfa: Seq[UserNoMfa])
+case class VulnerableUser(username: String, tags: List[Tag], disableDeadline: Option[DateTime] = None) extends IAMAlert
+case class VulnerableUsers(users: Seq[VulnerableUser])
```
philmcmahon, line 270: you could probably get rid of this class entirely and just use `Seq[VulnerableUser]` directly

hq/app/schedule/IamAudit.scala
```diff
@@ -56,19 +51,17 @@ object IamAudit extends Logging {
...
 
     // merge groups into IAMAlertTargetGroup seq
     ssaStrings.map { ssaString =>
-      val outdatedKeysUsers =  outdatedKeysGroups.getOrElse(ssaString, Seq())
-      val missingMfaUsers = missingMfaGroups.getOrElse(ssaString, Seq())
+      val users =  groups.getOrElse(ssaString, Seq())
 
       // assume that within a group all tags are the same. Use first element of the group to generate tags
-      val targets = (outdatedKeysUsers ++ missingMfaUsers).headOption.map(k => Tag.tagsToAnghammaradTargets(k.tags)).getOrElse(List())
+      val targets = (users).headOption.map(k => Tag.tagsToAnghammaradTargets(k.tags)).getOrElse(List())
```
philmcmahon, line 62: no brackets needed on (users)

hq/app/schedule/IamMessages.scala
```diff
@@ -1,17 +1,17 @@
...
   val sourceSystem = "Security HQ Credentials Notifier"
-  def outdatedKeysMessage(account: AwsAccount) =
-    s"Please rotate the following IAM access keys in AWS Account ${account.name}/${account.accountNumber} or delete them if they are disabled and unused (if you're already planning to do this, please ignore this message). If this is not rectified before the disable deadline, Security HQ will automatically disable this user:"
-  def missingMfaMessage(account: AwsAccount) =
-    s"Please add multi-factor authentication to the following AWS IAM users in Account ${account.name}/${account.accountNumber}. If this is not rectified before the disable deadline, Security HQ will automatically disable this user:"
+  def message(account: AwsAccount) =
+    s"Please check the following permanent credentials in AWS Account ${account.name}/${account.accountNumber}, which have been flagged as either needing to be rotated or requiring multi-factor authentication (if you're already planning on doing this, please ignore this message). If this is not rectified before the disable deadline, Security HQ will automatically disable this user:"
```
philmcmahon, line 11: not relevant to this change but I don't think "disable deadline" sounds quite right. Maybe just 'deadline'?

philmcmahon (approved):

PR #256, opened 2021-06-10
IAM alert message determined by flagged user deadline

Nirvikalpa108:
## What is the problem?

Security HQ's IAM alerts send an email to a Guardian AWS account if one of the users in this account has permanent credentials which need rotating or are missing multi factor authentication.

Currently one generic email is sent on a monthly basis. The credentials reaper feature will send two emails; a warning email (similar to the current one) and a final email notifying the AWS account that the problematic permanent credential will be disabled the next day. 

Other credentials reaper PRs: #247 and #249.

## What is the solution?
This change makes it possible for SHQ to determine which text it should send in the email; the warning text or the final text.

It determines this by checking the deadline that is assigned to the problematic permanent credential. If the deadline is tomorrow, SHQ knows to send a final email. If the deadline has not been assigned yet, SHQ knows to send a warning email and **set** the deadline.

## Has this been tested?
There are unit tests for some of the Joda DateTime functions. We will manually test the GET request to the database.

### DIFF
--- a/hq/app/schedule/IamAudit.scala
+++ b/hq/app/schedule/IamAudit.scala
@@ -1,29 +1,27 @@
 package schedule
 
-import com.gu.anghammarad.models.{AwsAccount => Account}
+import com.gu.anghammarad.models.{Target, AwsAccount => Account}
 import config.Config.{iamAlertCadence, iamHumanUserRotationCadence, iamMachineUserRotationCadence}
 import logic.DateUtils
 import model._
 import org.joda.time.{DateTime, Days}
 import play.api.Logging
-import schedule.IamMessages.createMessage
+import schedule.IamMessages._
 import schedule.IamNotifier.createNotification
 import utils.attempt.FailedAttempt
 
 object IamAudit extends Logging {
 
-  def makeIamNotification(flaggedCreds: Map[AwsAccount, Seq[IAMAlertTargetGroup]]): List[IamNotification] = {
-    flaggedCreds.toList.flatMap { case (awsAccount, targetGroups) =>
+  def makeIamNotification(flaggedCredsToNotify: Map[AwsAccount, Seq[IAMAlertTargetGroup]]): List[IamNotification] = {
+    flaggedCredsToNotify.toList.flatMap { case (awsAccount, targetGroups) =>
       if (targetGroups.isEmpty) {
         logger.info(s"found no IAM user issues for ${awsAccount.name}. No notification required.")
         None
       } else {
         targetGroups.map { tg =>
-          logger.info(s"for ${awsAccount.name}, generating iam notification message for ${tg.users.length} user(s) with outdated keys and missing mfa")
-          tg.users.flatMap { user =>
-            val message = createMessage(tg.users, awsAccount)
-            Some(createNotification(awsAccount, tg.targets :+ Account(awsAccount.accountNumber), message, user.username, createDeadline(user.disableDeadline)))
-          }
+          logger.info(s"for ${awsAccount.name}, generating iam notification message for ${tg.users.length} user(s) with outdated keys and and/or missing mfa")
+          val (usersToReceiveWarningAlerts, usersToReceiveFinalAlerts) = sortUsersIntoWarningOrFinalAlerts(tg.users)
+          createNotifications(warning = true, usersToReceiveWarningAlerts, awsAccount, tg.targets) ++ createNotifications(warning = false, usersToReceiveFinalAlerts, awsAccount, tg.targets)
         }
       }
     }.flatten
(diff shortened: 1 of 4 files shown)


hq/app/schedule/IamAudit.scala
```diff
@@ -6,29 +6,51 @@ import logic.DateUtils
...
+
+          val warningNotification = usersToReceiveWarningAlerts.map { user =>
+            val message = createWarningMessage(awsAccount, usersToReceiveWarningAlerts)
+            createNotification(awsAccount, tg.targets :+ Account(awsAccount.accountNumber), message, warningSubject(awsAccount), user.username, createDeadline(user.disableDeadline))
           }
+
+          val finalNotification = usersToReceiveFinalAlerts.map { user =>
+          val message = createFinalMessage(awsAccount, usersToReceiveWarningAlerts)
+          createNotification(awsAccount, tg.targets :+ Account(awsAccount.accountNumber), message, warningSubject(awsAccount), user.username, createDeadline(user.disableDeadline))
+           }
```
philmcmahon, line 40: the above two blocks look pretty similar - can they be simplified, maybe using a function?

hq/app/schedule/IamAudit.scala
```diff
@@ -6,29 +6,51 @@ import logic.DateUtils
...
       } else {
         targetGroups.map { tg =>
-          logger.info(s"for ${awsAccount.name}, generating iam notification message for ${tg.users.length} user(s) with outdated keys and missing mfa")
-          tg.users.flatMap { user =>
-            val message = createMessage(tg.users, awsAccount)
-            Some(createNotification(awsAccount, tg.targets :+ Account(awsAccount.accountNumber), message, user.username, createDeadline(user.disableDeadline)))
+          logger.info(s"for ${awsAccount.name}, generating iam notification message for ${tg.users.length} user(s) with outdated keys and and/or missing mfa")
+
+          val (usersToReceiveWarningAlerts, usersToReceiveFinalAlerts) =
+            tg.users.foldRight[(Seq[VulnerableUser], Seq[VulnerableUser])]((Seq.empty, Seq.empty)) { case (user, (warningAcc, finalAcc)) =>
```
philmcmahon, line 25: I think this foldRight should go into its own function

hq/app/schedule/IamAudit.scala
```diff
@@ -6,29 +6,51 @@ import logic.DateUtils
...
         }
       }
     }.flatten
   }
 
+  def createDeadline(date: Option[DateTime]): DateTime = date.getOrElse(DateTime.now.plusDays(iamAlertCadence))
+  // if the deadline is in 3 weeks OR in 1 week, return true
+  def isWarningAlert(date: DateTime): Boolean = ???
+  // if deadline is tomorrow, return true
+  def isFinalAlert(date: DateTime): Boolean = ???
```
philmcmahon, line 52: these functions look quite easy to test!

hq/app/schedule/IamAudit.scala
```diff
@@ -1,34 +1,55 @@
...
-          }
+          logger.info(s"for ${awsAccount.name}, generating iam notification message for ${tg.users.length} user(s) with outdated keys and and/or missing mfa")
+          val (usersToReceiveWarningAlerts, usersToReceiveFinalAlerts) = sortAlertsIntoWarningOrFinalAlert(tg.users)
+          createNotifications(warning = true, usersToReceiveWarningAlerts, awsAccount, tg.targets) ++ createNotifications(warning = false, usersToReceiveFinalAlerts, awsAccount, tg.targets)
         }
       }
     }.flatten
   }
 
+  private def sortAlertsIntoWarningOrFinalAlert(users: Seq[VulnerableUser]): (Seq[VulnerableUser], Seq[VulnerableUser]) = {
```
jorgeazevedo, line 30: I struggled to read this method a bit. This is creating a List where isWarningAlert is true and another where isFinalAlert is true correct? Could the implementation be streamlined a bit by calling `filter` twice instead of a single call to `foldRight`?

hq/app/schedule/IamAudit.scala
```diff
@@ -1,34 +1,55 @@
...
+      createNotification(awsAccount, targets :+ Account(awsAccount.accountNumber), message, warningSubject(awsAccount), user.username, createDeadline(user.disableDeadline))
+    }
+  }
+
+  private def createDeadline(date: Option[DateTime]): DateTime = date.getOrElse(DateTime.now.plusDays(iamAlertCadence))
+
+  private def isWarningAlert(deadline: DateTime, today: DateTime = DateTime.now): Boolean = {
+    deadline == today.plusWeeks(1) || deadline == today.plusWeeks(3)
+  }
+  private def isFinalAlert(deadline: DateTime, today: DateTime = DateTime.now): Boolean = deadline == today.plusDays(1)
```
jorgeazevedo, line 51: I think the intention of this function is to compare if `deadline` and `today.plusDays(1)` have the same calendar date (e.g. `2021-04-01`), but from what I gather the `==` operator calls [equals](http://joda-time.sourceforge.net/apidocs/org/joda/time/base/AbstractInstant.html#equals%28java.lang.Object%29) which according to the docs

> Compares this object with the specified object for equality based on the millisecond instant, chronology and time zone.

So I think in practice this will always return false since it's effectively comparing the full time stamp (e.g. `2021-04-01T19:05:52.182+01:00`).

I wasn't able to run these methods in isolation in the cli so I haven't been able to confirm. Am I way off? I think it'd be worthwhile to have some unit tests for this method and `isWarningAlert` to express the intention with some examples. It's very easy for code that uses the calendar to behave in unexpected ways, I've been burned by this kind of thing in the past!

There's some prior art for tests using DateTime in CredentialsReportTest.scala. If this is an issue, possible solutions for this include calling [withTimeAtStartOfDay](https://stackoverflow.com/questions/13764106/comparing-two-dates-using-joda-time) or using the [DateTimeComparator](https://www.joda.org/joda-time/apidocs/org/joda/time/DateTimeComparator.html#getDateOnlyInstance--)

Nirvikalpa108: @jorgeazevedo you're spot on! 

Thank you so much for identifying this and another huge thank you for identifying a pattern in the codebase that I can follow and potential Joda DateTime functions that I can use. What a star! Amazing!⭐️😃💯

hq/app/schedule/IamAudit.scala
Nirvikalpa108: Thank you so much for this feedback @jorgeazevedo. I love writing `fold` functions 😂  , but I love even more to write readable code, so this feedback is much appreciated and I have changed the function to use `filter` 💖

hq/app/schedule/IamAudit.scala
```diff
@@ -43,6 +41,65 @@ object IamAudit extends Logging {
...
+  private def findVulnerableUsers(report: CredentialReportDisplay): Seq[VulnerableUser] = {
+    outdatedKeysInfo(findOldAccessKeys(report)) ++ missingMfaInfo(findMissingMfa(report))
+  }
+
+  // if the user is not present in dynamo, that means they've never been alerted before, so mark them as ready to be alerted
+  private def filterUsersToAlert(users: Seq[VulnerableUser], awsAccount: AwsAccount, dynamo: Dynamo): Seq[VulnerableUser] = {
+    getExistingDeadlines(users, awsAccount, dynamo).filter { user =>
+      user.disableDeadline match {
+        case Some(u) => isWarningAlert(u) || isFinalAlert(u)
+        case None => true
```
philmcmahon, line 80: Scala experts in my first team at the guardian were of the opinion that pattern matching on options is rarely a good idea so I've sort of never done it. This is pretty readable though!

`user.disableDeadline.exists(u => isWarningAlert(u) || isFinalAlert(u)) || user.disableDeadline.isEmpty` would be another way of writing this

hq/app/schedule/IamAudit.scala
```diff
@@ -43,6 +41,65 @@ object IamAudit extends Logging {
...
+    }
+  }
+
+  // adds deadline to users when this field is present in dynamoDB
+  private def getExistingDeadlines(users: Seq[VulnerableUser], awsAccount: AwsAccount, dynamo: Dynamo): Seq[VulnerableUser] = {
+    users.map { user =>
+      dynamo.getAlert(awsAccount, user.username) match {
+        case Some(u) => user.copy(user.username, user.tags, Some(getNearestDeadline(u.alerts)))
+        case None => user
+      }
```
philmcmahon, line 91: I think I'd rewrite this pattern match as a `.map` with a `.getOrElse(user)` on the end

hq/app/schedule/IamAudit.scala
```diff
@@ -43,6 +41,65 @@ object IamAudit extends Logging {
...
+    }
+  }
+
+  private def getNearestDeadline(alerts: List[IamAuditAlert], today: DateTime = DateTime.now): DateTime = {
```
philmcmahon, line 95: I think this functon needs a test ideally

philmcmahon (approved):
Great work 👍🏻

PR #260, opened 2021-06-21
Add endpoint to get historical IAM alerts report

Nirvikalpa108:
This PR adds a new route, which returns all of the data in the IAM dynamoDB table. This is useful because a) it's helped us
test that PR #256 works! and b) it means that any user can see which alerts have been sent and what the deadlines are on any flagged credentials.

### DIFF
--- a/hq/app/schedule/Dynamo.scala
+++ b/hq/app/schedule/Dynamo.scala
@@ -1,7 +1,7 @@
 package schedule
 
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
-import com.amazonaws.services.dynamodbv2.model.{AttributeValue, GetItemRequest, PutItemRequest}
+import com.amazonaws.services.dynamodbv2.model.{AttributeValue, GetItemRequest, PutItemRequest, ScanRequest}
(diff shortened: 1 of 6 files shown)


hq/app/schedule/Dynamo.scala
```diff
@@ -27,13 +27,35 @@ class Dynamo(client: AmazonDynamoDB, tableName: Option[String]) extends Attribut
       "error"
   }
 
+  def scan: Seq[Map[String, AttributeValue]] = {
+    client.scan(new ScanRequest().withTableName(table)).getItems.asScala.map(_.asScala.toMap)
+  }
+
+  def scanAlert(): Seq[IamAuditUser] = {
```
philmcmahon, line 34: I forgot to mention this at the time but I think it would be god to long term add some error handling here  so that if the thing in the database  is e.g. missing 'alerts' then we return a HTTP error explaining what's gone wrong. but initially you could wrap this in a Try and in the failure case just return an empty list - I think that might be slightly better than just having an exception

philmcmahon (approved):

PR #270, opened 2021-06-28
Fix testing and get notifications endpoints

Nirvikalpa108:
The test and get notifications endpoints were not working. This PR fixes the dynamoDB parsing fields, which were incorrect. It removes the default parameter and corrects a log line. This change has been tested on PROD and locally.


philmcmahon (approved):
👍🏻

PR #271, opened 2021-07-01
Fix duplicate IAM notification alerts

Nirvikalpa108:
## Problem
Security HQ sends emails to AWS accounts which have permanent credentials that pose a security risk. After making some changes to this feature in order to implement the credentials reaper (this will automate the disabling of credentials), the tests on PROD showed that more than one email was being sent per AWS account. Sometimes the emails would include a subsection of the flagged users, instead of all flagged users in one email. 

## Solution
We've fixed the code that was causing multiple emails to be sent. 

In addition, we also added that only the ophan account should split emails by tags, as this AWS account is shared by both the ophan and data lake teams. We did this, because it was causing emails to be split by stack for all the AWS accounts when the benefit of this was not clear for other teams. 

## Testing
This has been tested locally and on PROD with the new /test-notifications route.

### DIFF
--- a/hq/app/model/models.scala
+++ b/hq/app/model/models.scala
@@ -288,5 +288,5 @@ case class IamAuditUser(id: String, awsAccount: String, username: String, alerts
 object IamAuditUser {
   implicit val iamAuditUserWrites = Json.writes[IamAuditUser]
 }
-case class IamNotification(iamUser: IamAuditUser, anghammaradNotification: Notification)
+case class IamNotification(warningN: Option[Notification], finalN: Option[Notification], alertedUsers: Seq[IamAuditUser])
 
--- a/hq/app/schedule/IamAudit.scala
+++ b/hq/app/schedule/IamAudit.scala
@@ -76,13 +76,16 @@ object IamAudit extends Logging {
 
   // if the user is not present in dynamo, that means they've never been alerted before, so mark them as ready to be alerted
   private def filterUsersToAlert(users: Seq[VulnerableUser], awsAccount: AwsAccount, dynamo: Dynamo): Seq[VulnerableUser] = {
-    getExistingDeadlines(users, awsAccount, dynamo).filter { user =>
+    val usersWithDeadline = enrichUsersWithDeadline(users, awsAccount, dynamo)
+    logger.info(s"(maybe) enriched users ${usersWithDeadline.map(_.username).mkString(",")}")
+    usersWithDeadline.filter { user =>
+
       user.disableDeadline.exists(u => isWarningAlert(u) || isFinalAlert(u)) || user.disableDeadline.isEmpty
     }
   }
 
   // adds deadline to users when this field is present in dynamoDB
-  private def getExistingDeadlines(users: Seq[VulnerableUser], awsAccount: AwsAccount, dynamo: Dynamo): Seq[VulnerableUser] = {
+  private def enrichUsersWithDeadline(users: Seq[VulnerableUser], awsAccount: AwsAccount, dynamo: Dynamo): Seq[VulnerableUser] = {
     users.map { user =>
       dynamo.getAlert(awsAccount, user.username).map { u =>
         user.copy(user.username, user.tags, Some(getNearestDeadline(u.alerts)))
--- a/hq/app/schedule/IamAudit.scala
+++ b/hq/app/schedule/IamAudit.scala
@@ -1,30 +1,53 @@
 package schedule
 
-import com.gu.anghammarad.models.{Target, AwsAccount => Account}
+import com.gu.anghammarad.models.{Notification, Target, AwsAccount => Account}
 import config.Config.{iamAlertCadence, iamHumanUserRotationCadence, iamMachineUserRotationCadence}
 import logic.DateUtils
 import model._
 import org.joda.time.{DateTime, Days}
 import play.api.Logging
 import schedule.IamMessages._
-import schedule.IamNotifier.createNotification
+import schedule.IamNotifier.notification
 import utils.attempt.FailedAttempt
 
 object IamAudit extends Logging {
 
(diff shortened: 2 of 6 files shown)


hq/app/schedule/IamJob.scala
```diff
@@ -28,11 +29,17 @@ class IamJob(enabled: Boolean, cacheService: CacheService, snsClient: AmazonSNSA
...
       for {
         _ <- send(notification, topicArn, snsClient, testMode)
-        _ = dynamo.putAlert(notification.iamUser)
+        _ = users.map(dynamo.putAlert)
       } yield ()
     }
+
+    makeNotification(getFlaggedCredentialsReports(credsReport, dynamo)).foreach { iamNotification =>
+      //TODO fix option handling
+      if (iamNotification.warningNotification.nonEmpty) sendNotification(iamNotification.warningNotification.get, iamNotification.alertedUsers)
```
philmcmahon, line 41: `iamNotification.warningNotification.map(wn => sendNotification(wn, iamNotification.alertedUsers))` and I think you can do the same thing below

hq/app/schedule/IamAudit.scala
```diff
@@ -1,30 +1,56 @@
...
+      )
+    }
+  }
+
+  def createWarningAndFinalNotification(tg: IAMAlertTargetGroup, awsAccount: AwsAccount): (Option[Notification], Option[Notification]) = {
+    logger.info(s"for ${awsAccount.name}, generating iam notification message for ${tg.users.length} user(s) with outdated keys and and/or missing mfa")
+    val (usersToReceiveWarningAlerts, usersToReceiveFinalAlerts) = sortUsersIntoWarningOrFinalAlerts(tg.users)
+
+    val warningNotifications =
+      if (usersToReceiveWarningAlerts.nonEmpty)
```
philmcmahon, line 44: you can use map here as well I think

hq/app/schedule/IamJob.scala
Nirvikalpa108: Thanks for this 👍🏽. 
IntelliJ is being silly and complaining that map shouldn't return Unit. I read about using pattern matching instead: https://stackoverflow.com/questions/13268017/whats-the-idiomatic-way-to-do-a-map-getorelse-returning-unit-in-scala. If you feel really against this, let me know. Otherwise, this is all working and tested on PROD 🎉

hq/app/schedule/IamJob.scala
```diff
@@ -28,11 +29,25 @@ class IamJob(enabled: Boolean, cacheService: CacheService, snsClient: AmazonSNSA
...
+    def sendNotificationAndRecord(notification: Notification, users: Seq[IamAuditUser]): Unit = {
       for {
         _ <- send(notification, topicArn, snsClient, testMode)
-        _ = dynamo.putAlert(notification.iamUser)
+        _ = users.map(dynamo.putAlert)
       } yield ()
     }
+
+    def sendNotification(notification: IamNotification): Unit = {
+      notification.warningN match {
```
philmcmahon, line 40: notification.warningN.foreach(alert => sendNotificationAndRecord(alert, notification.alertedUsers))

hq/app/model/models.scala
```diff
@@ -288,5 +288,5 @@ case class IamAuditUser(id: String, awsAccount: String, username: String, alerts
 object IamAuditUser {
   implicit val iamAuditUserWrites = Json.writes[IamAuditUser]
 }
-case class IamNotification(iamUser: IamAuditUser, anghammaradNotification: Notification)
+case class IamNotification(warningN: Option[Notification], finalN: Option[Notification], alertedUsers: Seq[IamAuditUser])
```
philmcmahon, line 291: I would rename warningN -> warning as the 'notification' bit should be obvious from the type

Nirvikalpa108: I was thinking the same, but I get a keyword clash for `final`, so I thought I would keep them both the same.

philmcmahon: 👍🏻

hq/app/schedule/IamAudit.scala
```diff
@@ -1,30 +1,56 @@
...
       if (targetGroups.isEmpty) {
         logger.info(s"found no IAM user issues for ${awsAccount.name}. No notification required.")
         None
       } else {
-        targetGroups.map { tg =>
-          logger.info(s"for ${awsAccount.name}, generating iam notification message for ${tg.users.length} user(s) with outdated keys and and/or missing mfa")
-          val (usersToReceiveWarningAlerts, usersToReceiveFinalAlerts) = sortUsersIntoWarningOrFinalAlerts(tg.users)
-          createNotifications(warning = true, usersToReceiveWarningAlerts, awsAccount, tg.targets) ++ createNotifications(warning = false, usersToReceiveFinalAlerts, awsAccount, tg.targets)
+        targetGroups.map{ tg =>
+          val (warningNotification, finalNotification) = createWarningAndFinalNotification(tg, awsAccount)
```
philmcmahon, line 22: I think these 3 lines could be neater  if you made the return type of `createWarningAndFinalNotification` an `IamNotification` and make it take `users` as a parameter - this avoids the awkward (Option, Option) tuple return type

philmcmahon (approved):
Nice work, just one final suggestion about simplifying one of the return types.

hq/app/schedule/IamAudit.scala
Nirvikalpa108: This is a great idea, thanks!

hq/app/schedule/IamAudit.scala
Nirvikalpa108: Hey, we discussed putting this if statement logic into the `createNotification` function, which is a great idea except warning and final notifications need different boolean parameters to differentiate the type of notification created. I think this will have to stay as is for now.

PR #276, opened 2021-07-12
Use report time instead of `now` for GCP report

jfsoul:
## What does this change?
This fixes an issue where the GCP report findings each claimed to have been generated at the current time, rather than when the report was actually run (every 4 hours, and when redeployed).

Instead of creating a `GcpReport` type, which contains the report time, during the lifecycle of the request, we instead make that the type stored by the GCP `box` in the cache service and write the current date/time when the data is fetched from GCP and the box is filled in the `CacheService`.

## What is the value of this?
Accurate report dates on the GCP page.

No

## Any additional notes?
No

### DIFF
--- a/hq/app/controllers/GcpController.scala
+++ b/hq/app/controllers/GcpController.scala
@@ -2,8 +2,6 @@ package controllers
 
 import auth.SecurityHQAuthActions
 import com.gu.googleauth.GoogleAuthConfig
-import model.GcpReport
-import org.joda.time.DateTime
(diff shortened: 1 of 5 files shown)


jfsoul:
[Fix 'report generated at' for GCP dashboard](https://trello.com/c/2AxL3KVK/703-fix-report-generated-at-for-gcp-dashboard)

jorgeazevedo (approved):
Nicely done!

PR #279, opened 2021-07-15
Should security HQ disable a user today?

Nirvikalpa108:
This PR adds logic to pull the data from dynamoDB that tells us if an AWS account has been alerted about users with vulnerable permanent credentials in their account.

It checks to see if there are any deadlines that should be disabled today. Unit tests have been added to check that the DateTime check is working. I'm happy that the logic for pulling from the db is working as `scanAlert` is very similar to `getAlert` and is currently working nicely in PROD.


jorgeazevedo (approved):
Seems pretty straightforward!

PR #281, opened 2021-07-21
Improve log line message

Nirvikalpa108:
Improves this log line message, so that it makes debugging of this feature easier in the future.

### DIFF
--- a/hq/app/schedule/IamFlaggedUsers.scala
+++ b/hq/app/schedule/IamFlaggedUsers.scala
@@ -18,7 +18,7 @@ object IamFlaggedUsers extends Logging {
     allCreds.map { case (awsAccount, maybeReport) => maybeReport match {
       case Left(error) =>
         error.failures.foreach { failure =>
-          val errorMessage = s"failed to collect credentials report for IAM notifier: ${failure.friendlyMessage}"
+          val errorMessage = s"failed to collect credentials report display for ${awsAccount.name}: ${failure.friendlyMessage}"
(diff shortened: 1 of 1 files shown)


hq/app/schedule/IamFlaggedUsers.scala
```diff
@@ -18,7 +18,7 @@ object IamFlaggedUsers extends Logging {
     allCreds.map { case (awsAccount, maybeReport) => maybeReport match {
       case Left(error) =>
         error.failures.foreach { failure =>
-          val errorMessage = s"failed to collect credentials report for IAM notifier: ${failure.friendlyMessage}"
+          val errorMessage = s"failed to collect credentials report display for ${awsAccount.name}: ${failure.friendlyMessage}"
```
jfsoul, line 21: Is it worth leaving `IAM` in the message somewhere to explicitly differentiate from the GCP creds report?
```suggestion
          val errorMessage = s"failed to collect IAM credentials report display for ${awsAccount.name}: ${failure.friendlyMessage}"
```

jfsoul (approved):

hq/app/schedule/IamFlaggedUsers.scala
Nirvikalpa108: Thanks for taking a look at this @jfsoul. My understanding is that there's no need, because a `CredentialReportDisplay` is made up of `IamUser`s, so it's AWS specific.

PR #223, opened 2021-04-12
Add notifier

Nirvikalpa108:
## What does this change?
This PR starts the work of adding a scheduler and notification service to security hq. 

## What is the value of this?
The aim is to send automated notifications about insecure credentials.

Ideally, every week (or another time period!) SHQ will check the ages of all permanent credentials. If the age is > 90 days an email will be sent to the email address associated with the AWS account contained in anghammarad. (A follow up piece of work is, if the age goes over 120 days (or another time period!) delete the credentials!)

The scheduling service is reusable, so if in the future we would like SHQ to send notifications for other reasons, this is possible.

## What is next?

1. Complete createMessage function
2. Address TODOs
3. Write tests

PR #282, opened 2021-07-22
Remove password

Nirvikalpa108:
# Vision
IAM users with permanent credentials present a security vulnerability to the Guardian, however sometimes they are required. In this case, the credentials should be rotated regularly to minimise the security risk of them being compromised.

If credentials are not rotated regularly, then Security HQ will automatically disable them 🧚🏾‍♂️

SHQ will remove the access key and password for users with permanent credentials who have been notified twice that their credentials require rotation, but have taken no action. Once this has been done, a final email notification will be sent to inform the AWS account owners that this has been done.

## What does this change?
Remove the password of a permanent credential which is vulnerable and has not been remediated after 2 email notifications.

## How has this been tested?
It's been successfully tested locally with the password removal of a test user in the Deploy Tools account:

```
[info] s.IamRemovePassword - successfully deleted password for user xxxxxx in Deploy Tools account. DeleteLoginProfile Response: {}.
```

### Some background to this change:
This PR removes the password of a user whose disable deadline in dynamoDB is today.

The disable deadline is created as soon as a user is flagged as having vulnerable credentials. This deadline acts as a way for SHQ to know when to send a warning notification, final notification and eventually to disable the credential.

Currently the cadence for this is that as soon as a user is flagged as vulnerable, a warning email is sent, 2 weeks later a final email is sent and then 1 week from then, the credential is automatically disabled and an email is sent to say this has happened.

To note, the emails are sent to the email belonging to the AWS account.

### Previous PRs that tell the story of the Credentials Reaper feature
https://github.com/guardian/security-hq/pull/223
https://github.com/guardian/security-hq/pull/247
https://github.com/guardian/security-hq/pull/256
https://github.com/guardian/security-hq/pull/249


hq/app/schedule/IamRemovePassword.scala
```diff
@@ -0,0 +1,32 @@
...
+import scala.util.{Failure, Success}
+
+object IamRemovePassword extends Logging {
+
+  def deleteUserLoginProfile(client: AwsClient[AmazonIdentityManagementAsync], user: IamAuditUser)(implicit ec: ExecutionContext): Unit = {
+    val request = new DeleteLoginProfileRequest().withUserName(user.username)
+    awsToScala(client)(_.deleteLoginProfileAsync)(request).onComplete {
+      case Failure(exception) => logger.warn(s"failed to delete password for user ${user.username} in ${user.awsAccount} account.", exception)
+      case Success(result) => logger.info(s"successfully deleted password for user ${user.username} in ${user.awsAccount} account. DeleteLoginProfile Response: ${result.toString}.")
+    }
```
akash1810, line 21: We could add markers to these log lines to enable us to create visualisations in ELK - number of items deleted per account over time, for example.

akash1810 (approved):
Looks great! One minor comment about logging, but not a blocker!

img

hq/app/schedule/IamRemovePassword.scala
Nirvikalpa108: Nice idea 👍🏽. 

Hopefully this shouldn't happen too often so this could be something we consider for a different SHQ feature. Thanks for the tip.

PR #284, opened 2021-08-09
Credentials report refactoring

jfsoul:
## What does this change?
I've been looking at updating the way that the report status is determined in the IAM Credentials Report, so that it better matches our expectations regarding credentials rotation. _However_ there were a few small refactors of the logic around the report that I feel improve readability. On first reading I found some of these areas of the logic to be hard to understand, so the hope is that this helps.

There is a further refactor that'd be worth doing that's closer to the logic I actually need to change around outdated credentials. However I'm leaving that until #283 is merged, because I think there will be overlap. That may mean we implement the change itself and come back for the refactor.

## What is the value of this?
Less code, more readable code (some of that is subjective)

No

## Any additional notes?
No

### DIFF
--- a/hq/app/logic/CredentialsReportDisplay.scala
+++ b/hq/app/logic/CredentialsReportDisplay.scala
@@ -101,9 +101,9 @@ object CredentialsReportDisplay {
   def reportStatusSummary(report: CredentialReportDisplay): ReportSummary = {
     val reportStatusSummary = (report.humanUsers ++ report.machineUsers).map(_.reportStatus)
 
-    val warnings = reportStatusSummary.count(_.isInstanceOf[Amber.type])
-    val errors = reportStatusSummary.count(_.isInstanceOf[Red.type])
-    val others = reportStatusSummary.count(_.isInstanceOf[Blue.type])
+    val warnings = reportStatusSummary.count({ case Amber => true })
+    val errors = reportStatusSummary.count({ case Red => true })
+    val others = reportStatusSummary.count({ case Blue => true })
 
     ReportSummary(warnings, errors, others)
   }
--- a/hq/app/logic/CredentialsReportDisplay.scala
+++ b/hq/app/logic/CredentialsReportDisplay.scala
@@ -99,11 +99,13 @@ object CredentialsReportDisplay {
   }
 
   def reportStatusSummary(report: CredentialReportDisplay): ReportSummary = {
-    val reportStatusSummary = (report.humanUsers ++ report.machineUsers).map(_.reportStatus)
+    val reportStatusSummary = (report.humanUsers ++ report.machineUsers)
+      .groupBy(_.reportStatus)
+      .withDefaultValue(Seq.empty)
 
-    val warnings = reportStatusSummary.collect({ case Amber => true }).size
-    val errors = reportStatusSummary.collect({ case Red => true }).size
-    val others = reportStatusSummary.collect({ case Blue => true }).size
+    val warnings = reportStatusSummary(Amber).size
+    val errors = reportStatusSummary(Red).size
+    val others = reportStatusSummary(Blue).size
 
     ReportSummary(warnings, errors, others)
   }
--- a/hq/app/logic/CredentialsReportDisplay.scala
+++ b/hq/app/logic/CredentialsReportDisplay.scala
@@ -4,6 +4,7 @@ import logic.DateUtils.dayDiff
 import model._
 import org.joda.time.{DateTime, DateTimeZone, Days}
 import utils.attempt.FailedAttempt
+
 import java.net.URLEncoder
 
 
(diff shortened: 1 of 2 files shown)


hq/app/logic/CredentialsReportDisplay.scala
```diff
@@ -56,40 +57,34 @@ object CredentialsReportDisplay {
...
+          accessKey1Details(cred),
+          accessKey2Details(cred),
+          machineReportStatus(cred),
+          dayDiff(lastActivityDate(cred)),
+          stack = cred.stack,
+          tags = cred.tags
+        )
     }
+
+    CredentialReportDisplay(report.generatedAt, machineUsers, humanUsers)
```
jfsoul, line 87: the refactor of this method is perhaps the subjective one - but I think going from a `foldLeft` that conditionally appends an item to a list using `Option` and then copies the whole report is more complicated than 2 `filter`s and a single instantiation

hq/app/logic/CredentialsReportDisplay.scala
```diff
@@ -104,24 +99,20 @@ object CredentialsReportDisplay {
...
 
   def reportStatusSummary(report: CredentialReportDisplay): ReportSummary = {
-    val reportStatusSummary = report.humanUsers.map(_.reportStatus) ++ report.machineUsers.map(_.reportStatus)
+    val reportStatusSummary = (report.humanUsers ++ report.machineUsers).map(_.reportStatus)
 
-    val (warnings, errors, other) = reportStatusSummary.foldLeft(0,0,0) {
-      case ( (war, err, oth), Amber ) => (war+1, err, oth)
-      case ( (war, err, oth), Red ) => (war, err+1, oth)
-      case ( (war, err, oth), Blue ) => (war, err, oth+1)
-      case ( (war, err, oth), _ ) => (war, err, oth)
```
jfsoul, line 113: I found this a little hard to grep at first, but we're just counting statuses - so let's use `count`

hq/app/logic/CredentialsReportDisplay.scala
```diff
@@ -104,24 +99,20 @@ object CredentialsReportDisplay {
...
+    ReportSummary(warnings, errors, others)
   }
 
   def exposedKeysSummary(allReports: Map[AwsAccount, Either[FailedAttempt, List[ExposedIAMKeyDetail]]]): Map[AwsAccount, Boolean] = {
-    for {
-      (account, report) <- allReports
-    } yield (account, report match {
-        case Right(keys) if keys.nonEmpty => true
-        case _ => false
-    })
```
jfsoul, line 124: here we're doing an operation on the values of a map only - so `mapValues` is better suited

hq/app/logic/CredentialsReportDisplay.scala
jorgeazevedo: Yep! I can read your refactor pretty well, but I struggle with the original

hq/app/logic/CredentialsReportDisplay.scala
```diff
@@ -56,40 +57,34 @@ object CredentialsReportDisplay {
...
-            tags = cred.tags
-          ))
-        } else None
-
-      report.copy(
-        machineUsers = report.machineUsers ++ machineUser,
-        humanUsers = report.humanUsers ++ humanUser
-      )
+    val machineUsers = report.entries.filterNot(_.rootUser).collect {
+      case cred if !cred.passwordEnabled.getOrElse(false) =>
```
jorgeazevedo, line 75: I'm confused that this conditional is different from the humanUsers one. Should this read `!cred.passwordEnabled.contains(true)`?

hq/app/logic/CredentialsReportDisplay.scala
jorgeazevedo: Agreed, I'm having some difficulty reading the original to make sure the behaviour is the same. This should be easy to test no? Run the application locally and see if the iam creds are listed in the same way as before in the UI?

jfsoul: Yeah, it tests out fine locally!

Nirvikalpa108: I ran this locally and the IAM creds are indeed listed in the same way as in PROD 👍🏽

hq/app/logic/CredentialsReportDisplay.scala
jfsoul: Yes, good spot thank you. I think they're the same logically but I meant to update this one to match.

jorgeazevedo (approved):
Nice one!

hq/app/logic/CredentialsReportDisplay.scala
jorgeazevedo: thanks! that clears it up

Nirvikalpa108 (approved):

PR #285, opened 2021-08-10
Add credential status checks for outdated access keys

jfsoul:
## What does this change?
This adds extra criteria to IAM credential status checks to include access key age warnings. That is:

- For a human user (i.e. one with a password), if either access key is over 90 days old then the status will now be `Red`. Currently, the status is `Amber` if there is a key at all (because we discourage keys for human users).
- For a machine user (i.e. one without a password), if either access key is over 365 days old then the status will now be `Red`. Currently the status is `Green` or `Blue` depending on activity, `Red` is not used for machine keys.

image

## What is the value of this?
The motivation for this is to set expectations for key rotation more clearly in security HQ before we start sending alerts or automatically disabling keys as part of the work continuing in #283 

No

## Any additional notes?
This reuses the logic for `hasOutdated*Key` from the `IamFlaggedUsers` logic. I think this _could_ be refactored to put that logic in a shared place, rather than importing `IamFlaggedUsers` into the `CredentialsReportDisplay` but I think that's worth reflecting on when #283 is merged because it contains some changes relating to access key age in the models. **EDIT: this branch now includes that refactor done in #288**

There were also a few paths to take in terms of this implementation to deal with the fact we now have multiple reasons for why a human user might be in `Red` status - meaning we have to pass that information to the views so it can be surfaced in the UI. I decided to make `Red` a case class that carries reason(s) for the status as I think that is where the information sits most naturally - rather than as a separate entity in the `AccessKey`. It's not the cleanest because there are some niceties of using an `object` for statuses, so I'm open to thoughts.

### DIFF
--- a/hq/app/logging/Cloudwatch.scala
+++ b/hq/app/logging/Cloudwatch.scala
@@ -4,7 +4,7 @@ import com.amazonaws.services.cloudwatch.AmazonCloudWatchClientBuilder
 import com.amazonaws.services.cloudwatch.model.{Dimension, MetricDatum, PutMetricDataRequest, StandardUnit}
 import com.google.cloud.securitycenter.v1.Finding
 import logic.CredentialsReportDisplay
-import logic.CredentialsReportDisplay.reportStatusSummary
+import logic.CredentialsReportDisplay.{ReportSummary, reportStatusSummary, reportStatusSummaryWithoutOutdatedKeys}
 import model.{AwsAccount, CredentialReportDisplay, GcpFinding, GcpReport}
 import play.api.Logging
 import utils.attempt.FailedAttempt
@@ -43,7 +43,7 @@ object Cloudwatch extends Logging {
   def logMetricsForCredentialsReport(data: Map[AwsAccount, Either[FailedAttempt, CredentialReportDisplay]] ) : Unit = {
     data.toSeq.foreach {
       case (account: AwsAccount, Right(details: CredentialReportDisplay)) =>
-        val reportSummary: CredentialsReportDisplay.ReportSummary = reportStatusSummary(details)
+        val reportSummary: ReportSummary = reportStatusSummaryWithoutOutdatedKeys(details)
         putAwsMetric(account, DataType.iamCredentialsCritical, reportSummary.errors)
         putAwsMetric(account, DataType.iamCredentialsWarning, reportSummary.warnings)
         putAwsMetric(account, DataType.iamCredentialsTotal, reportSummary.errors + reportSummary.warnings)
--- a/hq/app/logic/CredentialsReportDisplay.scala
+++ b/hq/app/logic/CredentialsReportDisplay.scala
@@ -3,7 +3,6 @@ package logic
 import logic.DateUtils.dayDiff
 import model._
 import org.joda.time.{DateTime, DateTimeZone, Days}
-import schedule.IamFlaggedUsers
 import utils.attempt.FailedAttempt
 
 import java.net.URLEncoder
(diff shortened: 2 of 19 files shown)


jorgeazevedo:
So, we're sending metrics to AWS based on the ReportSummary errors and warnings https://github.com/guardian/security-hq/blob/31b69c9de3d0e4deca3a10d114903ecc6ce3c9a9/hq/app/logging/Cloudwatch.scala#L47

If I read this right, the new Red statuses show up in the ReportSummary as errors, meaning they're going to be added to our "iam/critical" cloudwatch metric, which will  show up in the vulnerabilities dashboard. Right? 

It'd be good to decouple these two things somehow, so we can roll them out separately. It's going to be quite shocking to our stakeholders if our numbers just shoot up one day. Can we filter them out somehow from the cloudwatch object method I linked?

jfsoul:
Good spot Jorge, I hadn't considered this, but maybe it's the right thing to do anyway? According to the dashboard these are critical vulnerabilities I think? (number 3):

image

I was a bit torn on whether these should be Red or Amber alerts, but actually the above is convincing me we should be using Red and surfacing them so teams and stakeholders see the priority of regular rotation. Perhaps it means we'll need to think about how we can provide ways for teams to do key rotation more easily as well (e.g. for common use cases like fastly logging)

jorgeazevedo:
Yeah I definitely agree critical is the right category for it and we should display it in the dashboard. My suggestion is to add them to security hq first, and to grafana afterwards.

This will give us an opportunity to see what the numbers are and see how we want to communicate it to stakeholders. We might remediate some low hanging fruit if there is some, we might want to change the dashboard itself to better differentiate IAM categories so it's more obvious that this is new data and not a regression. Right now the split is `iam/credentials` and `iam/keys` (which turned out we don't actually use). We might want to split `iam/credentials` into 2 and have it map to the bullet point in your screenshot, for instance.

jfsoul:
> My suggestion is to add them to security hq first, and to grafana afterwards.

Ah ok, I see now - I'll take a look at that.

jorgeazevedo:
@jfsoul ace, much appreciated!

Nirvikalpa108:
Great work Jon! 🎉 

## Follow up work - problem description
A follow up piece of work could be to improve the communication of the outdated access key rule, given it is triggered on disabled access keys as well as enabled ones. I had a conversation with a colleague in the apps team recently about this as it had caused confusion, which is very understandable. 

Here is a machine user that has been flagged because access key 1, which is disabled, is nevertheless flagged because it was last rotated over 365 days ago:
Screenshot 2021-08-12 at 14 47 02

## Proposed solution
We could make it clearer by adding some text in the pop up that is generated when a user clicks on the question mark in the round circle icon? Perhaps this could be added in a subsequent PR.

Here is the pop-up. Maybe we could add the outdated access keys rule to this in a subsequent PR. We could note that disabled access keys which surpass the rotated cadence rule should either be rotated or deleted.
Screenshot 2021-08-12 at 14 52 35

jfsoul:
That's a great point @Nirvikalpa108 - I meant to add to the helper descriptions in this PR but it slipped my mind. I'll have a look and see whether it makes sense to include here or in another PR.

Nirvikalpa108:
There's a tiny bug here where this user should have a red flag because of the outdated key. Let's try and fix this in this PR.

Screenshot 2021-08-12 at 17 20 28

Nirvikalpa108 (approved):
Thank you so much for doing this work @jfsoul! 💯 🎉 👍🏽
It's amazing to have the IAM dashboard corrected. It will be really great for when we speak with engineers about the changes they need to make before we roll out the credentials reaper.

jfsoul:
Great spot with that bug, should be addressed with a tweak to the twirl syntax I used: https://github.com/guardian/security-hq/pull/285/commits/349b3e203c2955976b274ef6044e519ab0085091

For the updates to the help modal dialog, I'm going to add that to a separate PR for easier review.

jfsoul:
This branch also now includes the refactor done subsequently and approved in #288

PR #287, opened 2021-08-12
Small documentation fixes

jfsoul:
## What does this change?
- Fixes typo on docs main page
- Removes duplicate section from vulnerability management
- Adjust the image links so that they'll work in the web app itself as well as when reading the markdown on github. 
 
N.B. The solution for the images isn't ideal, because it'll always refer to the main branch, but after wrangling with other options (serving images as assets from the app in a location that relative URLs would work) for too long, this seems liked the sensible compromise for now.

image


jorgeazevedo (approved):
Nice one Jon! Can't believe I never noticed the security hub duplicated section

PR #288, opened 2021-08-12
Centralise access key logic

jfsoul:
## What does this change?
This is a small change to centralise some access key rotation logic, which is now being shared by more than one report/job.

I also spotted a warning about duplicate code in one of the files I was changing, so the second commit adds a smaller conversion method to the model that avoids duplicated class creation code in the logic.


Nirvikalpa108 (approved):
Looks great, thanks Jon!

PR #290, opened 2021-08-13
Update credential help dialog to include rotation errors

jfsoul:
## What does this change?
This updates the credentials helper dialog to include warnings about the newly-flagged "outdated key" statuses in the IAM creds report. It differs a bit from human to machine.

## What is the value of this?
Help for people trying to understand why their IAM users are being flagged.

No

## Any additional notes?
No

### DIFF
--- a/hq/app/views/fragments/credentialsHelp.scala.html
+++ b/hq/app/views/fragments/credentialsHelp.scala.html
@@ -1,12 +1,3 @@
-@newUser = {
-    <li class="collection-item avatar">
-        <i class="material-icons circle large-icon white purple-text">new_releases</i>
-        <span class="title">New user</span>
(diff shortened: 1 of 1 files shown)


jorgeazevedo (approved):
Nice one, I think the explanations read very well!

hq/app/views/fragments/credentialsHelp.scala.html
```diff
@@ -1,12 +1,3 @@
-@newUser = {
```
jorgeazevedo, line 1: Did you mean to delete this?

jfsoul: Yeah, it doesn't seem to be used at all

PR #291, opened 2021-08-16
Improve logging for CloudWatch metrics service

jacobwinch:
## What does this change?

This PR improves the logging related to CloudWatch metrics:

1. When individual metrics are pushed successfully, we now log at `debug` (rather than `info`) level
2. When metrics cannot be pushed due to problems with data collection, we now log more details about what has gone wrong

## What is the value of this?

1. Reduces [noise](https://logs.gutools.co.uk/goto/51d80808f26ebdf629c283e40c643be5), which should make it easier to analyse the logs
2. Helps us to debug in scenarios where [no metrics are being pushed (which seems to be the case at the time of writing!)](https://logs.gutools.co.uk/goto/4d189c03575774fcc3d9c44a6115a8f7)

No.

### DIFF
--- a/hq/app/logging/Cloudwatch.scala
+++ b/hq/app/logging/Cloudwatch.scala
@@ -75,8 +75,8 @@ object Cloudwatch extends Logging {
     val request = new PutMetricDataRequest().withNamespace(namespace).withMetricData(datum)
 
     Try(cloudwatchClient.putMetricData(request)) match {
-      case Success(response) => logger.info(s"putMetric: ${datum}")
-      case Failure(e) => logger.error(s"putMetric failure: ${datum}", e)
+      case Success(_) => logger.debug(s"putMetric success: $datum")
+      case Failure(e) => logger.error(s"putMetric failure: $datum", e)
     }
   }
 }
--- a/hq/app/services/MetricService.scala
+++ b/hq/app/services/MetricService.scala
@@ -18,9 +18,11 @@ class MetricService(
     cacheService: CacheService
   )(implicit ec: ExecutionContext) extends Logging {
 
-  def dataMissingFrom[T](list: List[Map[AwsAccount, Either[FailedAttempt, T]]]): Boolean = {
-    list.exists { dataMap =>
-      dataMap.toSeq.exists(_._2.isLeft)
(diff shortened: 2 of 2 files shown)


hq/app/logging/Cloudwatch.scala
```diff
@@ -75,8 +75,8 @@ object Cloudwatch extends Logging {
     val request = new PutMetricDataRequest().withNamespace(namespace).withMetricData(datum)
 
     Try(cloudwatchClient.putMetricData(request)) match {
-      case Success(response) => logger.info(s"putMetric: ${datum}")
-      case Failure(e) => logger.error(s"putMetric failure: ${datum}", e)
+      case Success(_) => logger.debug(s"putMetric success: $datum")
```
jacobwinch, line 78: This is the change requested in https://trello.com/c/y2SPZYvz/799-%F0%9F%9A%A8-security-hq-delete-noisy-log-lines - I stumbled upon the other problem whilst figuring out what to update!

hq/app/services/MetricService.scala
```diff
@@ -52,19 +54,21 @@ class MetricService(
     val allPublicBuckets = cacheService.getAllPublicBuckets
     val allCredentials = cacheService.getAllCredentials
 
-    if(dataMissingFrom(List(allSgs, allExposedKeys, allPublicBuckets, allCredentials))) {
-      logger.info("At least part of the data is missing - skipping cloudwatch update!")
-      return
```
jacobwinch, line 57: I found this a little surprising, so I've refactored it away

hq/app/services/MetricService.scala
```diff
@@ -74,7 +78,6 @@ class MetricService(
       else Duration.Zero
 
     val cloudwatchSubscription = Observable.interval(initialDelay, 6.hours).subscribe { _ =>
-      logger.info("Posting new metrics to cloudwatch")
```
jacobwinch, line 77: Logging this here isn't that helpful, because we get this log line even if we skip due to problems with the data

jacobwinch:
> 2\. Helps us to debug in scenarios where [no metrics are being pushed (which seems to be the case at the time of writing!)](https://logs.gutools.co.uk/goto/4d189c03575774fcc3d9c44a6115a8f7)

I think this might be due to the current stack set problems with the Media Service account:

image

Hopefully this PR will confirm that and make it easier to discover these problems in the future.

jfsoul (approved):
Nice!

PR #293, opened 2021-08-31
Send notification when Credentials Reaper disables an access key or removes the password of a user

Nirvikalpa108:
## What is this change?
This PR sends a notification email when Security HQ disables permanent IAM users (by removing the user's password and/or access key(s)). 

Here is an example of the email:
Screenshot 2021-09-01 at 08 35 18

## Why is it valuable?
This aims to communicate to AWS account holders that there has been a change in the status of the IAM users in their account and how they can remediate this if required. 

## What's next?
By the time this email is sent, the AWS account holder should have received three emails warning them about the impeding automatic disablement and telling them how to avoid this.  In the case that SHQ fails to disable the user, it's important that DevX take responsibility for remediating this vulnerable user as the AWS account holder has not taken any action. 

I've added TODOs in the code to trigger a cloudwatch alarmso that DevX can debug this and possible manually disable the user.

### DIFF
--- a/hq/app/schedule/IamMessages.scala
+++ b/hq/app/schedule/IamMessages.scala
@@ -1,5 +1,8 @@
 package schedule
 
+import aws.AwsClient
+import com.amazonaws.services.identitymanagement.AmazonIdentityManagementAsync
+import config.Config.{iamHumanUserRotationCadence, iamMachineUserRotationCadence}
 import model.{AwsAccount, VulnerableUser}
 import org.joda.time.DateTime
 import org.joda.time.format.DateTimeFormat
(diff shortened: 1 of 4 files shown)


hq/app/schedule/IamMessages.scala
```diff
@@ -10,6 +10,15 @@ object IamMessages {
 
   def warningSubject(account: AwsAccount): String = s"Action ${account.name}: Insecure credentials"
   def finalSubject(account: AwsAccount): String = s"Action ${account.name}: Deactivating credentials tomorrow"
+  def disabledUsersSubject(account: AwsAccount): String = s"AWS IAM User(s) DISABLED in ${account.name} Account"
+
+  def disabledUsersMessage(users: Seq[VulnerableUser]): String = {
+    s"""
+       |The following Permanent IAM user(s) have been disabled today: ${users.map(_.username).mkString(",")}.
+       |Please check Security HQ (https://security-hq.gutools.co.uk/iam) to remediate this.
```
jfsoul, line 18: (Optional) How about "Please check Security HQ to review the IAM users in your account" instead? As we're not necessarily expecting anything to be remediated are we?

I guess we could also include instructions like: "If you still require the disabled user, you should add new access credentials and ensure their access keys are regularly rotated going forwards, or MFA is added for human users. If you no longer require the disabled user, they should be deleted."

jfsoul (approved):
Looks great! It'd be marginally preferable to avoid emailing about any credentials that are failed to be removed, but I like the simplicity of this and if we are dealing with that edge case with an alarm to DevX then I think it's ok.

hq/app/schedule/IamMessages.scala
Nirvikalpa108: Great suggestion, thank you!

PR #294, opened 2021-09-02
Layout changes to accommodate an additional IAM related scheduled job

jfsoul:
## What does this change?
This PR moves some of the IAM job code into a specific `vulnerable` package in order to distinguish it from an expected second scheduled job which will be focussed on `unrecognised` human users. It also starts to separate and flesh out some of the shared code/locations between these jobs (although this is only partially done). Namely:
- Adds a `type` of notification to the dynamo schema (`VulnerableCredential` or `UnrecognisedHumanUser` at the moment)
- Creates some placeholder texts for an `Unrecognised` email notification

There's likely more work required in this area so that the 2 jobs can coexist cleanly, but I didn't want this PR to become too large before continuing.

image

## What is the value of this?
Lays the foundation for further work on a second scheduled job

No

## Any additional notes?
I'd like to test the dynamo code changes, but need to work out how best to do that.

### DIFF
--- a/hq/app/model/models.scala
+++ b/hq/app/model/models.scala
@@ -333,16 +333,21 @@ case class VulnerableAccessKey(
   humanUser: Boolean
 )
 
-sealed trait IamAuditNotificationType {def name: String}
-object Warning extends IamAuditNotificationType {val name = "Warning"}
-object Final extends IamAuditNotificationType {val name = "Final"}
+sealed trait IamAuditNotificationType { def name: String }
+object InsecureCredential extends IamAuditNotificationType { val name = "insecureCredential" }
+object UnrecognisedHumanUser extends IamAuditNotificationType { val name = "unrecognisedHumanUser" }
+object IamAuditNotificationType {
+  def fromName(name: String): IamAuditNotificationType =
+    Seq(InsecureCredential, UnrecognisedHumanUser)
+      .find(name == _.name)
+      .getOrElse(InsecureCredential)
+}
 
-case class IamAuditAlert(dateNotificationSent: DateTime, disableDeadline: DateTime)
+case class IamAuditAlert(`type`: IamAuditNotificationType, dateNotificationSent: DateTime, disableDeadline: DateTime)
 object IamAuditAlert {
-  implicit val jodaDateWrites: Writes[DateTime] = new Writes[DateTime] {
-    def writes(d: DateTime): JsValue = JsString(d.toString())
-  }
-  implicit val iamAuditAlerts = Json.writes[IamAuditAlert]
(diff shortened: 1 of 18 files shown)


hq/app/model/models.scala
```diff
@@ -333,16 +333,21 @@ case class VulnerableAccessKey(
   humanUser: Boolean
 )
 
-sealed trait IamAuditNotificationType {def name: String}
-object Warning extends IamAuditNotificationType {val name = "Warning"}
-object Final extends IamAuditNotificationType {val name = "Final"}
```
Nirvikalpa108, line 338: I hadn't realised that these `Warning` and `Final` objects were unused, so good to get rid of them!

hq/app/model/models.scala
```diff
@@ -333,16 +333,21 @@ case class VulnerableAccessKey(
...
+}
 
-case class IamAuditAlert(dateNotificationSent: DateTime, disableDeadline: DateTime)
+case class IamAuditAlert(`type`: IamAuditNotificationType, dateNotificationSent: DateTime, disableDeadline: DateTime)
 object IamAuditAlert {
-  implicit val jodaDateWrites: Writes[DateTime] = new Writes[DateTime] {
-    def writes(d: DateTime): JsValue = JsString(d.toString())
-  }
-  implicit val iamAuditAlerts = Json.writes[IamAuditAlert]
+  implicit val jodaDateWrites: Writes[DateTime] = (d: DateTime) => JsString(d.toString())
```
Nirvikalpa108, line 348: Nice to see a terser implementation 👍🏽

hq/app/schedule/Dynamo.scala
```diff
@@ -42,6 +42,7 @@ class Dynamo(client: AmazonDynamoDB, tableName: Option[String]) extends Attribut
       val alerts = r("alerts").getL.asScala.map { a =>
         val alertMap = a.getM.asScala
         IamAuditAlert(
+          IamAuditNotificationType.fromName(alertMap("type").getS),
```
Nirvikalpa108, line 45: I wonder what the dynamoDB table will start to look like once we turn on the IAM job and this new job? Will it be too busy? Will it be easy to understand? Do we need a CODE table as well as a PROD one for testing things? This PR looks great, so let's merge this, but perhaps we can have this discussion later.

Nirvikalpa108 (approved):

hq/app/schedule/Dynamo.scala
jfsoul: Yeah, happy to talk about that soon. CODE table sounds useful on the face of it but we don't really have the concept of CODE IAM users which makes it harder to isolate a CODE job/table. Which I suppose is why you didn't set up a CODE table in the first place. Something to think about!

PR #295, opened 2021-09-06
Add ignored tests to enable sending real test notifications

Nirvikalpa108:
# What does this change?
This adds a (disabled) test suite to the app, which can be used for testing notification emails locally. Sometimes it's useful to send notification emails to check email formatting when making changes - this enables that if you make the test discoverable and remove the `ignore`s from the appropriate tests.

The `@DoNotDiscover` annotation turned out to be necessary because we're creating a fake Play application, which requires config - which is not stubbed out in teamcity (this is the first time in this project this has been tried). An alternative solution would be to provide some values for the required config in teamcity/DEVINFRA so that an application can be initialised. I believe this would be necessary if we wanted to start testing controllers using Play's [suggested methods](https://www.playframework.com/documentation/2.8.x/ScalaTestingWithScalaTest), for example.


hq/test/schedule/SendTestNotifications.scala
```diff
@@ -0,0 +1,92 @@
...
+class SendTestNotifications extends FreeSpec with OneAppPerSuiteWithComponents with Matchers with AttemptValues {
+
+  override def components: BuiltInComponents = new BuiltInComponentsFromContext(context) with NoHttpFiltersComponents {
+      lazy val router: Router = Router.empty
+  }
+
+  sealed trait NotificationType
+  object Warning extends NotificationType
+  object Final extends NotificationType
+  object Disabled extends NotificationType
```
jfsoul, line 37: These seem like they'd be useful in the the app code itself!

jfsoul (approved):
As discussed with @Nirvikalpa108 we're happy for this change to be merged - we've each reviewed the other's part of the PR while pairing 👍

PR #296, opened 2021-09-06
Add basic dynamo tests

jfsoul:
## What does this change?
Adds some basic Dynamo tests using the[ DynamoDB local](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.html) jar provided by Amazon.  This jar (or rather the plugin that downloads it) provides a convenient way to run tests against an in-memory instance of DynamoDB. Unfortunately, it works by downloading the jar when the tests are run (unless it already exists) rather than as a classic project dependency. The [plugin project](https://github.com/localytics/sbt-dynamodb) also seems to be archived but it's quite simple so we could fork or recreate it. Other projects at the Guardian also already depend on it. On balance I would be willing to take the convenience over these downsides.

## What is the value of this?
More confidence making changes to dynamo code (a little more)

No

## Any additional notes?
I couldn't really find a direct alternative to the sbt plugin, but I suppose the options would be:
- Add the jar to VCS directly
- Don't test dynamo this way

### DIFF
--- a/hq/test/schedule/DynamoTest.scala
+++ b/hq/test/schedule/DynamoTest.scala
@@ -1,0 +1,114 @@
+package schedule
+
+import com.amazonaws.auth.{AWSStaticCredentialsProvider, BasicAWSCredentials}
+import com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder
+import com.amazonaws.services.dynamodbv2.model._
+import model._
+import org.joda.time.DateTime
+import org.scalatest.{BeforeAndAfterAll, FreeSpec, Matchers}
+
+import scala.collection.JavaConverters._
+
+
+class DynamoTest extends FreeSpec with BeforeAndAfterAll with Matchers with AttributeValues {
+
+  private val client = {
+    val conf = new EndpointConfiguration("http://localhost:8000", config.Config.region.name)
+    AmazonDynamoDBClientBuilder.standard
+      .withEndpointConfiguration(conf)
+      .withCredentials(new AWSStaticCredentialsProvider(new BasicAWSCredentials("dummy", "credentials")))
+      .build
+  }
+
+  private val tableName = "security-hq-iam-TEST"
+  private val tableAttributes = Seq((Symbol("id"), ScalarAttributeType.S))
+  private val keySchemaAttributes = tableAttributes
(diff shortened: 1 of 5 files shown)


hq/test/schedule/DynamoTest.scala
```diff
@@ -0,0 +1,114 @@
...
+      .withCredentials(new AWSStaticCredentialsProvider(new BasicAWSCredentials("dummy", "credentials")))
+      .build
+  }
+
+  private val tableName = "security-hq-iam-TEST"
+  private val tableAttributes = Seq((Symbol("id"), ScalarAttributeType.S))
+  private val keySchemaAttributes = tableAttributes
+  private val arbitraryThroughputThatIsIgnoredByDynamoDBLocal = new ProvisionedThroughput(1L, 1L)
+
+  override def beforeAll () {
```
jorgeazevedo, line 29: Is this boilerplate?

jorgeazevedo (approved):

jorgeazevedo:
This is distributed in a bit of a weird way, but I don't think it's particularly deadly. It does feel like it's been deprecated, so if the jar isn't that big we could consider committing it, in the off chance they pull the plug on the s3 bucket or something. I don't feel very strongly about it though

hq/test/schedule/DynamoTest.scala
jfsoul: If you mean what's `beforeAll` for - it's the way you tell scalatest to run some side effects before any of your tests. There's also `afterAll` and `beforeEach`/`afterEach` for similar.

jorgeazevedo: Oh sorry, I didn't really make it obvious what I was referring to. I meant the implementation of `beforeAll`. Specifically I was wondering if `attributeDefinitions` and `keySchema` were boilerplate picked up from documentation, or if it was your own implementation

jfsoul: Ah yes, it is quite unpleasant. Mostly taken from https://github.com/scanamo/scanamo/blob/master/testkit/src/main/scala/org/scanamo/LocalDynamoDB.scala

jfsoul:
@jorgeazevedo I think the DynamoDB Local jar itself is pretty safe, as it's amazon-provided - I think they'll be fairly committed to maintaining it. It's this sbt plugin that integrates the download of the jar that feels less solid. I honestly think we should fork it and maintain ourselves if we think it's useful as there's a few projects that depend on it.

jorgeazevedo:
Ohh  see, thanks for clarifying! Yes, agreed.

PR #299, opened 2021-09-08
Don't flag disabled keys in credentials reaper

jfsoul:
## What does this change?
This is a slight adjustment to the logic for determining flagged IAM users in the credentials reaper job. At the moment, even disabled keys will be flagged, but because the action on those keys is only to disable them, we would end up with circular logic until we actually implement key _deletion_.  We've decided to focus on the rollout of the reaper over deletion for now, so this change allows us to move ahead with a working system.

For now, we've duplicated some of the `isOutdated*` logic because we want to continue to flag in the UI that disabled keys should be removed.

## What is the value of this?
Decouples the rollout of the reaper from implementing key deletion. We'd like to have confidence running the current version in production before adding deletion.

No

## Any additional notes?
Paired on with @Nirvikalpa108 and @jorgeazevedo, thanks!


Nirvikalpa108 (approved):
Great 👍🏽

PR #300, opened 2021-09-08
Failure case for disable access key request 

Nirvikalpa108:
This PR adds a failure case to the disable access key request so that a cloudwatch alarm can be triggered. 

This is valuable, because in the case that SHQ is unable to disable a vulnerable access key, the DevX team will be notified and they can take appropriate to action to understand what caused the failure, fix it or get in touch with the AWS account owner to disable the access key manually if required.


jfsoul (approved):
⚡

PR #301, opened 2021-09-09
Update references to outdated methods for the dashboard

jfsoul:
## What does this change?
Noticed that #299 missed a step to start using the newly duplicated methods for checking outdated status of keys. This fixes the dashboard to use the _old_ logic - that is to consider disabled keys as vulnerable.


Nirvikalpa108 (approved):
Great that you spotted this! Sorry I missed this in review!

PR #302, opened 2021-09-10
Add cloudwatch metrics and alerts to Creds Reaper

jorgeazevedo:
## What does this change?

This adds two cloudwatch metrics to track the operation of creds reaper. One tracks removal of passwords, the other disabling of keys. It uses a Dimension to track if the operation was successful or not

CloudWatch_Management_Console

It also adds two cloudwatch alarms to email us when a failure happens. One for each type.

Inbox__15__-_jorge_azevedo_guardian_co_uk_-_guardian_co_uk_Mail

The alarm description contains a short URL to security hq's logs, which should contain the messages logged just before the `putMetric` methods are called.
https://logs.gutools.co.uk/s/devx/goto/f9915a6e4e94a000732d67026cea91be

ALARM___Security_HQ_failed_to_remove_a_vulnerable_password__in_EU__Ireland__-_jorge_azevedo_guardian_co_uk_-_guardian_co_uk_Mail

## What is the value of this?

Helps us track the execution of creds reaper and be alerted when it fails.

Yes! This requires an SNS topic passed as a parameter. I was hoping we had one already in the security account, but I think we'll have to create one.

## Any additional notes?

Question: are SNS topics cloudformed? Are they just done in the UI?

### DIFF
--- a/cloudformation/security-hq.template.yaml
+++ b/cloudformation/security-hq.template.yaml
@@ -396,10 +396,10 @@ Resources:
         - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:${SNSTopicForAlerts}
       AlarmName: !Sub Security HQ failed to remove a vulnerable password
       AlarmDescription: !Sub The credentials reaper feature of Security HQ logs either success or failure to cloudwatch, and this alarm lets us know when it logs a failure. Check the application logs for more details https://logs.gutools.co.uk/s/devx/goto/f9915a6e4e94a000732d67026cea91be
-      MetricName: IamRemovePasswordExecution
+      MetricName: IamRemovePassword
       Namespace: SecurityHQ
       Dimensions:
-        - Name: Status
+        - Name: ReaperExecutionStatus
           Value: Failure
       ComparisonOperator: GreaterThanOrEqualToThreshold
       Threshold: 1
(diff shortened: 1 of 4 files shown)


cloudformation/security-hq.template.yaml
```diff
@@ -386,6 +389,44 @@ Resources:
...
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      AlarmActions:
+        - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:${SNSTopicForAlerts}
+      AlarmName: !Sub Security HQ failed to remove a vulnerable password
+      AlarmDescription: !Sub The credentials reaper feature of Security HQ logs either success or failure to cloudwatch, and this alarm lets us know when it logs a failure. Check the application logs for more details https://logs.gutools.co.uk/s/devx/goto/f9915a6e4e94a000732d67026cea91be
+      MetricName: IamRemovePasswordExecution
+      Namespace: SecurityHQ
+      Dimensions:
+        - Name: Status
```
jfsoul, line 402: Could we call it something more specific than `Status`? It strikes me looking at your screenshots that the way cloudwatch drills down into metric dimensions makes `Status` a bit too generic because you end up with `SecurityHQ -> Status` which sounds like status of the whole app(?)

cloudformation/security-hq.template.yaml
```diff
@@ -54,6 +54,9 @@ Parameters:
   TopicARN:
     Type: String
     Description: ARN for Anghammarad's SNS topic
+  SNSTopicForAlerts:
```
jfsoul, line 57: I think we can cloudform this but perhaps we'll end up supplying the topic subscriber email address as a parameter

hq/app/logging/Cloudwatch.scala
```diff
@@ -69,6 +69,22 @@ object Cloudwatch extends Logging {
     putMetric("SecurityHQ", "Vulnerabilities", Seq(("GcpProject", project),("DataType", dataType.toString)), value)
   }
 
+  def putIamRemovePasswordSuccessMetric(): Unit = {
+    putMetric("SecurityHQ", "IamRemovePasswordExecution", Seq(("Status", "Success")), 1)
```
jfsoul, line 73: Minor thing but might be nice to put some of these strings in `val`s or an enumeration/type.

cloudformation/security-hq.template.yaml
jorgeazevedo: Indeed! It's actually easier than I expected

jfsoul (approved):

PR #306, opened 2021-09-14
Create new scheduled job for unrecognised staff

Nirvikalpa108:
This PR creates a new quartz scheduler job which intends to remove permanent credentials of Guardian staff members who have left the organisation.


jorgeazevedo (approved):
Having these jobs be about vulnerable/unrecognised iam users works really well I think. It's a pretty intuitive distinction

PR #307, opened 2021-09-14
Create a dummy Janus user

Nirvikalpa108:
This PR creates a dummy Janus user using Janus's config library for 2 reasons.

First, so that the unrecognised former staff feature can use the library's built in methods for parsing a file as this feature intends to read a file from S3 which holds all of the users in Janus. The intention for this is to ensure that no IAM permanent credentials are assigned to anyone who is NOT included in Janus.

Second, the dummy user is being created so that we can concentrate on writing the main functionality of the feature comparing the Janus user name to names in IAM tags) and write the S3 file reading afterwards.


jorgeazevedo (approved):
I think this approach makes sense!

PR #308, opened 2021-09-15
Vulnerable IAM user job test reorganisation

jfsoul:
## What does this change?
This PR reorganises some of the tests of the vulnerable scheduled job. In particular we have:
- Split out tests from IamNotificationsTest into classes and files that match the source code structure
- Adds `private` to internal methods in the source code to make it clearer what the surface area for testing should be. This includes some methods that were used externally, but only in tests
- Renames the `Dynamo` class so we can introduce a mock version of it i.e. `DynamoAlertService` is the abstract trait, implemented by `AwsDynamoAlertService` and `MockDynamoAlertService` (the latter being useful for tests)
- Adds one test to the `IamUsersToDisableTest` class. This is to cover the case where the deadline for disabling has passed. The test is written to match the current behaviour, but I think the desired behaviour might be different (i.e. we should attempt disable if past the deadline)

## What is the value of this?
The aim of these changes is ultimately to make it clearer where we are missing test coverage (if anywhere), so that adding tests will be easier in later PRs (we've only added one actual test case here). We believe having good test coverage of this code will better help to document its intended behaviour, give us more confidence when making changes. We may also see some benefit in the source code if we make it more testable.

No

## Any additional notes?
No


hq/test/schedule/IamNotificationsTest.scala
```diff
@@ -1,212 +1,15 @@
...
-  val outdatedUser1 = VulnerableUser("lesleyKnope", humanUser = true, tags = List())
-  val outdatedUser2 = VulnerableUser("ronSwanson", humanUser = true, tags = List())
-  val outdatedUser3 = VulnerableUser("tomHaverford", humanUser = true, tags = List())
-
-  val noMfaUser1 = VulnerableUser("april", humanUser = true, tags = List())
-  val noMfaUser2 = VulnerableUser("andy", humanUser = true, tags = List())
-  val noMfaUser3 = VulnerableUser("diane", humanUser = true, tags = List())
-
-
-  "getNotificationTargetGroups" - {
```
jfsoul, line 23: All these tests were moved elsewhere, rather than deleted

hq/test/schedule/IamUsersToDisableTest.scala
```diff
@@ -0,0 +1,96 @@
...
+        List(Stack("test-stack")),
+        Seq(
+          VulnerableUser(
+            username = username,
+            humanUser = true,
+            tags = List.empty,
+            disableDeadline = None
+  )))))
+
+  "usersToDisable" - {
```
jfsoul, line 26: These tests are more complicated than the ones they replace - previously we were only testing one part of the class (an internal method). We could consider if there is any way to simplify the logic that helps with testability, but that felt outside the scope of this PR

hq/app/schedule/IamNotifications.scala
```diff
@@ -21,16 +21,16 @@ object IamNotifications extends Logging {
...
       IamAuditUser(
-        Dynamo.createId(account, user.username), account.name, user.username,
+        DynamoAlerts.createId(account, user.username), account.name, user.username,
         List(IamAuditAlert(VulnerableCredential, DateTime.now, createDeadlineIfMissing(user.disableDeadline)))
       )
     }
   }
 
-  def createWarningAndFinalNotification(tg: IAMAlertTargetGroup, awsAccount: AwsAccount, users: Seq[IamAuditUser]): IamNotification = {
+  private def createWarningAndFinalNotification(tg: IAMAlertTargetGroup, awsAccount: AwsAccount, users: Seq[IamAuditUser]): IamNotification = {
```
Nirvikalpa108, line 33: We test this function in `test/schedule/SendTestNotifications.scala`.

```suggestion
  def createWarningAndFinalNotification(tg: IAMAlertTargetGroup, awsAccount: AwsAccount, users: Seq[IamAuditUser]): IamNotification = {
```

Nirvikalpa108 (approved):

hq/app/schedule/IamNotifications.scala
jfsoul: Good spot, I noticed too after the TC build failed. It came from the rebase of main, so I did the same thing as your suggestion for now.

PR #309, opened 2021-09-15
Find IAM users who are not Janus users

Nirvikalpa108:
## What does this change?
This adds logic to find all the latest human users who are not janus users. This means that these are the people with IAM permanent credentials who have left the Guardian. The next PR will be to disable these users' permanent credentials.

### DIFF
--- a/hq/test/schedule/IamUnrecognisedUserTest.scala
+++ b/hq/test/schedule/IamUnrecognisedUserTest.scala
@@ -1,0 +1,46 @@
+package schedule
+
+import com.gu.janus.model.{ACL, AwsAccount, JanusData, SupportACL}
+import model.{AccessKey, CredentialReportDisplay, Green, HumanUser, NoKey, Tag, AwsAccount => Account}
+import org.joda.time.{DateTime, Seconds}
+import org.scalatest.{FreeSpec, Matchers}
+import schedule.unrecognised.IamUnrecognisedUsers.{filterUnrecognisedIamUsers, getHumanUsers, getJanusUsernames}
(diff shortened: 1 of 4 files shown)


hq/test/schedule/IamUnrecognisedUserTest.scala
```diff
@@ -0,0 +1,46 @@
...
+      getJanusUsernames(dummyJanusData) shouldEqual List("firstName.secondName")
+    }
+    "get human users from credentials report" in {
+      val credsReport: Map[Account, Either[FailedAttempt, CredentialReportDisplay]] =
+        Map(Account("id", "name", "roleArn", "accountNumber") -> Right(credsReportDisplay))
+      val result: List[HumanUser] = List(humanUser1, humanUser2, humanUser3, humanUser4)
+      getHumanUsers(credsReport) shouldEqual result
+    }
+    "get unrecognised human users" in {
+      val permanentIamUsers: List[HumanUser] = List(humanUser1, humanUser2, humanUser3, humanUser4)
```
jorgeazevedo, line 41: Consider adding a user without tags to this list, just to cover that code path

jorgeazevedo (approved):
I think this is really solid!

hq/test/schedule/IamUnrecognisedUserTest.scala
Nirvikalpa108: Great idea! Have created a ticket so I remember to add this next week! Thanks! https://trello.com/c/rajCb8Np

PR #310, opened 2021-09-15
Disable keys and logins in unrecognised users job

jfsoul:
## What does this change?
This adds password removal and access key disablement to the new unrecognised users job. It also sends a disablement email notification. 

## What is the value of this?
Fleshes out more of the functionality of the job before we hook it up to real data and do some end to end testing.

No

## Any additional notes?
No

### DIFF
--- a/hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
+++ b/hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
@@ -53,7 +53,7 @@ class IamUnrecognisedUserJob(cacheService: CacheService, snsClient: AmazonSNSAsy
         disableAccessKeys(account, unrecognisedIamUsers, iamClients)
         removePasswords(account, unrecognisedIamUsers, iamClients)
 
-        if(unrecognisedIamUsers.nonEmpty) {
+        if (unrecognisedIamUsers.nonEmpty) {
           val message = notification(
             disabledUsersSubject(account),
             disabledUsersMessage(unrecognisedIamUsers),
@@ -67,5 +67,6 @@ class IamUnrecognisedUserJob(cacheService: CacheService, snsClient: AmazonSNSAsy
           )
         }
       }
+    }
   }
 }
--- a/hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
+++ b/hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
@@ -48,7 +48,7 @@ class IamUnrecognisedUserJob(cacheService: CacheService, snsClient: AmazonSNSAsy
     allCredsReports.map { case (account, eitherCredsReportOrFailure) =>
       eitherCredsReportOrFailure.map { credsReport =>
         val humanUsers = credsReport.humanUsers
-        val unrecognisedIamUsers: Seq[VulnerableUser] = filterUnrecognisedIamUsers(humanUsers, "name", janusUsers)
+        val unrecognisedIamUsers: Seq[VulnerableUser] = filterUnrecognisedIamUsers(humanUsers, janusUsers)
 
         if(unrecognisedIamUsers.nonEmpty) {
           //TODO these should return a value that we can inspect to only send notification when successful
--- a/hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
+++ b/hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
@@ -43,9 +43,9 @@ class IamUnrecognisedUserJob(cacheService: CacheService, snsClient: AmazonSNSAsy
     )
 
     val janusUsers: Seq[String] = getJanusUsernames(dummyJanusData)
-    val credsReport = cacheService.getAllCredentials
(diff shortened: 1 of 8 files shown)


hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
```diff
@@ -1,33 +1,71 @@
...
+      None
+    )
+
+    val janusUsers: Seq[String] = getJanusUsernames(dummyJanusData)
+    val allCredsReports = cacheService.getAllCredentials
+
+    allCredsReports.map { case (account, eitherCredsReportOrFailure) =>
+      eitherCredsReportOrFailure.map { credsReport =>
+        val humanUsers = credsReport.humanUsers
+        val unrecognisedIamUsers: Seq[VulnerableUser] = filterUnrecognisedIamUsers(humanUsers, "name", janusUsers)
```
jorgeazevedo, line 51: The interface for `filterUnrecognisedIamUsers` opens up the possibility for the method to be called with different `targetKey` values. Is that the case? If not I'm wondering if it should be a constant in `IamUnrecognisedUsers`

jfsoul: Absolutely, we discussed that but didn't go back and change it. Maybe we could address in #309 where it's introduced

hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
```diff
@@ -1,33 +1,71 @@
...
 
-  //TODO disable access key and remove password
-  //TODO send notification that this has been done
+        if(unrecognisedIamUsers.nonEmpty) {
```
jorgeazevedo, line 53: Question on scala best practice. What's the party line on side effects from a map? Should the  bit of code that runs when this is true be in a separate forEach?

jorgeazevedo (approved):

hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
jfsoul: Party line is use foreach 👍 

There's probably a valuable refactor to give some of these methods return types and centralise the error handling, but I think that can come later and applies to the other scheduled job too

hq/test/schedule/IamUnrecognisedUserTest.scala
```diff
@@ -1,17 +1,16 @@
...
-import model.{AccessKey, CredentialReportDisplay, Green, HumanUser, NoKey, Tag, AwsAccount => Account}
+import model._
 import org.joda.time.{DateTime, Seconds}
 import org.scalatest.{FreeSpec, Matchers}
-import schedule.unrecognised.IamUnrecognisedUsers.{filterUnrecognisedIamUsers, getHumanUsers, getJanusUsernames}
-import utils.attempt.FailedAttempt
+import schedule.unrecognised.IamUnrecognisedUsers.{USERNAME_TAG_KEY, filterUnrecognisedIamUsers, getJanusUsernames}
 
 class IamUnrecognisedUserTest extends FreeSpec with Matchers {
   val humanUser1 = HumanUser("", true, AccessKey(NoKey, None), AccessKey(NoKey, None), Green, None, None, List(Tag("name", "ade.bimbola")))
```
Nirvikalpa108, line 10:
```suggestion
  val humanUser1 = HumanUser("", true, AccessKey(NoKey, None), AccessKey(NoKey, None), Green, None, None, List(Tag(USERNAME_TAG_KEY, "ade.bimbola")))
```

hq/app/schedule/unrecognised/IamUnrecognisedUsers.scala
```diff
@@ -1,22 +1,20 @@
...
 
-  def filterUnrecognisedIamUsers(iamHumanUsersWithTargetTag: List[HumanUser], targetKey: String, janusUsernames: List[String]): List[HumanUser] =
+  def filterUnrecognisedIamUsers(iamHumanUsersWithTargetTag: Seq[HumanUser], janusUsernames: Seq[String]): Seq[VulnerableUser] =
     iamHumanUsersWithTargetTag.filterNot { iamUser =>
-      val maybeTag = iamUser.tags.find(tag => tag.key == targetKey)
+      val maybeTag = iamUser.tags.find(tag => tag.key == USERNAME_TAG_KEY)
       maybeTag match {
         case Some(tag) => janusUsernames.contains(tag.value) // filter out human users that have tags which match the janus usernames
         case None => true
       }
```
Nirvikalpa108, line 18:
```suggestion
      iamUser.tags
        .find(tag => tag.key == USERNAME_TAG_KEY)
        .exists(tag => janusUsernames.contains(tag.value))
```

Nirvikalpa108 (approved):

hq/app/schedule/unrecognised/IamUnrecognisedUsers.scala
jfsoul: Does that work if the tag doesn't exist (where we want the expression to return true)? Either way, we can update this code in another PR if worthwhile.

PR #311, opened 2021-09-16
Creds reaper testing: finding flagged users

jfsoul:
## What does this change?
This is the second part of a closer look at the test coverage of the credentials reaper feature. With this PR I was targeting testing of the functionality that isolates users from the credentials report that are actually vulnerable. This logic is contained in `IamFlaggedUsers` and so I was looking at how to effectively test that class. In the end, both the tests and source code were changed in the following ways:
- `IamFlaggedUsers` and the tests were simplified a bit by moving a small piece of notifications logic. This leads to a return type including `VulnerableUser` rather than `IamAlertTargetGroup`. This is beneficial to isolate the notification types/logic from `IamFlaggedUsers`, and it'll mean we have decent test coverage of `IamFlaggedUsers` because that was the main untested part
- `getVulnerableUsers` was also moved into `IamDeadline` as a notifictions-related piece of logic (although this was much simpler because it was already coupled with `IamDeadline` more than `IamFlaggedUsers`)
- The tests were also refactored a little to share more test data `val`s - without changing the tests themselves
- `Seq[VulnerableUser]` was used as the primary type in a couple of other places, now that it was available in the top level job. Previously we had to extract the vulnerable users out from each `IamAlertTargetGroup`.

## What is the value of this?
We're working towards better test coverage for this feature and this should cover `IamFlaggedUsers` as part of that.

No

## Any additional notes?
No

### DIFF
--- a/hq/app/schedule/IamNotifications.scala
+++ b/hq/app/schedule/IamNotifications.scala
@@ -48,7 +48,8 @@ object IamNotifications extends Logging {
     IamNotification(warningNotifications, finalNotifications, users)
   }
 
-  private def createVulnerableCredentialsNotification(warning: Boolean, users: Seq[VulnerableUser], awsAccount: AwsAccount, targets: List[Target]): Notification = {
+  // TODO: this should be private as it's only used by tests and internally
+  def createVulnerableCredentialsNotification(warning: Boolean, users: Seq[VulnerableUser], awsAccount: AwsAccount, targets: List[Target]): Notification = {
     val usersWithDeadlineAddedIfMissing = users.map { user =>
       user.copy(disableDeadline = Some(createDeadlineIfMissing(user.disableDeadline)))
     }
--- a/hq/app/schedule/IamNotifications.scala
+++ b/hq/app/schedule/IamNotifications.scala
@@ -5,18 +5,29 @@ import model._
 import org.joda.time.DateTime
 import play.api.Logging
 import schedule.IamMessages._
+import schedule.IamTargetGroups.getNotificationTargetGroups
 import schedule.Notifier.notification
 import schedule.vulnerable.IamDeadline.{createDeadlineIfMissing, sortUsersIntoWarningOrFinalAlerts}
 
 object IamNotifications extends Logging {
 
-  def makeNotification(flaggedCredsToNotify: Map[AwsAccount, Seq[IAMAlertTargetGroup]]): List[IamNotification] = {
-    flaggedCredsToNotify.toList.flatMap { case (awsAccount, targetGroups) =>
-      if (targetGroups.isEmpty) {
+  def makeNotifications(flaggedCredsToNotify: Map[AwsAccount, Seq[VulnerableUser]]): List[IamNotification] = {
+    flaggedCredsToNotify.toList.flatMap { case (awsAccount, users) =>
+      if (users.isEmpty) {
         logger.info(s"found no vulnerable IAM users for ${awsAccount.name}. No notification required.")
         None
       } else {
-        targetGroups.map(tg => createWarningAndFinalNotification(tg, awsAccount, createIamAuditUsers(tg.users, awsAccount)))
+        // TODO: can we just apply getNotificationTargetGroups for all accounts?
+        val targetGroupsFromTags: Seq[IAMAlertTargetGroup] =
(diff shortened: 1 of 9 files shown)


hq/app/schedule/IamNotifications.scala
```diff
@@ -5,18 +5,29 @@ import model._
...
-    flaggedCredsToNotify.toList.flatMap { case (awsAccount, targetGroups) =>
-      if (targetGroups.isEmpty) {
+  def makeNotifications(flaggedCredsToNotify: Map[AwsAccount, Seq[VulnerableUser]]): List[IamNotification] = {
+    flaggedCredsToNotify.toList.flatMap { case (awsAccount, users) =>
+      if (users.isEmpty) {
         logger.info(s"found no vulnerable IAM users for ${awsAccount.name}. No notification required.")
         None
       } else {
-        targetGroups.map(tg => createWarningAndFinalNotification(tg, awsAccount, createIamAuditUsers(tg.users, awsAccount)))
+        // TODO: can we just apply getNotificationTargetGroups for all accounts?
```
jfsoul, line 20: As discussed elsewhere, we think this could probably be consolidated, but we'll leave that for another PR

hq/app/schedule/IamNotifications.scala
```diff
@@ -5,18 +5,29 @@ import model._
...
+        val targetGroupsFromTags: Seq[IAMAlertTargetGroup] =
+          if(awsAccount.name == "Ophan") getNotificationTargetGroups(users)
+          else Seq(IAMAlertTargetGroup(List.empty, users))
+
+        targetGroupsFromTags.map(tg => {
+          // Always include the account as a notification target
+          val targetGroupWithAccounts = tg.copy(targets = tg.targets :+ Account(awsAccount.accountNumber))
+          val iamAuditUsers = createIamAuditUsers(targetGroupWithAccounts.users, awsAccount)
+          createWarningAndFinalNotification(targetGroupWithAccounts, awsAccount, iamAuditUsers)
+        })
```
jfsoul, line 30: This method is a bit more complicated than before, but the justification is that the notification target logic is located in one place now:
- Different rule/method for Ophan using tags
- The addition of an Account target for every account

hq/app/schedule/IamUsersToDisable.scala
```diff
@@ -1,13 +1,13 @@
...
 import org.joda.time.DateTime
 import schedule.vulnerable.IamDeadline.getNearestDeadline
 
 object IamUsersToDisable {
-  def usersToDisable(flaggedUsers: Map[AwsAccount, Seq[IAMAlertTargetGroup]], dynamo: DynamoAlertService, today: DateTime = DateTime.now): Map[AwsAccount, Seq[VulnerableUser]] = {
-    flaggedUsers.map { case (awsAccount, targetGroups) =>
-      awsAccount -> getUsersToDisable(targetGroups.flatMap(_.users), awsAccount, dynamo, today)
+  def usersToDisable(flaggedUsers: Map[AwsAccount, Seq[VulnerableUser]], dynamo: DynamoAlertService, today: DateTime = DateTime.now): Map[AwsAccount, Seq[VulnerableUser]] = {
+    flaggedUsers.map { case (awsAccount, users) =>
+      awsAccount -> getUsersToDisable(users, awsAccount, dynamo, today)
```
jfsoul, line 10: targetGroups was only used for the users, so now they're available at the call site, this is simpler

hq/app/schedule/vulnerable/IamDeadline.scala
```diff
@@ -26,6 +26,12 @@ object IamDeadline {
   }
   def isFinalAlert(deadline: DateTime, today: DateTime = DateTime.now): Boolean = deadline.withTimeAtStartOfDay == today.withTimeAtStartOfDay.plusDays(1)
 
+  def getVulnerableUsersToAlert(users: Map[AwsAccount, Seq[VulnerableUser]], dynamo: DynamoAlertService): Map[AwsAccount, Seq[VulnerableUser]] = {
+    users.map { case (account, users) =>
+      account -> filterUsersToAlert(users, account, dynamo)
+    }
+  }
```
jfsoul, line 33: This was already calling the method below, so seems like a good place for it

Nirvikalpa108 (approved):
This looks good, thank you @jfsoul 👍🏽

PR #313, opened 2021-09-20
Add test cases for edge case bugs in getNearestDeadline

jfsoul:
## What does this change?
This adds a couple of test cases for the `getNearestDeadline` method in `IamDeadline`. I was originally looking to update the tests for all of `IamDeadline` but spotted 2 edge cases with `getNearestDeadline` we might not be covering, which I think makes this worth its own PR. Hopefully the test cases are self-explanatory. They didn't pass without the accompanying changes.
- getNearestDeadline should ignore deadlines in the past(?)
- getNearestDeadline should return "today" when choosing between today and tomorrow

## What is the value of this?
Potentially fixes some edge case bugs

No

## Any additional notes?
Please verify the test cases are valid, and if you can think of any other meaningful cases to add for this method specifically.


Nirvikalpa108 (approved):
Thank you for writing tests for these edge cases and fixing a bug in doing so @jfsoul 👍🏽

PR #315, opened 2021-09-22
Gets Janus Data file in S3

Nirvikalpa108:
Unrecognised staff are verified as such by comparing Janus users to the tags of those with IAM permanent credentials.

If any of the tags (which will have matching usernames to the ones in Janus) are present amongst the IAM users, which are NOT in the list of Janus users, that IAM cred will be disabled.

In order for the above to work, Security HQ needs to read a file in S3 which contains a list of Janus users.


hq/conf/application.conf
```diff
@@ -58,6 +58,10 @@ anghammaradSnsArn = ${ANGHAMMARAD_SNS_TOPIC_ARN}
 
 iamDynamoTableName = ${IAM_DYNAMO_TABLE_NAME}
 
+iamUnrecognisedUserS3Bucket = ${IAM_UNRECOGNISED_USER_S3_BUCKET}
```
adamnfish, line 61: Now there are a few of these, is it worth grouping them for clarity? There's no practical difference in having them be grouped so no need unless you it makes this file easier to read and understand.

hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
```diff
@@ -33,9 +38,17 @@ class IamUnrecognisedUserJob(cacheService: CacheService, snsClient: AmazonSNSAsy
       logger.info(s"Running scheduled job: $description")
     }
 
-    val dummyJanusData: JanusData = ???
-
-    val janusUsers: Seq[String] = getJanusUsernames(dummyJanusData)
+    val janusUsers: Seq[String] = try {
+      val file = scala.io.Source
+        .fromInputStream(s3Client.getObject(s3BucketName.getOrElse(""), s3FileKey.getOrElse("")).getObjectContent)
```
adamnfish, line 43: What does `getOrElse("")` really mean for us here? Do we want to try and load data from different locations in S3 if the configuration is missing?

I'd be tempted to structure this as a bunch of dependent steps, where fetching the data depends on being able to access both of these pieces of configuration. `flatMap` (and therefor a for-comprehension in Scala) is normally a good way to go when you want to express this kind of dependency chain, although it does mean choosing a consistent representation for failure.

Security HQ uses [`Attempt`](https://github.com/guardian/security-hq/blob/main/hq/app/utils/attempt/Attempt.scala) to handle failures elsewhere, so that's probably a good choice for the "effect" to use here, which means each of these steps will return an `Attempt` that represents either the successful value, or a description of the failure.

This is only looking at the diff though, when I look at the context around here I see that this is only one part of the work that's done, so a bigger for-comp that includes the work to get the credentials and perform side effects on the results is probably even better!

I spotted there's a nested foreach a bit lower down, which is always a bit of a red flag to my mind. It sounds like you're already looking at making this more testable, so that'll naturally mean those things get refactored away 👍

hq/app/AppComponents.scala
```diff
@@ -107,6 +109,7 @@ class AppComponents(context: Context)
   private val efsClients = AWS.efsClients(configuration, availableRegions)
   val securityCredentialsProvider =
     new AWSCredentialsProviderChain(new ProfileCredentialsProvider("security"), DefaultAWSCredentialsProviderChain.getInstance())
+  private val securityS3Client = AmazonS3ClientBuilder.standard().withRegion("eu-west-1").withCredentials(securityCredentialsProvider).build()
```
adamnfish, line 112: There's already a mechanism above for getting hold of s3 clients, would it be better to re-use that? Creating instances of AWS clients comes with a bit of overhead - IIRC each has its own threadpool for example. If so, better to re-use the existing cached client created in the `AWS.s3Clients` call, just above.

jorgeazevedo (approved):
Minor comments from me, I think this is pretty solid. Learned a lot going through the implementation!

hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
```diff
@@ -33,27 +43,73 @@ class IamUnrecognisedUserJob(cacheService: CacheService, snsClient: AmazonSNSAsy
...
+    removePasswords(account, users, iamClients)
+    sendNotification(account, users, testMode)
+  }
 
-    allCredsReports.foreach { case (account, eitherCredsReportOrFailure) =>
-      eitherCredsReportOrFailure.foreach { credsReport =>
-        val humanUsers = credsReport.humanUsers
-        val unrecognisedIamUsers: Seq[VulnerableUser] = filterUnrecognisedIamUsers(humanUsers, janusUsers)
+  private def getS3Object(s3Client: AwsClient[AmazonS3], bucket: String, key: String): Attempt[BufferedSource] = {
+    Attempt.Right(
```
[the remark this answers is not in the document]
jorgeazevedo: I've been able to step through the code now and yes - the exception is caught, the for comprehension exits early and the error message gets logged in the result fold.

PR #316, opened 2021-09-27
Add user with no tags to users under test in IamUnrecognisedUsersTest

jorgeazevedo:
## What does this change?
This raises the test coverage for the `IamUnrecognisedUsers` object to 100%, as having no tags is part of the pattern match in `filterUnrecognisedIamUsers`.

## What is the value of this?

The behaviour for untagged users is an important part of the algorithm, so it's useful to have that expressed in the test. Namely, untagged users are left alone

No


Nirvikalpa108 (approved):
Thanks for adding this @jorgeazevedo 👍🏽

PR #317, opened 2021-09-29
Remove configraun dependency and read Snyk config from file

jacobwinch:
## What does this change?

This PR removes the [configraun](https://github.com/guardian/configraun) dependency, as it is no longer maintained and is not published for Scala 2.13. 

I was originally planning to replace [configraun](https://github.com/guardian/configraun) with [simple-configuration](https://github.com/guardian/simple-configuration), but whilst researching this I noticed that we are only using this library to read 2 strings from parameter store anyway. All [other private config is currently fetched from S3](https://github.com/guardian/security-hq/blob/035c0f8de5d1a7a50b2b065d5d6878f1b779e09f/cloudformation/security-hq.template.yaml#L352) on startup.

The configraun dependency was originally introduced in https://github.com/guardian/security-hq/pull/64 with a view to migrating all config to parameter store, but as there are no current plans to do this I've opted to consolidate all config in S3 and simplify the code here instead.

## What is the value of this?

* Helps to unblock the migration to Scala 2.13
* Simplifies the codebase as all config is now read from the same place, using the same pattern

No

## Any additional notes?

Strangely the Snyk functionality (which relies on the config being changed here) will not work locally (regardless of whether I'm running `main` or this feature branch), which made testing this change a bit painful. I was getting a request timeout (after 2 minutes!) when attempting to make HTTP requests to Snyk's API. I've tested this in PROD and it's working fine, but I'd be interested to hear any theories which could explain this unusual behaviour in my local environment!

## To do (after merge)

- [ ] Remove redundant config from parameter store
- [ ] Remind team to pull new local config file (the app will not start up correctly if Snyk config is not found)

### DIFF
--- a/hq/app/api/Snyk.scala
+++ b/hq/app/api/Snyk.scala
@@ -5,7 +5,6 @@ import model._
 import play.api.libs.json._
 import play.api.libs.ws.{WSClient, WSResponse}
 import utils.attempt.{Attempt, FailedAttempt, Failure}
-import com.gu.configraun.models.Configuration
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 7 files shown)


hq/app/api/Snyk.scala
```diff
@@ -80,18 +76,4 @@ object Snyk {
     } yield results
   }
 
-  def getSnykToken(configraun: Configuration): Attempt[SnykToken] = configraun.getAsString("/snyk/token") match {
```
jacobwinch, line 83: The app will now fail to start if these config values are not present. This makes it more consistent with other configuration e.g. https://github.com/guardian/security-hq/blob/035c0f8de5d1a7a50b2b065d5d6878f1b779e09f/hq/app/AppComponents.scala#L101

hq/app/config/Config.scala
```diff
@@ -110,8 +110,15 @@ object Config {
     } yield AwsAccount(id, name, roleArn, number)
   }
 
+  def getSnykConfig(config: Configuration): SnykConfig = {
+    SnykConfig(
+      SnykToken(requiredString(config, "snyk.token")),
+      SnykGroupId(requiredString(config, "snyk.organisation"))
```
jacobwinch, line 116: I'm not sure why this is called `GroupId` here and `organisation` in the private config; I've left that unchanged as I'm not too familiar with Snyk jargon

Nirvikalpa108:
I also haven't been able to get a response from https://security-hq.local.dev-gutools.co.uk/snyk. I've created a card here: https://trello.com/c/tAo9ou6W.

Nirvikalpa108 (approved):
These changes all look very clear and straight forward to me @jacobwinch thank you for doing this work. I've written up a ticket about /synk not running locally and hopefully we can fix that soon.

jacobwinch:
> I've written up a ticket about /synk not running locally and hopefully we can fix that soon.

Thanks for adding this card @Nirvikalpa108!

PR #318, opened 2021-09-29
Port across functionality from Box

jacobwinch:
## What does this change?

This PR ports across a simplified version of https://github.com/guardian/box. 

This library was originally used to help teams to move away from akka Agents, but it is no longer worked on and is not published for more recent Scala versions. As the code that we need to keep is minimal, I think this copy+paste option is acceptable.

## What is the value of this?

* Removes a dependency and unblocks Scala 2.13 upgrade

No


Nirvikalpa108 (approved):
Thank you @jacobwinch, this change makes sense to me 👍🏽

PR #319, opened 2021-10-04
Create separate dynamoDB table for DEV mode

Nirvikalpa108:
## What does this change?
Currently, the PROD dynamo table is used for testing purposes, but this is not ideal. This PR creates a DEV table to be used for this instead.

## What is the value of this?
This means that developers working on SHQ will need to change the value of IAM_DYNAMO_TABLE_NAME in their local conf file.

Also this PR deletes an unused bash script related to an old local dynamoDB implementation.

No

Yes, this PR adds a new CFN template.

N/A


cloudformation/security-hq-dev.template.yaml
```diff
@@ -0,0 +1,30 @@
...
+      Value: security
+    App:
+      Value: security-hq
+
+Resources:
+  SecurityHqIamDynamoTableDev:
+    Type: AWS::DynamoDB::Table
+    DeletionPolicy: Retain
+    Properties:
+      TableName: !Sub security-hq-iam-DEV
```
akash1810, line 16: I don't think we need the `!Sub` here.

cloudformation/security-hq-dev.template.yaml
```diff
@@ -0,0 +1,30 @@
...
+        - { AttributeName: id, AttributeType: S }
+      KeySchema:
+        - { AttributeName: id, KeyType: HASH }
+      ProvisionedThroughput: { ReadCapacityUnits: 5, WriteCapacityUnits: 5 }
+      Tags:
+        - Key: Stage
+          Value: DEV
+        - Key: Stack
+          Value:
+            Fn::FindInMap: [ Constants, Stack, Value ]
```
akash1810, line 27: As the template is small, I'd be tempted to inline these values and remove the complexity of `Mappings`.

cloudformation/security-hq-dev.template.yaml
jorgeazevedo: good catch!

cloudformation/security-hq-dev.template.yaml
jorgeazevedo: Yeah, agree, there's no real need to keep the approach from the PROD stack

PR #321, opened 2021-10-08
 Restrict aws effects

Nirvikalpa108:
There are a couple of things this PR does.

# ⭐️ IAM unrecognised job

### What does this PR do?
This PR completes the logic required to disable the IAM permanent credentials of an engineer who does not have any [janus](https://github.com/guardian/janus-app) credentials. 

### Why is this beneficial? 
If an engineer has left the Guardian and they have been removed from Janus, but their permanent IAM credential has _not_ been removed, Security HQ will fix this automatically by disabling the IAM user.

### How is it implemented?
This is implemented by comparing Janus users to the tags of those with IAM permanent credentials. If any of the tags (which will have matching usernames to the ones in Janus) are present amongst the IAM users, which are NOT in the list of Janus users, that IAM cred will be disabled.

### Are there any next steps?
Once this code is merged, the next step for this feature will be to add tags to all IAM users which match their janus usernames and show on SHQ's IAM dashboard when this has not been done.

# ⭐️ AWS API restrictions by Stage
Up until now Security HQ DEV could make changes to AWS infrastructure if the developer running it had janus credentials to the AWS account. This is not a pattern we want to continue. 

Rather, SHQ PROD should have the power to make changes to AWS infra, rather than DEV.

To fix this a config property has been added, which means developers must explicitly list the AWS account ids that
Security HQ is permitted to make changes to. 

This has been added to the unrecognised job's side effects of removing passwords and disabling keys. The same will need to be done for the vulnerable user job.

### Are there any next steps?
Also this change requires an additional config field, which will need to be added to the two config files in S3.


hq/app/config/Config.scala
```diff
@@ -114,6 +116,57 @@ object Config {
...
+      FailedAttempt(Failure("unable to get IAM unrecognised user bucket from config",
+        "unable to get IAM unrecognised user bucket from config",
+        500)
+      )
+    )
+  }
+
+  def getSecurityAccount(config: Configuration): Attempt[AwsAccount] = {
+    Attempt.fromOption(
+      Config.getAwsAccounts(config).find(_.name == "security"),
```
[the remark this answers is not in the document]
Nirvikalpa108: Thank you so much for spotting this @jorgeazevedo, this is a bug! It was changed to use the id, instead of the name, but this part of the code mistakenly wasn't changed. Thank you!

hq/test/schedule/unrecognised/IamUnrecognisedUsersTest.scala
```diff
@@ -0,0 +1,125 @@
...
+  val humanUser1 = HumanUser("ade.bimbola", true, AccessKey(NoKey, None), AccessKey(NoKey, None), Green, None, None, List(Tag(USERNAME_TAG_KEY, "ade.bimbola")))
+  val humanUser2 = humanUser1.copy(username = "john.akindele", tags = List(Tag(USERNAME_TAG_KEY, "john.akindele")))
+  val humanUser3 = humanUser1.copy(username = "khadija.omodara", tags = List(Tag(USERNAME_TAG_KEY, "khadija.omodara")))
+  val humanUser4 = humanUser1.copy(username = "nneka.obi", tags = List(Tag(USERNAME_TAG_KEY, "nneka.obi")))
+  val humanUser5 = humanUser1.copy(username = "john.doe", tags = List())
+  val humanUser6 = humanUser1.copy(username = "jane.doe", tags = List(Tag(USERNAME_TAG_KEY, "jane.doe")))
+  val allJanusUsernames = List("ade.bimbola", "john.akindele", "khadija.omodara", "nneka.obi", "jane.doe")
+
+  "findUnrecognisedIamUsers" - {
+    "get janus usernames" in {
```
adamnfish, line 21: I don't think you've actually added this one (it's copied from elsewhere) but this test label doesn't appear to be doing a great job of describing what's being checked. Happy to leave this for now and come back to it later, or perhaps we can add some clarity to the label now?

hq/test/schedule/unrecognised/IamUnrecognisedUsersTest.scala
```diff
@@ -0,0 +1,125 @@
...
+      getJanusUsernames(dummyJanusData) shouldEqual List("firstName.secondName")
+    }
+  }
+
+  "getCredsReportDisplayForAccount" - {
+    val failedAttempt: FailedAttempt = FailedAttempt(Failure("error", "error", 500))
+
+    "if an either is a left, aws account does not appear in response" in {
+      val accountCredsLeft = Map(1 -> Left(failedAttempt))
+      getCredsReportDisplayForAccount(accountCredsLeft) shouldEqual Nil
```
adamnfish, line 39: This assertion doesn't seem to *quite* match the test description. It says we're checking that the account doesn't appear in the response, but we're actually checking that the response for this test data is an empty list. Could the assertion instead specifically check that the test element is absent, or could this test description switch to say words to the effect of "empty list for this input"?

hq/test/schedule/unrecognised/IamUnrecognisedUsersTest.scala
```diff
@@ -0,0 +1,125 @@
...
+    }
+    "if an either is a right, aws account included in response" in {
+      val accountCredsRight = Map(1 -> Right(1))
+      getCredsReportDisplayForAccount(accountCredsRight) shouldEqual List((1,1))
```
adamnfish, line 43: Here as well, we're not quite checking what we say we will. Instead of checking whether the account is included in the response, we're checking whether the response is precisely a list containing only this element.

I do't think either of these are a big deal, very happy with either approach (tweak the description or the assertion).

hq/test/schedule/unrecognised/IamUnrecognisedUsersTest.scala
```diff
@@ -0,0 +1,125 @@
...
+      val accountCredsAllLeft = Map(1 -> Left(failedAttempt), 2 -> Left(failedAttempt), 3 -> Left(failedAttempt))
+      getCredsReportDisplayForAccount(accountCredsAllLeft) shouldEqual Nil
+    }
+    "if every either is a right, output list contains same number of elements" in {
+      val accountCredsAllRight = Map(1 -> Right(1), 2 -> Right(2), 3 -> Right(3))
+      getCredsReportDisplayForAccount(accountCredsAllRight) should have length 3
+    }
+  }
+
+  "unrecognisedUsersForAllowedAccounts" - {
```
adamnfish, line 58: Love these! There are a couple of minor points, but they're a bit tangled with each other so easier to discuss on a call / in person.

hq/test/schedule/unrecognised/IamUnrecognisedUsersTest.scala
```diff
@@ -0,0 +1,125 @@
...
+import schedule.unrecognised.IamUnrecognisedUsers.{USERNAME_TAG_KEY, getCredsReportDisplayForAccount, getJanusUsernames, unrecognisedUsersForAllowedAccounts}
+import utils.attempt.{FailedAttempt, Failure}
+
+class IamUnrecognisedUsersTest extends FreeSpec with Matchers {
+  val humanUser1 = HumanUser("ade.bimbola", true, AccessKey(NoKey, None), AccessKey(NoKey, None), Green, None, None, List(Tag(USERNAME_TAG_KEY, "ade.bimbola")))
+  val humanUser2 = humanUser1.copy(username = "john.akindele", tags = List(Tag(USERNAME_TAG_KEY, "john.akindele")))
+  val humanUser3 = humanUser1.copy(username = "khadija.omodara", tags = List(Tag(USERNAME_TAG_KEY, "khadija.omodara")))
+  val humanUser4 = humanUser1.copy(username = "nneka.obi", tags = List(Tag(USERNAME_TAG_KEY, "nneka.obi")))
+  val humanUser5 = humanUser1.copy(username = "john.doe", tags = List())
+  val humanUser6 = humanUser1.copy(username = "jane.doe", tags = List(Tag(USERNAME_TAG_KEY, "jane.doe")))
```
adamnfish, line 17: John / Jane Doe are traditionally used as stand-ins for unknown identities, so it's a little odd to see Jane Doe be used here for a Janus user? Is there a different name we could use that doesn't have that association? This is a spectacularly minor point now, apologies!

adamnfish (commented):
Wonderful! This is a great change, I'm really happy to see it. I've done a very detailed review, but the only really important things are to for-comp the `disableUser` implementation and have `removePasswords` return an Attempt. Let's chat these over together.

hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
```diff
@@ -33,24 +43,39 @@ class IamUnrecognisedUserJob(cacheService: CacheService, snsClient: AmazonSNSAsy
...
     }
 
-    val dummyJanusData: JanusData = ???
-
-    val janusUsers: Seq[String] = getJanusUsernames(dummyJanusData)
-    val allCredsReports = cacheService.getAllCredentials
+    val result = for {
+      config <- getConfigProperties(config)
+      client <- s3Clients.get(config.securityAccount)
+      s3Object <- getS3Object(client, config.janusUserBucket, config.janusDataFileKey)
```
jorgeazevedo, line 49: This s3GetObject call fails with 
```
The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'eu-west-1' (Service: Amazon S3; Status Code: 400; Error Code: AuthorizationHeaderMalformed;)
```

It works if s3Clients in AppComponents is initialised with the eu-west-1 region
```
  private val s3Clients = AWS.s3Clients(configuration, Regions.EU_WEST_1)
```
This is not necessarily the right solution, but there's something going on with the region!

hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
```diff
@@ -33,30 +46,48 @@ class IamUnrecognisedUserJob(cacheService: CacheService, snsClient: AmazonSNSAsy
...
+      { notificationIds =>
+        logger.info(s"Successfully ran unrecognised user job and sent ${notificationIds.length} notifications.")
+      }
+    )
+  }
 
-    val janusUsers: Seq[String] = getJanusUsernames(dummyJanusData)
-    val allCredsReports = cacheService.getAllCredentials
+  private def makeFile(s3Object: String): File =
+    Files.write(Paths.get("janusData.txt"), s3Object.getBytes(StandardCharsets.UTF_8)).toFile
```
Nirvikalpa108, line 71: Let's discuss this part please @adamnfish

adamnfish (approved):
Wonderful work, thank you! This is looking loads better, and gives us a much stronger foundation to keep building upon 👍

PR #326, opened 2021-10-27
Add basic GHA workflow

nicl:
## What does this change?

Replaced the current Teamcity build with Github Actions. Also bumped sbt-riffraff-artifact as part of this to the latest version.

## What is the value of this?

In anticipation of further build changes (to generate cloudformation via CDK) it made sense to update the build process to GHA first - GHA is preferred now as it allows us to more easily codify (+ version control) the entire build process.

No.

No.

## Any additional notes?

* Still to test (on PROD?).
* Uncertainty over how the lambdas get built, so for now I've just ported the existing 'security-hq' TC build


build.sbt
```diff
@@ -85,15 +86,25 @@ lazy val hq = (project in file("hq"))
...
+    serverLoading in Debian := Some(Systemd),
+    debianPackageDependencies := Seq("openjdk-8-jre-headless"),
+    maintainer := "Security Team <devx.sec.ops@guardian.co.uk>",
+    packageSummary := "Brief description",
+    packageDescription := """Slightly longer description""",
     riffRaffPackageType := (packageBin in Debian).value,
     riffRaffUploadArtifactBucket := Option("riffraff-artifact"),
     riffRaffUploadManifestBucket := Option("riffraff-builds"),
+
+    riffRaffAwsRegion := "eu-west-1", // Needed for GHA while we improve how the plugin works here.
```
akash1810, line 99: This shouldn't be needed as the assume-aws-role action [sets some env vars already](https://github.com/guardian/actions-assume-aws-role/blob/main/src/main.ts#L39-L40). Curious to know how change performs without this.

.github/workflows/ci-sbt.yml
```diff
@@ -0,0 +1,49 @@
...
+          distribution: 'adopt'
+
+      # 1. Seed the build number with last number from TeamCity.
+      #    Update `LAST_TEAMCITY_BUILD` as needed or remove entirely if changing Riff-Raff project name.
+      # 2. Execute `script/ci`, a script that will compile, run tests etc.
+      #    Ensure this file exists in your repository and has executable permissions (`chmod u+x script/ci`).
+      # See https://github.com/github/scripts-to-rule-them-all
+      - name: Run script/ci
+        run: |
+          LAST_TEAMCITY_BUILD=0
```
akash1810, line 47: You'll want to set this to non-zero, and pause TeamCity, if you want to have continuous numbers with TeamCity.

akash1810: I think we're currently overwriting previous builds with this being set to 0. Not a huge deal as we're never going to deploy such old versions, but better to seed GHA from the last TC build.

build.sbt
nicl: Interesting, will take another look. I was seeing errors on this when not explicitly set - about failure to retrieve from instance metadata.

build.sbt
```diff
@@ -85,15 +86,25 @@ lazy val hq = (project in file("hq"))
     testOnly in Test := (testOnly in Test).dependsOn(startDynamoDBLocal).evaluated,
     testQuick in Test := (testQuick in Test).dependsOn(startDynamoDBLocal).evaluated,
     testOptions in Test += dynamoDBLocalTestCleanup.value,
+
+    serverLoading in Debian := Some(Systemd),
+    debianPackageDependencies := Seq("openjdk-8-jre-headless"),
+    maintainer := "Security Team <devx.sec.ops@guardian.co.uk>",
+    packageSummary := "Brief description",
```
nicl, line 93: Doh, I should populate these!

build.sbt
akash1810: Ah. I _think_ this is the result of [this call](https://github.com/guardian/sbt-riffraff-artifact/blob/master/src/main/scala/com/gu/riffraff/artifact/RiffRaffArtifact.scala#L65), specifically `Regions.getCurrentRegion`?

This is reported incorrectly by the plugin, it is a warning, not an error.

.github/workflows/ci-sbt.yml
nicl: @akash1810 hey yep, at the moment we're on 8 here, and I was content to overwrite (a couple of very old) things as didn't want to block TC deploys. But yeah, we should set this once I've got the okay to merge.

adamnfish: If negative numbers work then maybe there's no need to overwrite things? No harm though, I can't see us wanting to deploy anything that old.

.github/workflows/ci-sbt.yml
```diff
@@ -0,0 +1,49 @@
+# Find full documentation here https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions
+name: CI
+
+on:
+  pull_request:
```
adamnfish, line 5: Do we need to do anything to be cautious about building PRs from forks, here? I think there are a few interacting settings in GHA for this, so it'd be worth making sure we're 100% sure how this behaves for public repos, and that we're all consistently using a safe approach.

adamnfish (approved):
Looks great, thanks heaps for making this change. There are a few tangential changes, so if there's any difficulty with testing this I'd recommend stripping it back to just the CI change. It all looks sensible though, thanks again.

.github/workflows/ci-sbt.yml
akash1810: From https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request

> Workflows don't run on forked repositories by default. You must enable GitHub Actions in the Actions tab of the forked repository.
> 
> When a first-time contributor submits a pull request to a public repository, a maintainer with write access may need to approve running workflows on the pull request.

I think this is safe.

More here https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks.

PR #328, opened 2021-11-03
Introduce CDK

nicl:
## What does this change?

Introduces CDK. At the moment it merely wraps the existing template.

Note, most of the CDK boilerplate/scripts are from https://github.com/guardian/cdk-cli. They work (though I've done minor edits), but I think we could tidy this a bit/make things more concise. In general, it feels like a lot of files to check in, which works against our 'CDK simplifies things' motivation a bit.

## What is the value of this?

CDK is more concise and better supported with patterns for best practice. We want to migrate Security HQ to CDK fully and this is the initial step.

Not for the stackset lambda, but it does affect the normal app deploy (including Cloudformation).

The Cloudformation generated should be equivalent though, just with some additional tags.

No.

## Any additional notes?

TODO - check the diff and share here.


cdk/lib/security-hq.ts
```diff
@@ -0,0 +1,17 @@
+import { CfnInclude } from '@aws-cdk/cloudformation-include';
```
nicl, line 1: This is the file with the actual CDK code for Security HQ. At the moment, it simply includes the existing template.

cdk/script/ci
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -e
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+
+source ${DIR}/setup
+source ${DIR}/test
+
+for file in bin/*; do
```
akash1810, line 10: With only one file in the `bin` directory, we can drop the `for` I think (the CLI needs updating!)

cdk/tsconfig.eslint.json
```diff
@@ -0,0 +1,5 @@
+{
+  "extends": "./tsconfig.json",
+  "include": ["lib/**/*", "bin/**/*"],
+  "exclude": ["node_modules", "../hq/public/javascripts"]
```
akash1810, line 4: 😨  Does this mean ESLint looks at the entire repo and not the directory it's installed in?!

script/ci
```diff
@@ -1,2 +1,7 @@
 #!/usr/bin/env bash
+
+pushd cdk > /dev/null || exit
+./script/ci
+popd > /dev/null
```
akash1810, line 5: A slightly easier version of this which I've seen in other repos is:

```bash
(
  cd cdk
  ./script/ci
)
```

cdk/script/tsc
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
akash1810, line 1: I think under the [scripts pattern](https://github.com/github/scripts-to-rule-them-all) this file would be `script/build`. I think the key is using agnostic language that applies to any repository to provide a common experience.

cdk/script/ci
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -e
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+
+source ${DIR}/setup
```
akash1810, line 7: I'd inline these into the `ci` file for simplicity. I think we'd also want to add `--frozen-lockfile` to `yarn install` in CI for a deterministic build.

From https://classic.yarnpkg.com/en/docs/cli/install/:

> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass --frozen-lockfile flag.

cdk/jest.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
akash1810, line 1: We should add the GuCDK mocks too - https://github.com/guardian/cdk/blob/main/docs/001-general-usage.md#-mocking-gucdkversion.

akash1810 (commented):
Looks great! A few small comments. You've reminded me how out of date the gucdk-cli is and how much boilerplate is needed!

A couple of scripts can be simplified - [AMIgo](https://github.com/guardian/amigo/tree/main/cdk/script) might provide a reference point?

cdk/tsconfig.eslint.json
nicl: apparently so

script/ci
nicl: ah wow, nice.

cdk/script/ci
nicl: I've updated to drop for now, but suspect will reintroduce when the follow on PR hits, which will also introduce a stack for the VPC.

cdk/tsconfig.eslint.json
nicl: Although I'm not sure why! I'll check this.

nicl: Okay, can remove this. I think I must have been linting the wrong directory at the time.

cdk/script/ci
akash1810: Ah, forgot about that. Sorry for the pain!

cdk/lib/security-vpc.ts
```diff
@@ -0,0 +1,11 @@
+import type { App } from '@aws-cdk/core';
+import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
+import { GuStack } from '@guardian/cdk/lib/constructs/core';
+import { GuVpc } from '@guardian/cdk/lib/constructs/vpc';
+
+export class SecurityVpc extends GuStack {
+  constructor(scope: App, id: string, props: GuStackProps) {
+    super(scope, id, props);
+    new GuVpc(this, 'SecurityVpc', { cidr: 'TODO - get from Enterprise.' });
```
akash1810, line 9: Would the CIDR be private or secret? If so we should draw this from https://github.com/guardian/private-infrastructure-config and [mock it for tests](https://github.com/guardian/private-infrastructure-config#using-with-jest) so it's not exposed in snapshots.

I'd suggest adding the VPC stack in a PR of it's own, to keep this change set in this PR small. Well, as small as it can be!

nicl: I'm not sure why the CIDR would need to be private - do you know why this would matter? 

In any case, I should probably remove this file from this PR, as it's not in use yet. But will add it back in the follow up.

cdk/.nvmrc
```diff
@@ -0,0 +1 @@
+v16.13.0
```
akash1810, line 1: ✨ shiny!

cdk/script/ci
```diff
@@ -0,0 +1,30 @@
...
+  if ! nvm_available; then
+    [ -e "$HOME/.nvm/nvm.sh" ] && source "$HOME/.nvm/nvm.sh"
+  fi
+}
+
+source_nvm
+nvm install
+nvm use
+
+npm install -g yarn
```
akash1810, line 22: Can we use the `guardian/actions-setup-node@v2.4.1` in the workflow to provide a node version from the `.nvmrc` file auto-magically instead of this `nvm` dance? It also comes with `yarn` so we don't have to globally install it.

https://github.com/guardian/.github/blob/main/workflow-templates/ci-sbt-node.yml#L34 might be helpful.

nicl: Ah great, yep was copying from Amigo but this is nicer.

cdk/script/ci
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -e
+
+npm install -g yarn
```
akash1810, line 5: I'm fairly certain `yarn` is already available via `actions-setup-node`, so this would be a no-op / can be removed.

akash1810 (approved):
One last minor comment, not necessarily a blocker though. Fantastic work.

img

PR #330, opened 2021-11-05
Add CDK VPC

nicl:
## What does this change?

Adds a new VPC. This is in preparation for a fuller CDK migration of the main app, including into this new VPC. The new VPC is required as the existing one does not have private subnets.

## What is the value of this?

Required step towards full CDK migration of Security HQ, including running the app in private subnets for security in depth.

Not to the StackSet but there is the new VPC cloudformation CDK/template.

No.

## Any additional notes?

For now, this is continuously built and deployed with the Security HQ Play app. It makes sense to separate this if/when the VPC is used for additional Security apps.

TODOs:

*~ensure name of cfn stack is security-vpc-PROD~
* ~protect VPC resources in riffraff during cloudformation deploys~
* ~set removal policy on GuVpc to prevent deletion by default~ (Scrapped as RR policy plus in-built AWS protections are sufficient)


cdk/README.md
```diff
@@ -13,7 +14,26 @@ find useful scripts within the [`script`](./script) directory for common tasks.
...
+Both are continuously built and deployed.
+
+As `SecurityHQ` depends on the output of `SecurityVpc`, if creating a new
+staging environment, you will need to deploy only the VPC step in Riffraff
+initially and use the resulting VPC ID to configure the `SecurityHQ` stack
+before you can do a full deploy of both.
+
+If/when the VPC is used for other apps it would make sense to move the
+`SecurityVpc` stack into another repo and deploy it at a separate cadence rather
+than for every Security HQ update.
```
akash1810, line 39: 👍🏽

akash1810 (commented):
LGTM - will formally approve once the CIDR value is known.

cdk/bin/security-hq.ts
```diff
@@ -4,6 +4,8 @@
 import 'source-map-support/register';
 import { App } from '@aws-cdk/core';
 import { SecurityHQ } from '../lib/security-hq';
+import { SecurityVpc } from '../lib/security-vpc';
 
 const app = new App();
+new SecurityVpc(app, 'security-vpc', { stack: 'security ' });
```
akash1810, line 10: Is the extra space for `stack` intentional?

akash1810 (approved):
A comment on a potential typo, LGTM otherwise.

cdk/bin/security-hq.ts
nicl: oh no, not sure how that got added! Thanks.

PR #335, opened 2021-11-17
Unrecognised User job - access keys should be disabled regardless of age, when a colleague departs

Nirvikalpa108:
## What does this change?
The `disableAccessKeys` function used by the Unrecognised User job was only disabling outdated keys. This is a mistake! All keys should be disabled regardless of age, because the person will have left the company.

No.

No.

### DIFF
--- a/hq/app/schedule/vulnerable/IamDisableAccessKeys.scala
+++ b/hq/app/schedule/vulnerable/IamDisableAccessKeys.scala
@@ -24,8 +24,7 @@ object IamDisableAccessKeys extends Logging {
   )(implicit ec: ExecutionContext): Attempt[List[UpdateAccessKeyResult]] = {
     val result = for {
       accessKeys <- listAccountAccessKeys(account, vulnerableUsers, iamClients)
-      outdatedAccessKeys = getOutdatedKeys(accessKeys)
-      updateAccessKeyRequests = outdatedAccessKeys.map(updateAccessKeyRequest)
+      updateAccessKeyRequests = accessKeys.map(updateAccessKeyRequest)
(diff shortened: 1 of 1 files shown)


hq/test/schedule/AwsDynamoAlertServiceTest.scala
```diff
@@ -1,114 +0,0 @@
-package schedule
```
jorgeazevedo, line 1: @Nirvikalpa108 was removing the dynamo tests an intentional change?

jfsoul: From the commit message, I think it was failing. Going to take a quick look because I think I've worked on these tests.

jfsoul: They're passing on this branch for me if I check it out and re-add the test. Happy to help diagnose if they're failing for you @Nirvikalpa108 , let me know.

adamnfish (approved):
This change looks ok to me! Thank you

hq/test/schedule/AwsDynamoAlertServiceTest.scala
adamnfish: It's true that this test is for the work that we are re-doing, so it doesn't need to be in `main` (and will be removed by anyway in a subsequent PR to merge these two streams of work).

I agree it's not worth the time to fix them in-place. We won't need Database-backed tests for the logic of the new jobs in general, but we may need a bit of coverage for the Database wrapper. I expect while writing that job we can refer back to this (now deleted) file as a reference.

PR #329, opened 2021-11-05
CDK full

nicl:
Previous steps:

* [x] https://github.com/guardian/security-hq/pull/328
* [x] https://github.com/guardian/security-hq/pull/330

## What does this change?

This replaces the existing main cloudformation template with a version generated fully by CDK.

Importantly, the stack will run in a new VPC, and there is a separate (CDK-generated) template for this. Ultimately, it may make sense to move this VPC template elsewhere, as it will likely be shared across security apps - the security team are best placed to decide where. But for the initial CDK migration, it seems easiest to keep it here.

Note, the VPC stack is not continuously built or deployed as part of this PR. Ultimately, I think it should be, but  not as part of a standard Security HQ build. It may make sense to add this if/once it is moved into a more suitable location. In any case, VPC updates are rare, so it is not a large concern.

## What is the value of this?

Completes the CDK migration and runs Security HQ in private subnets for security in depth.

## Will this require CloudFormation and/or updates to the AWS StackSet?

No updates to the StackSet, but careful migration of the Cloudformation:

* to create the new stack and verify behaviour (note, how can we do this safely given the app is designed to run as a singleton?)
* to migrate the DNS once we are happy with the new stack

## Will this require changes to config?

No.

## Any additional notes?

Changeset JSON info below (note, purely `add` actions):

```
[
  {
    "resourceChange": {
      "logicalResourceId": "AutoScalingGroupSecurityhqASG8DD277A5",
      "action": "Add",
      "physicalResourceId": null,
      "resourceType": "AWS::AutoScaling::AutoScalingGroup",
      "replacement": null,
      "moduleInfo": null,
      "details": [],
      "changeSetId": null,
      "scope": []
    },
    "hookInvocationCount": null,
    "type": "Resource"
  },
  {
    "resourceChange": {
      "logicalResourceId": "AutoScalingGroupSecurityhqInstanceProfile9E45D8A7",
      "action": "Add",
      "physicalResourceId": null,
      "resourceType": "AWS::IAM::InstanceProfile",
      "replacement": null,
      "moduleInfo": null,
      "details": [],
      "changeSetId": null,
      "scope": []
    },
    "hookInvocationCount": null,
    "type": "Resource"
  },
  {
    "resourceChange": {
      "logicalResourceId": "AutoScalingGroupSecurityhqLaunchConfig4307528E",
      "action": "Add",
      "physicalResourceId": null,
      "resourceType": "AWS::AutoScaling::LaunchConfiguration",
      "replacement": null,
[...]

PR #336, opened 2021-11-18
Take over DNS

nicl:
## What does this change?

Addition to https://github.com/guardian/security-hq/pull/329 to add DNS record.

## What is the value of this?

We can then repoint the DNS in code without having to do (risky) manual NS1 changes.

### DIFF
--- a/cdk/lib/security-hq.ts
+++ b/cdk/lib/security-hq.ts
@@ -10,17 +10,21 @@ import {
   InstanceType,
   Peer,
 } from '@aws-cdk/aws-ec2';
+import type { CfnLoadBalancer } from '@aws-cdk/aws-elasticloadbalancing';
 import type { CfnTopic } from '@aws-cdk/aws-sns';
 import { CfnInclude } from '@aws-cdk/cloudformation-include';
 import { Duration } from '@aws-cdk/core';
 import type { App } from '@aws-cdk/core';
 import { AccessScope, GuApplicationPorts, GuEc2App } from '@guardian/cdk';
+import { Stage } from '@guardian/cdk/lib/constants/stage';
 import { GuAlarm } from '@guardian/cdk/lib/constructs/cloudwatch';
 import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
 import {
   GuDistributionBucketParameter,
   GuStack,
 } from '@guardian/cdk/lib/constructs/core';
+import type { AppIdentity } from '@guardian/cdk/lib/constructs/core/identity';
+import { GuCname } from '@guardian/cdk/lib/constructs/dns';
 import {
   GuDynamoDBReadPolicy,
   GuDynamoDBWritePolicy,
@@ -36,6 +40,10 @@ import {
  * - finally delete old template + stack
  */
 export class SecurityHQ extends GuStack {
+  private static app: AppIdentity = {
+    app: 'security-hq',
+  };
+
   migratedFromCloudFormation = true;
 
   constructor(scope: App, id: string, props: GuStackProps) {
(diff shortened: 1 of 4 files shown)


cdk/lib/security-hq.ts
```diff
@@ -118,6 +120,14 @@ dpkg -i /tmp/installer.deb`,
       },
     });
 
+    // TODO reduce TTL and point to new ALB after going live.
+    const oldElb = template.getResource('LoadBalancer') as CfnLoadBalancer;
+    new GuCname(this, 'security-hq.gutools.co.uk', {
+      name: 'security-hq.gutools.co.uk',
```
akash1810, line 126: Although we don't have a CODE stack, I think it's safest to make this stage aware, otherwise if/when we try to create a CODE stack, it'll fail immediately.

cdk/lib/security-hq.ts
```diff
@@ -118,6 +121,23 @@ dpkg -i /tmp/installer.deb`,
...
+    this.withStageDependentValue({
+      variableName: 'domain',
+      stageValues: {
+        [Stage.CODE]: 'security-hq.code.dev-gutools.co.uk',
+        [Stage.PROD]: 'security-hq.gutools.co.uk',
+      },
+    });
+
+    new GuCname(this, 'security-hq.gutools.co.uk', {
+      name: this.mappings.findInMap(this.stage, 'domainName'),
```
akash1810, line 136: `withStageDependentValue` returns a `T` (`string` in this case), so we can simplify this a little ([inspiration](https://github.com/guardian/cdk/blob/624f45762ec875a905c7d8759166ce0f20e747fd/src/constructs/dns/dns-records.ts#L72-L79)):

```typescript
const domainName = this.withStageDependentValue({
  variableName: 'domain'Name,
  stageValues: {
    [Stage.CODE]: 'security-hq.code.dev-gutools.co.uk',
    [Stage.PROD]: 'security-hq.gutools.co.uk',
  },
});

new GuCname(this, 'security-hq.gutools.co.uk', {
  name: domainName,
  ttl: Duration.hours(1),
  resourceRecord: oldElb.attrDnsName,
});
```

This might suggest that `withStageDependentValue`'s docs need refining to improve clarity.

Looking at the snapshot diff, I think we can re-use the `domainName` mapping value too, rather than create a new `domain` mapping.

jacobwinch: Alternatively if you upgrade the `@guardian/cdk` version you can use the updated version of this construct: https://github.com/guardian/cdk/pull/912#issuecomment-968930718.

nicl: @akash1810 ah nice, I did wonder if something like this was possible!

Re: re-using domainName (from ec2/certificate), I considered this but was a bit wary, as thought it could be confusing/a bit magical.

nicl: (Though realise, I'd inadvertently done this by `findInMap(this.stage, 'domainName'),`, which was an accident.)

nicl: @jacobwinch nice! That is a lot cleaner.

cdk/lib/security-hq.ts
```diff
@@ -118,6 +128,16 @@ dpkg -i /tmp/installer.deb`,
       },
     });
 
+    // TODO reduce TTL and point to new ALB after going live.
+    const oldElb = template.getResource('LoadBalancer') as CfnLoadBalancer;
+
+    new GuCname(this, 'security-hq.gutools.co.uk', {
+      app: 'security-hq',
```
jacobwinch, line 135: Minor, I think we can use this here too?

```suggestion
      app: SecurityHQ.app.app,
```

jacobwinch (approved):
Nice work 💯 🚢

PR #338, opened 2021-11-18
Drop TTL during migration

nicl:
## What does this change?

(What it says.)

## What is the value of this?

Required during CDK migration to reduce risk as we switch DNS to the new stack.


akash1810 (approved):

akash1810 (approved):

nicl:
thanks @akash1810 !

PR #343, opened 2021-11-18
Commit synthed template for VPC stack

nicl:
Step **3** of the VPC fix. Follows:

* 1 https://github.com/guardian/security-hq/pull/341
* 2 https://github.com/guardian/security-hq/pull/342
* 3 <-- THIS PR
* 4 revert step number 2
* 5 revert step number 1

**NOTE: old VPC will need to be deleted manually first.**

## What does this change?

Recreates VPC and makes synth for VPC manual.

This is required because the stack needs to be 'environment-aware'
to generate the right number of subnets and associated resources.
We don't have a way to provide credentials for this in CI (yet) so
need to do it manually for now.

https://docs.aws.amazon.com/cdk/latest/guide/environments.html

## What is the value of this?

Corrects VPC setup.

### DIFF
--- a/cdk/bin/security-vpc.ts
+++ b/cdk/bin/security-vpc.ts
@@ -1,0 +1,14 @@
+import 'source-map-support/register';
+import { App } from '@aws-cdk/core';
+import { SecurityVpc } from '../lib/security-vpc';
+
+// Note, this stack is synthed locally rather than in CI as it is
+// 'environment-aware'
+// (https://docs.aws.amazon.com/cdk/latest/guide/environments.html) and so
+// requires AWS crendentials. Ideally, longer-term we'd support this in Github
+// Actions somehow.
+const app = new App();
+new SecurityVpc(app, 'security-vpc', {
+  stack: 'security',
+  env: { region: 'eu-west-1', account: process.env.CDK_DEFAULT_ACCOUNT }, // Crucial to ensure VPC uses all AZs.
+});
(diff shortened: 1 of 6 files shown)


jacobwinch (approved):

cdk/bin/security-vpc.ts
```diff
@@ -0,0 +1,14 @@
...
+
+// Note, this stack is synthed locally rather than in CI as it is
+// 'environment-aware'
+// (https://docs.aws.amazon.com/cdk/latest/guide/environments.html) and so
+// requires AWS crendentials. Ideally, longer-term we'd support this in Github
+// Actions somehow.
+const app = new App();
+new SecurityVpc(app, 'security-vpc', {
+  stack: 'security',
+  env: { region: 'eu-west-1', account: process.env.CDK_DEFAULT_ACCOUNT }, // Crucial to ensure VPC uses all AZs.
```
jacobwinch, line 13: Any idea why this needs the account too? I would have hoped that the environment would have been enough 😞

nicl: @jacobwinch unfortunately not - I was surprised by this too. To be 'environment-aware' requires both region and account number. (https://docs.aws.amazon.com/cdk/latest/guide/environments.html - I did try testing with just region too in case it worked but alas not.)

jacobwinch: That's a shame, thanks for explaining this though 👍

PR #344, opened 2021-11-18
Support dual stacks in RR

nicl:
Improves on the hacky solution in:

https://github.com/guardian/security-hq/pull/337

### DIFF
--- a/cloudformation/security-hq.template.yaml
+++ b/cloudformation/security-hq.template.yaml
@@ -373,7 +373,8 @@ Resources:
       - Ref: LoadBalancer
       Tags:
       - Key: Stage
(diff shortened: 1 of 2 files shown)


shtukas (approved):

jacobwinch:
It's probably not worth it right now because we're about to tear down the new ASG (https://github.com/guardian/security-hq/pull/342), but when we reinstate it I think we also need a special tag to get dual-ASG deployments working nicely e.g. https://github.com/guardian/amigo/pull/632/files#diff-78d894e83ca4cb48af587f94e9f8986cc0592d412273ef99246b30c644ae5098R303-R308.

PR #345, opened 2021-11-19
More unrecognised user bug fixes

Nirvikalpa108:
## What does this change?

This PR fixes two bugs.

First, a notification will only be sent if one of the access keys of the unrecognised user was enabled or the user had a password. If either of these conditions are true, SHQ has a reason to notify the aws account.

Second, SHQ no longer attempts to disable a key which is already disabled.

## What is the value of this?

These fixes mean that a notification will not be sent if no action has been taken by SHQ. Second, SHQ will not make unnecessary requests to disable an access key.

No.

No.

## Any additional notes?

No.

### DIFF
--- a/hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
+++ b/hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
@@ -8,7 +8,7 @@ import com.amazonaws.services.sns.AmazonSNSAsync
 import com.gu.anghammarad.models.{AwsAccount => TargetAccount}
 import com.gu.janus.JanusConfig
 import config.Config.{getAnghammaradSNSTopicArn, getIamUnrecognisedUserConfig}
-import model.{CronSchedule, VulnerableUser, AwsAccount => Account}
+import model.{AccessKeyEnabled, CronSchedule, VulnerableUser, AwsAccount => Account}
 import play.api.{Configuration, Logging}
 import schedule.IamMessages.FormerStaff.disabledUsersMessage
 import schedule.IamMessages.disabledUsersSubject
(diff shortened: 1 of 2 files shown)


hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
```diff
@@ -77,11 +77,15 @@ class IamUnrecognisedUserJob(
 
   private def sendNotification(accountCrd: (Account, Seq[VulnerableUser]), testMode: Boolean, topicArn: String): Attempt[Option[String]] = {
     val (account, users) = accountCrd
-    if (users.isEmpty) {
+    val usersWithAtLeastOneEnabledKeyOrHuman = users.filter( user =>
+      user.key1.keyStatus == AccessKeyEnabled ||
+        user.key2.keyStatus == AccessKeyEnabled ||
+        user.humanUser
```
jfsoul, line 83: As we worked out (eventually) yesterday, this condition will always be true because the vulnerable users passed to this function are always human _however_ because this is pretty fragile maybe we are better keeping this in, in case something changes in the future. Hopefully it'll be refactored away before a bug could be introduced, but I don't think it hurts?

jfsoul (approved):
👍

hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
Nirvikalpa108: Yes good point @jfsoul it's worth noting that we are anticipating doing a significant refactor of this job along with the old credentials job, so we will get an opportunity to write this much more elegantly soon.

PR #347, opened 2021-11-19
Revert "Removes CDK resources (temporarily)"

nicl:
Step **4** of the VPC fix. Follows:

* 1 https://github.com/guardian/security-hq/pull/341
* 2 https://github.com/guardian/security-hq/pull/342
* 3 https://github.com/guardian/security-hq/pull/343
* 4 <-- THIS PR
* 5 https://github.com/guardian/security-hq/pull/350

Reverts guardian/security-hq#342

TODO: Also adds some missing policies which are required for the SHQ tasks. Thanks @jorgeazevedo for reporting the errors here.

### DIFF
--- a/cdk/lib/security-hq.ts
+++ b/cdk/lib/security-hq.ts
@@ -26,6 +26,7 @@ import {
 import type { AppIdentity } from '@guardian/cdk/lib/constructs/core/identity';
 import { GuCname } from '@guardian/cdk/lib/constructs/dns';
 import {
+  GuAllowPolicy,
   GuDynamoDBReadPolicy,
   GuDynamoDBWritePolicy,
 } from '@guardian/cdk/lib/constructs/iam';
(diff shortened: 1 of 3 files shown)


cdk/lib/security-hq.ts
```diff
@@ -124,9 +125,41 @@ dpkg -i /tmp/installer.deb`,
           new GuDynamoDBWritePolicy(this, 'DynamoWrite', {
             tableName: cfnTable.tableName as string,
           }),
+          new GuAllowPolicy(this, 'SsmCommands', {
```
nicl, line 128: Somehow I'd missed the load of custom permissions SHQ needs, but I've tested and fairly confident I've not missed anything now.

cdk/lib/security-hq.ts
```diff
@@ -124,9 +125,41 @@ dpkg -i /tmp/installer.deb`,
...
+              'ssm:UpdateInstanceInformation',
+              'ssm:ListInstanceAssociations',
+              'ssm:DescribeInstanceProperties',
+              'ssm:DescribeDocumentParameters',
+              'ssmmessages:CreateControlChannel',
+              'ssmmessages:CreateDataChannel',
+              'ssmmessages:OpenControlChannel',
+              'ssmmessages:OpenDataChannel',
+            ],
+          }),
```
akash1810, line 146: This looks the same as the [`GuSSMRunCommandPolicy` construct](https://github.com/guardian/cdk/blob/624f45762ec875a905c7d8759166ce0f20e747fd/src/constructs/iam/policies/ssm.ts#L5). Can it be used?

akash1810 (approved):
Approving as this is a revert of a previously approved PR 😄 .

I think we can cut down the SHQ template a little, let's do that separately though.

cdk/lib/security-hq.ts
nicl: Ooh possibly. Will add.

PR #348, opened 2021-11-19
Use region provided by trusted advisor for S3 client to check encryption status

jfsoul:
## What does this change?
This attempts to use an AWS S3 client with the region provided by the Trusted Advisor report. We are only using `US-EAST-1` for all S3 requests, which meant that any buckets outside that region would be assumed to have no encryption (even though they might).

You can see this with the following bucket which has default encryption turned on and is in `EU-WEST-1`:
image

As it turns out, there also appears to be a limitation with the regions provided by Trusted Advisor - they don't seem to be provided unless the bucket is public, so this change will not fix that. I've opened a support ticket to find out more.

## What is the value of this?
The S3 report should more often show the correct default encryption status for buckets outside `US-EAST-1`. As it happens, for buckets in `US-EAST-1` whose region is not provided by Trusted Advisor, we will now be less accurate, but I'm assuming that a lot of our buckets will be `EU-WEST-1`.

No

No

## Any additional notes?
Related to #340 from which I spotted the issue.

### DIFF
--- a/hq/app/aws/support/TrustedAdvisorS3.scala
+++ b/hq/app/aws/support/TrustedAdvisorS3.scala
@@ -1,16 +1,19 @@
 package aws.support
 
+import aws.s3.S3
 import aws.support.TrustedAdvisor.{getTrustedAdvisorCheckDetails, parseTrustedAdvisorCheckResult}
 import aws.{AwsClient, AwsClients}
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.s3.AmazonS3
+import com.amazonaws.services.s3.model.{AmazonS3Exception, GetBucketEncryptionResult}
 import com.amazonaws.services.support.AWSSupportAsync
 import com.amazonaws.services.support.model.TrustedAdvisorResourceDetail
 import model._
-import utils.attempt.{Attempt, FailedAttempt}
+import utils.attempt.{Attempt, FailedAttempt, Failure}
 
 import scala.collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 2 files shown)


hq/app/aws/support/TrustedAdvisorS3.scala
```diff
@@ -31,25 +31,26 @@ object TrustedAdvisorS3 {
...
-        }
+    val attemptFindEncryptionStatus = for {
+      region <- Try(Regions.fromName(bucket.region))
+      encryptionStatus <- Try(clients.get(account, region).map(_.client.getBucketEncryption(bucket.bucketName)))
+    } yield encryptionStatus
+
+    attemptFindEncryptionStatus match {
+      // If we fail to retrieve the correct client, then assume no encryption
+      case Success(attempt) => Attempt.Async.Right(attempt.fold(_ => bucket, _ => bucket.copy(isEncrypted = true)))
+      case scala.util.Failure(_) => Attempt.Right(bucket)
```
jfsoul, line 45: This seems like a bit of a mis-use of attempt but I couldn't find a cleaner way of achieving the logic (wrapped in an attempt):
- If region is null or invalid, return `bucket`
- If client for the provided region is not found, return `bucket`
- If client is found and the request is unsuccessful return `bucket`
- Otherwise, when region, client, request all "good" return `bucket.copy(isEncrypted = true)`

jfsoul:
AWS have seemingly fixed the mentioned issue with trusted advisor now.

Nirvikalpa108 (approved):
LGTM 👍🏽

PR #355, opened 2021-11-22
Remove duplicate SSM policy included in GuEc2App

jorgeazevedo:
## What does this change?

This removes the custom SSM policy as the exact same policy is built-in to GuEc2App

## What is the value of this?

No duplicate roles in IAM

IAM_Management_Console

Yes!

No!


jfsoul (approved):
Nice spot!

PR #356, opened 2021-11-22
Make scripts generic and document `-u` 

jorgeazevedo:
## What does this change?

This makes `build`, `lint` and `test` forward any argument passed to them to the underlying `yarn` call and updates the README accordingly.

Crucially, it mentions in the README that `./script/test -u` is how to update the snapshots, which always need to be ran when making changes to the cdk.

## What is the value of this?

I didn't know how to update the snapshots! The easier solution would be to add `yarn test -u` to the documentation, but I saw an opportunity here to consolidate our cdk actions under the `script` interface, as it were. It's weird that everything else is done with these scripts, except this one important action.

One downside is that it's `--watch` instead of the more brief `-w`.

Another is that it's not standard. Even if we update the [upstream version](https://github.com/guardian/cdk-cli/blob/main/src/template/script/test), every other cdk project in existence won't support it.

Nope

Nope

## Any additional notes?

I'm open not to merge this and update the README instead, but I thought a proposed changed would be a good way to anchor the conversation.

### DIFF
--- a/cdk/README.md
+++ b/cdk/README.md
@@ -9,16 +9,17 @@ We follow the
 find useful scripts within the [`script`](./script) directory for common tasks.
 
 - `./script/setup` to install dependencies
-- `./script/build [-w]` to compile the Typescript and check for errors
-- `./script/test [-w]` wrapper to lint, and run tests
-- `./script/lint [-fix]` to lint the code using ESLint
+- `./script/build [--watch]` to compile the Typescript and check for errors
+- `./script/test [-u] [--watch]` wrapper to lint, and run tests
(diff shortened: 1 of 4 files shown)


nicl (approved):
Thanks for the fixes. I agree updating the snapshot shouldn't require calling `yarn` directly, when everything else is script-based.

akash1810:
> Thanks for the fixes. I agree updating the snapshot shouldn't require calling `yarn` directly, when everything else is script-based.

It is beginning to feel like we're getting less value from the agnostic script pattern now and it might be time to reconsider this choice?

PR #358, opened 2021-11-23
Add Notification Message logic for iam-remediation

Nirvikalpa108:
## What does this change?
This PR adds the text for the notification messages that will be sent when an old credentials requires a warning, final and remediation notification.

## What is the value of this?
SHQ sends notifications when a credential has been identified as requiring rotation so that developers are reminded to rotate the credential's access key(s). If this does not happen, SHQ will disable the credential and a notification will be sent once this has happened. This PR crafts the text for those notification messages.

No.

## Any additional notes?
No.


adamnfish (approved):
Looks great!

PR #359, opened 2021-11-23
Grant access for unrecognised user job to S3 bucket for Janus data

jfsoul:
## What does this change?
This addresses a couple of issues identified while testing the unrecognised user job in the dev-playground account.

- The instance did not have access to the S3 bucket containing the Janus data, that has now been added in a42b20b38856c63c0315ca39c996d47b5cb17d76
- The S3 client used to fetch the data was from the set of clients we use to monitor watched accounts. In this case, we are trying to perform an action specific to security-hq and the security account, so we have switched to using a specific S3 client for that. This allows the app to make a successful get-object request to the resource because the credentials chain is appropriate for the task. d551d1e5da3b27daa1a97fcaf0eba867041fec6e

## What is the value of this?
Allows the job to perform its task by accessing S3

Yes, cdk updates included. I've already added the necessary SSM parameters.

We could remove the region parameter that's no longer used.

### DIFF
--- a/cdk/lib/__snapshots__/security-hq.test.ts.snap
+++ b/cdk/lib/__snapshots__/security-hq.test.ts.snap
@@ -38,6 +38,38 @@ Object {
       "Description": "CIDR block from which access to Security HQ should be allowed",
       "Type": "String",
     },
+    "AuditDataS3BucketName": Object {
+      "Default": Object {
+        "Fn::Join": Array [
+          "",
+          Array [
+            "/security/security-hq/",
+            Object {
+              "Ref": "Stage",
+            },
+            "/audit-data-s3-bucket/name",
+          ],
+        ],
+      },
+      "Description": "Name of the S3 bucket to fetch auditable data from (e.g. Janus data)",
+      "Type": "AWS::SSM::Parameter::Value<String>",
+    },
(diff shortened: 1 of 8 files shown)


jorgeazevedo (approved):
Great work on this. If you want to keep pushing we can remove mentions to the s3 bucket from places like Config.scala and application.conf. Can also do the cleanup as a separate card/pr

jfsoul:
> If you want to keep pushing we can remove mentions to the s3 bucket from places like Config.scala and application.conf

Good idea @jorgeazevedo - I think that's worth doing here to tidy things up. I've just removed the region because that's the only item we're not using any longer. We _could_ pass through the bucket name in the user data and remove the duplication there too, but I'm not sure it's worth it as we'd still need to configure the path separately (the path configured in CDK is to a parent directory, whereas the app needs the full path to the file)

jorgeazevedo:
oh sorry @jfsoul I just realised I misspoke there. I said "remove mentions to the s3 bucket" but I meant to say "remove mentions to the s3 bucket **region**" 🤦‍♂️. As we spoke this morning there is an opportunity to pass bucket name and dynamo db table name via user data as well and have them defined in a single place, but that's a separate bit of work.

PR #360, opened 2021-11-24
Turn on unrecognised user job for allowed accounts

jfsoul:
## What does this change?
This turns on the unrecognised user job for allowed accounts. The job checks all human IAM users against Janus credentials using a `GoogleUsername` tag. Any users for which the tag exists, but there is no Janus access linked to that username, will have their access keys and password disabled.

For now, the only configured allowed account in production is dev-playground.

## What is the value of this?
Releasing the feature which will increase the likelihood that the IAM access of leavers is removed in a timely fashion.

No

Already made

## Any additional notes?
While testing this, there were a couple of things to note that still haven't been addressed:
- The job is not entirely idempotent if run more regularly than the current schedule. This is because there is an assumption that the credentials report is up to date in identifying which users are human (and therefore when a password exists that needs disabling). In fact the credentials report may be up to 4 hours out of date, however we are currently only running the job daily. If we were to run the job more often then the consequence not major: unnecessary API calls to AWS, and failure messages in the logs/alerts.
- I was unable to view the notification, as there doesn't seem to be a target configured in anghammarad for the dev-playground account. If there are issues with the notification upon further testing in the deploy-tools account then I suggest we fix forward at that point rather than hold this up.

### DIFF
--- a/hq/app/AppLoader.scala
+++ b/hq/app/AppLoader.scala
@@ -1,7 +1,7 @@
 import config.LoggingConfig
 import logging.LogConfig
 import play.api.ApplicationLoader.Context
-import play.api.{Application, ApplicationLoader}
+import play.api.{Application, ApplicationLoader, Mode}
 
 import scala.concurrent.Future
 
@@ -19,10 +19,12 @@ class AppLoader extends ApplicationLoader {
 
     val components = new AppComponents(context)
 
-    components.quartzScheduler.start()
+    if (context.environment.mode != Mode.Test) {
+      components.quartzScheduler.start()
 
-    components.applicationLifecycle.addStopHook { () =>
-      Future.successful(components.quartzScheduler.shutdown())
+      components.applicationLifecycle.addStopHook { () =>
+        Future.successful(components.quartzScheduler.shutdown())
+      }
     }
 
     components.application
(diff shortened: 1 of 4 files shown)


jorgeazevedo (approved):
Agree on the point you made re notification

jfsoul:
Blocked behind https://github.com/guardian/security-hq/pull/363.

We'll address error handling/idempotent behaviour in a separate PR.

hq/app/AppLoader.scala
```diff
@@ -19,6 +19,12 @@ class AppLoader extends ApplicationLoader {
 
     val components = new AppComponents(context)
 
+    components.quartzScheduler.start()
```
adamnfish, line 22: Should this depend on the environment, like the cache service does? I don't think we want to start this in Test, for example?

hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
```diff
@@ -47,7 +46,7 @@ class IamUnrecognisedUserJob(
       s3Object <- getS3Object(s3Client, config.janusUserBucket, config.janusDataFileKey)
       janusData = JanusConfig.load(makeFile(s3Object.mkString))
       janusUsernames = getJanusUsernames(janusData)
-      accountCredsReports = getCredsReportDisplayForAccount(allCredsReports)
+      accountCredsReports = getCredsReportDisplayForAccount(cacheService.getAllCredentials)
```
adamnfish, line 49: Better! Good catch.

adamnfish (approved):

PR #361, opened 2021-11-24
POC to use private config repo via Deploy Key

nicl:
## What does this change?

POC of using a Github [Deploy Key](https://docs.github.com/en/developers/overview/managing-deploy-keys) (really an SSH key) for read access to the @guardian/private-infrastructure-config repo.

## What is the value of this?

Illustrate the concept before we decide whether to do further work.

The nice thing about Deploy Keys is that they provide read-only access to a specific repo. PATs, the alternative here, cannot be tightly scoped in this way.

Ultimately, the suggestion is we store the deploy key as an organisation secret but to test things here I've just added it as a repository secret.

### DIFF
--- a/.github/workflows/ci-sbt.yml
+++ b/.github/workflows/ci-sbt.yml
@@ -43,6 +43,18 @@ jobs:
         with:
           cache: 'yarn'
 
+      - name: Add SSH key for accessing @guardian/private-infrastructure-config
+        run: |
+          mkdir -m 700 ~/.ssh
+          
+          ssh-keyscan -H github.com >> ~/.ssh/known_hosts
+
+          echo '${{ secrets.PRIVATE_INFRASTRUCTURE_CONFIG_DEPLOY_KEY }}' > ~/.ssh/id_ed25519
+          chmod 400 ~/.ssh/id_ed25519
+          
+          eval "$(ssh-agent)"
+          ssh-add
+
       # 1. Seed the build number with last number from TeamCity.
       #    Update `LAST_TEAMCITY_BUILD` as needed or remove entirely if changing Riff-Raff project name.
       # 2. Execute `script/ci`, a script that will compile, run tests etc.
--- a/cdk/jest.setup.js
+++ b/cdk/jest.setup.js
@@ -1,1 +1,2 @@
 jest.mock('@guardian/cdk/lib/constants/tracking-tag');
+jest.mock('@guardian/private-infrastructure-config');
--- a/cdk/package.json
+++ b/cdk/package.json
@@ -15,7 +15,7 @@
   "devDependencies": {
     "@aws-cdk/assert": "1.132.0",
     "@guardian/eslint-config-typescript": "^0.7.0",
-    "@guardian/private-infrastructure-config": "git+ssh://git@github.com:guardian/private-infrastructure-config.git",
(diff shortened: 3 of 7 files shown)


.github/workflows/ci-sbt.yml
```diff
@@ -43,6 +43,18 @@ jobs:
         with:
           cache: 'yarn'
 
+      - name: Add SSH key for accessing @guardian/private-infrastructure-config
```
nicl, line 46: If we decide this is the right approach we should either write an action for this or adopt one of the public ones that already exist.

jacobwinch (approved):

.github/workflows/ci-sbt.yml
```diff
@@ -43,6 +43,18 @@ jobs:
...
+        run: |
+          mkdir -m 700 ~/.ssh
+          
+          ssh-keyscan -H github.com >> ~/.ssh/known_hosts
+
+          echo '${{ secrets.PRIVATE_INFRASTRUCTURE_CONFIG_DEPLOY_KEY }}' > ~/.ssh/id_ed25519
```
adamnfish, line 52: I'd be tempted to follow the full instructions for this feature (including [the multiple repositories section](https://docs.github.com/en/developers/overview/managing-deploy-keys#using-multiple-repositories-on-one-server)) even though there's only one DeployKey being used here. This means giving this a more descriptive name to the key, adding some detail to `.ssh/config`, and specifying the identity when you clone the repository in the build itself.

This means being explicit about what the key is for, and explicitly using it when the repository is cloned. I think this will be much clearer to someone that arrives at this later, who might otherwise be surprised to see the clone working without any obvious authentication.

It also means we only have one approach to cloning private repositories, regardless how many need to be cloned. If we write it this way while this is simple then we have to know about two different approaches, and people won't be able to use it as a basis for more complex builds.

Let's chat about it though, no harm either way.

cdk/package.json
```diff
@@ -15,6 +15,7 @@
   "devDependencies": {
     "@aws-cdk/assert": "1.132.0",
     "@guardian/eslint-config-typescript": "^0.7.0",
+    "@guardian/private-infrastructure-config": "git+ssh://git@github.com:guardian/private-infrastructure-config.git",
```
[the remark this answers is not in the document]
nicl: Yeah, a tag would definitely be better. Will do.

.github/workflows/ci-sbt.yml
nicl: Hey thanks for the link - really helpful. Definitely keen to go for a fuller solution (re-usable action probably) that supports more than one key etc.

cdk/jest.setup.js
```diff
@@ -1 +1,2 @@
 jest.mock('@guardian/cdk/lib/constants/tracking-tag');
+jest.mock('@guardian/private-infrastructure-config');
```
akash1810, line 2: 👍🏽

.github/workflows/ci-sbt.yml
akash1810: Would suggest it form part of a [composite action](https://github.com/guardian/cdk/pull/871) and exposed as a `usePrivateInfraConfig` flag resulting in downstream repos having GHA files like:

```yaml
- uses: guardian/cdk@main
  with:
      usePrivateInfraConfig: true
```

akash1810 (commented):
This is brilliant!

PR #363, opened 2021-11-25
Revert 362 revert 359 js/unrecognised user s3 access

jfsoul:
Reverts guardian/security-hq#362 to enable the unrecognised user job. Includes a fix for the previous issue by avoiding a reference to `stage` in the cloudformation parameters.

There are other options (custom resources, [dynamic resolution](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html#dynamic-references-pattern)) for stage-specific parameters, but this is the simplest and achieves what we want.

Changes from the original implementation are in 96f251da33d70fab201f53a7c5c337ae0c3dfbab

### DIFF
--- a/cdk/lib/security-hq.ts
+++ b/cdk/lib/security-hq.ts
@@ -11,8 +11,8 @@ import {
   Peer,
 } from '@aws-cdk/aws-ec2';
 import { EmailSubscription } from '@aws-cdk/aws-sns-subscriptions';
-import { Duration, RemovalPolicy } from '@aws-cdk/core';
 import type { App, CfnElement } from '@aws-cdk/core';
+import { Duration, RemovalPolicy } from '@aws-cdk/core';
 import { AccessScope, GuApplicationPorts, GuEc2App } from '@guardian/cdk';
 import { Stage } from '@guardian/cdk/lib/constants/stage';
 import { GuAlarm } from '@guardian/cdk/lib/constructs/cloudwatch';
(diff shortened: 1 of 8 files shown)


jorgeazevedo (approved):
I follow. Both PROD and CODE now fetch the bucket name from the same SSM parameter. To enable them to use different files, the bucket path is different for each stage and computed at build time. Nifty!

cdk/lib/security-hq.ts
```diff
@@ -69,6 +71,18 @@ export class SecurityHQ extends GuStack {
...
+      this,
+      'AuditDataS3BucketName',
+      {
+        description:
+          'Name of the S3 bucket to fetch auditable data from (e.g. Janus data)',
+        default: `/${this.stack}/${SecurityHQ.app.app}/audit-data-s3-bucket/name`,
+        fromSSM: true,
+      }
+    );
+    const auditDataS3BucketPath = [this.stack, this.stage, '*'].join('/');
```
jorgeazevedo, line 84: Very minor point, but I'm wondering if could this be string interpolation like on line 80 instead of this array join?

jfsoul: Yeah, good point - I was looking at an example elsewhere when I wrote this 😆

adamnfish (approved):
Looks great (I haven't run it to check the permissions, of course)

hq/app/schedule/unrecognised/IamUnrecognisedUserJob.scala
```diff
@@ -26,7 +26,7 @@ import scala.util.control.NonFatal
 class IamUnrecognisedUserJob(
   cacheService: CacheService,
   snsClient: AmazonSNSAsync,
-  s3Clients: AwsClients[AmazonS3],
+  s3Client: AmazonS3,
```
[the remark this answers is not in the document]
jfsoul: 👍  done

PR #364, opened 2021-11-25
ci: Attempt to improve build time

akash1810:
## What does this change?

A couple of updates to the GitHub Actions CI build:
  - Cache more aggressively
  - Use native `sbt` rather than the script in the repo. The [repo's script](https://github.com/guardian/security-hq/blob/d2ddeeddb1a987cdaf9f5799d37dfa806693faa5/sbt#L39-L40) sets some memory options for the JVM which won't use all the [available memory](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) (7GB)

## What is the value of this?

A reduction in build times means we can ship that much faster. These changes appear to bring the build time below 3 minutes (from ~5 mins). Not quite on the same level as it was in TeamCity, but an improvement nonetheless.

No.

No.

## Any additional notes?

No.

### DIFF
--- a/.github/workflows/ci-sbt.yml
+++ b/.github/workflows/ci-sbt.yml
@@ -49,6 +49,9 @@ jobs:
         uses: guardian/actions-setup-node@v2.4.1
         with:
           cache: 'yarn'
+          cache-dependency-path: |
+            yarn.lock
+            cdk/yarn.lock
 
       - name: Add SSH key for accessing @guardian/private-infrastructure-config
         run: |
--- a/.github/workflows/ci-sbt.yml
+++ b/.github/workflows/ci-sbt.yml
@@ -42,7 +42,7 @@ jobs:
             ~/.ivy2/cache
             ~/.sbt
             ~/.coursier
-          key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.sbt') }}
+          key: sbt
 
       # (Respects .nvmrc)
       - name: Setup Node
--- a/.github/workflows/ci-sbt.yml
+++ b/.github/workflows/ci-sbt.yml
@@ -36,6 +36,13 @@ jobs:
         with:
           java-version: '8'
           distribution: 'adopt'
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.ivy2/cache
+            ~/.sbt
+            ~/.coursier
+          key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.sbt') }}
 
       # (Respects .nvmrc)
       - name: Setup Node
--- a/script/ci
+++ b/script/ci
@@ -7,4 +7,4 @@ set -e
     ./script/ci
 )
 
-./sbt --no-conf '; project hq; clean; compile; riffRaffAddManifest; riffRaffUpload'
+sbt "project hq" "; clean; compile; riffRaffUpload"


script/ci
```diff
@@ -7,4 +7,4 @@ set -e
     ./script/ci
 )
 
-./sbt --no-conf '; project hq; clean; compile; riffRaffAddManifest; riffRaffUpload'
\ No newline at end of file
+sbt "project hq" "; clean; compile; riffRaffUpload"
```
adamnfish, line 10: I wonder if these lines are important?

https://github.com/guardian/security-hq/blob/main/sbt#L27-L36

When I originally worked on SHQ the *build* didn't require any configuration, they were just for running it. That may not be the case anymore, if someone has added these settings? I guess we'll need to look into it. We may also need to double check that the GHA migration has handled this, assuming it is important.

.github/workflows/ci-sbt.yml
```diff
@@ -36,12 +36,22 @@ jobs:
         with:
           java-version: '8'
           distribution: 'adopt'
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.ivy2/cache
+            ~/.sbt
+            ~/.coursier
+          key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.sbt') }}
```
adamnfish, line 45: I can see this is taken out the example, but I'm interested in why we're rebuilding the cache when any `build.sbt` file has changed. In particular, why that should trigger a new cache when things like `plugins.sbt` or any other build files in `project` directories should not?

Those cache directories are designed to be used globally on a machine, so there should be no (application-level) reason to have to very our cache hits.

script/ci
adamnfish: I also don't think the semicolons are required, although I can see it wasn't you that did it this way.

Clearer (or at least more conventional) might be:

`sbt "project hq" clean compile riffRaffUpload`

akash1810: IIUC those lines are efforts to help `sbt-riffraff-artifact`, specifically providing information for the [`build.json` file](https://github.com/guardian/sbt-riffraff-artifact/blob/548fabb329c292e2eb1f0351836edd9219d48e5e/src/main/scala/com/gu/riffraff/artifact/BuildInfo.scala#L11-L16).

`sbt-riffraff-artifact` is actually [already](https://github.com/guardian/sbt-riffraff-artifact/blob/548fabb329c292e2eb1f0351836edd9219d48e5e/src/main/scala/com/gu/riffraff/artifact/BuildInfo.scala#L100-L139) reading the TC property file, so we no longer need to 🎉 .

The version of `sbt-riffraff-artifact` we're using is GHA aware too. That is, those lines can be removed entirely.

.github/workflows/ci-sbt.yml
akash1810: You've answered this already 😄 :

> this is taken out the example

A closer look suggests we don't actually need this separate cache action as [`setup-java` has direct caching support](https://github.com/actions/setup-java#caching-packages-dependencies). Will investigate more.

akash1810: > A closer look suggests we don't actually need this separate cache action as setup-java has direct caching support. Will investigate more.

`setup-java` only supports [`maven` and `gradle` caching](https://github.com/actions/setup-java/blob/d61af71edfe1ca10c4bd4bdb96f0fddc0d002029/src/cache.ts#L23-L36), so we have to manually configure for `sbt` for now.

> Those cache directories are designed to be used globally on a machine, so there should be no (application-level) reason to have to very our cache hits.

IIUC GHA runners are tightly scoped to a repository, the cache isn't shared/available outside of that scope. That is, I don't think this is necessarily a concern.

akash1810: Updated now.

script/ci
akash1810: I propose we remove the redundant TC lines in a separate PR as it's a different concern.

jfsoul:
@akash1810 @adamnfish where do we stand on this PR now?

PR #365, opened 2021-11-25
Implement IamRemediation/partitionOperationsByAllowedAccounts

jorgeazevedo:
## What does this change?

This takes a list of RemedationOperations and splits it in 2, according to list of accounts we are allowed to perform remediation operations in

No

No


adamnfish (approved):
Looks great, thank you!

PR #367, opened 2021-11-29
IAM remediation lookupCredentialId implementation

Nirvikalpa108:
## What does this change?
Adds implementation for the lookupCredentialId function.

## What is the value of this?
This function attempts to find matching access key metadata from an AWS list-access-keys API call, with the problem access key that SHQ has identified.

If this fails, a FailedAttempt is returned, because this means that the access key cannot be disabled as an access key id has not been identified.

No love.

No.

## Any additional notes?
This change includes an important edge case.

Currently we differentiate access keys by date, but what if both a user's access keys were created on exactly the same date to the second?

Yes, this is unlikely, but an edge case nevertheless. I think this would be impossible if the keys were created in the console, but could potentially be possible if the CLI were used.

This has been accounted for by returning a FailedAttempt and logging this out as an error for developers to investigate. Having written this down, I realise that it would be helpful for Security developers to be alerted if this occurs. It would be good to get thoughts on this.

A potential solution to this problem, to be implemented in the future, would be to:
- enrich the credentials report with access key ids, similar to how tags are added to the report
- add access key id to the access key case class in the model
- add access key id to a field on the `OutdatedCredential` `IamProblem` type.

### DIFF
--- a/hq/app/logic/IamRemediation.scala
+++ b/hq/app/logic/IamRemediation.scala
@@ -6,7 +6,7 @@ import model.iamremediation.{CredentialMetadata, IamUserRemediationHistory, Part
 import model.{AccessKey, AccessKeyEnabled, AwsAccount, CredentialReportDisplay, IAMUser}
 import org.joda.time.DateTime
 import play.api.Logging
-import utils.attempt.{Attempt, FailedAttempt}
+import utils.attempt.{Attempt, FailedAttempt, Failure}
 
 import scala.concurrent.ExecutionContext
 
@@ -106,7 +106,24 @@ object IamRemediation extends Logging {
     * This might fail, because it may be that no matching key exists.
     */
   def lookupCredentialId(badKeyCreationDate: DateTime, userCredentials: List[CredentialMetadata]): Attempt[CredentialMetadata] = {
-    ???
+    userCredentials.filter { credentialMetadata =>
+      credentialMetadata.creationDate.withMillisOfSecond(0) == badKeyCreationDate.withMillisOfSecond(0)
+    } match {
+      case singleMatchingKey :: Nil => Attempt.Right(singleMatchingKey)
+      case Nil => Attempt.Left(FailedAttempt(Failure(
+        "unable to identify matching access key in user's metadata",
+        s"I've made a list-access-keys AWS API call for ${userCredentials.map(_.username)}, but I have not found a matching key in the response.",
+        500
+      )))
+      case _ =>
+        // This is an edge case where both the user's access keys both have the same creation date.
+        // This should be unlikely given the Credentials Reports creation dates are defined up to the second.
+        Attempt.Left(FailedAttempt(Failure(
+        s"both of ${userCredentials.map(_.username)}'s access keys have the exact same creation date - cannot decide which one to select for disablement",
+        s"I've hit an edge case for ${userCredentials.map(_.username)}'s access keys where both have the same creation date, please investigate as I can't decide which one to disable.",
+        500
+      )))
+    }
   }
 
   def formatRemediationOperation(remediationOperation: RemediationOperation): String = {
--- a/hq/test/logic/IamRemediationTest.scala
+++ b/hq/test/logic/IamRemediationTest.scala
@@ -1,15 +1,18 @@
 package logic
 
 import config.Config
-import logic.IamRemediation.{getCredsReportDisplayForAccount, identifyAllUsersWithOutdatedCredentials, identifyUsersWithOutdatedCredentials}
-import model.{AccessKey, AccessKeyDisabled, AccessKeyEnabled, AwsAccount, CredentialReportDisplay, Green, HumanUser, MachineUser, NoKey}
-import model.iamremediation.{IamUserRemediationHistory, OutdatedCredential, RemediationOperation, Warning}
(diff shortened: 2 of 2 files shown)


hq/app/logic/IamRemediation.scala
```diff
@@ -106,7 +106,24 @@ object IamRemediation extends Logging {
     * This might fail, because it may be that no matching key exists.
     */
   def lookupCredentialId(badKeyCreationDate: DateTime, userCredentials: List[CredentialMetadata]): Attempt[CredentialMetadata] = {
-    ???
+    userCredentials.filter { credentialMetadata =>
+      credentialMetadata.creationDate.withTimeAtStartOfDay() == badKeyCreationDate.withTimeAtStartOfDay()
```
jfsoul, line 110: If I'm reading this correctly then the edge case is a little broader because the date only has to  be on the same _day_ rather than the same _second_.

I would say though that (coincidentally) if 2 keys have the same creation date, and one of them is vulnerable, then the other might be as well. I haven't fully considered whether this is always true - because we probably can't rely on it here anyway, but thought I'd mention

hq/test/logic/IamRemediationTest.scala
```diff
@@ -144,7 +145,25 @@ class IamRemediationTest extends FreeSpec with Matchers {
   }
 
   "lookupCredentialId" - {
-    "TODO" ignore {}
+    val nonMatchingAccessKey = CredentialMetadata("adam.fisher", "AKIAIOSFODNN1EXAMPLE", date.minusDays(1), CredentialActive)
+    val matchingAccessKey = CredentialMetadata("amina.adewusi", "AKIAIOSFODNN2EXAMPLE", date, CredentialActive)
+
+    "given a key creation date matches a date in the metadata, return the correct metadata" in {
+      val result = lookupCredentialId(date, List(matchingAccessKey, nonMatchingAccessKey))
+      result.value.username shouldEqual "amina.adewusi"
```
jfsoul, line 153:
```suggestion
      result.value.username shouldEqual matchingAccessKey.username
```
(minor, but slightly less fragile if for whatever reason we update the test data)

hq/app/logic/IamRemediation.scala
```diff
@@ -106,7 +106,24 @@ object IamRemediation extends Logging {
...
+    } match {
+      case singleMatchingKey :: Nil => Attempt.Right(singleMatchingKey)
+      case Nil => Attempt.Left(FailedAttempt(Failure(
```
jfsoul, line 113: It might be a little neater to define a separate type for these specific failures and then it might also be easier in the tests to check for a specific failure rather than just `isFailedAttempt`

hq/test/logic/IamRemediationTest.scala
```diff
@@ -144,7 +145,25 @@ class IamRemediationTest extends FreeSpec with Matchers {
...
+      val result = lookupCredentialId(date, List(nonMatchingAccessKey))
+      result.isFailedAttempt() shouldBe true
+    }
+    // both user's access keys sharing the exact same date is an edge case, because the creation date is accurate to the second
+    // and it's unlikely both keys would be created at exactly the same time, but could happen, especially if created using the CLI.
+    // If this happens then we won't know how to identify the keys' id, which is required to disable it, so we return a Failure.
+    "given a key creation date matches two dates in the metadata, return a failure" in {
+      val matchingAccessKey2 = CredentialMetadata("amina.adewusi", "AKIAIOSFODNN3EXAMPLE", date, CredentialActive)
+      val result = lookupCredentialId(date, List(matchingAccessKey2, matchingAccessKey))
+      result.isFailedAttempt() shouldBe true
```
jfsoul, line 165: As referred to in the other comment, testing the type of failure here is important I think - otherwise the code could incorrectly return the wrong kind of failure in the future and these tests wouldn't fail

jfsoul (commented):
Good spot with the edge case, maybe we need to give a little more thought to how to handle that. Perhaps it's best left as you've implemented until we understand how much of an edge case this is!

hq/app/logic/IamRemediation.scala
Nirvikalpa108: Let's discuss this.

hq/test/logic/IamRemediationTest.scala
Nirvikalpa108: Let's discuss this.

hq/test/logic/IamRemediationTest.scala
Nirvikalpa108: Sounds good 👍🏽

hq/app/logic/IamRemediation.scala
Nirvikalpa108: You make a good point @jfsoul. I've removed the `.withTimeAtStartOfDay` and added a test to check that if the dates do not match to the second, a failure is returned.

hq/test/logic/IamRemediationTest.scala
jfsoul: Discussed this elsewhere and decided to address this common theme later, not in this PR 👍

jfsoul (approved):
Let's merge this into main once "the big merge" has happened

Nirvikalpa108:
Closing in favour of https://github.com/guardian/security-hq/pull/372, which branches off the latest version of main given https://github.com/guardian/security-hq/pull/370.

PR #369, opened 2021-12-01
Fix broken tests

jorgeazevedo:
## What does this change?

I botched a rebase in #365 and dropped this import statement by mistake! Was not caught by CI as it's currently not running against branches/PRs.

## What is the value of this?

Tests on this branch are currently broken


adamnfish:
I'm fixing this retrospectively as part of the unification work, so no worries here!

jorgeazevedo:
works for me!

PR #370, opened 2021-12-01
Main plus iam remediation jobs

adamnfish:
## What does this change?

This changes nothing, but manually rebases the long-running IAM Remediation Job branch onto an up-to-date main, that includes our working Unrecognised Users job.

The IamRemediation branch removed a lot of code that is still required by the Unrecognised User job, so this meant re-writing some of the branch's history to make sure they fit on top of each other. The PR *does* remove the Vulnerable Users job, because that is being replaced by the new IAM Remediation work found here. We also remove any code that is unused after the Vulnerable Users is gone.

## What is the value of this?

We've brought the IAM Remediation work in line with main, so we can raise PRs as normal. It also removes a lot of unneeded code from SHQ.

## Any additional notes?

**IMPORTANT:** We will need to fix any existing open PRs that are based against the iam-remediation-jobs branch, by cherry-picking that work afresh. I can help do that!


jorgeazevedo (approved):
Nothing stood out to me, but if anything fell through the cracks it should be fairly straightforward to recover. Thanks Adam!

hq/app/logic/CredentialsReportDisplay.scala
```diff
@@ -168,6 +168,15 @@ object CredentialsReportDisplay {
...
   }
 
+  val USERNAME_TAG_KEY = "GoogleUsername"
+  def isTaggedForUnrecognisedUser(tags: List[Tag]): Boolean = {
+    tags.exists(t =>
+      t.key == USERNAME_TAG_KEY &&
+        t.value != "" &&
+        t.value.contains(".")
+    )
+  }
```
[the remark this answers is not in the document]
jfsoul: Ah I see where it moved from, nevermind

hq/test/notifications/AnghammaradNotificationsTest.scala
```diff
@@ -0,0 +1,30 @@
+package notifications
+
+import org.scalatest.{FreeSpec, Matchers}
+
+
+class AnghammaradNotificationsTest extends FreeSpec with Matchers {
+  "outdatedCredentialWarning" - {
+    "TODO" ignore {}
```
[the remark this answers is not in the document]
jfsoul: Looks like we didn't write them yet

hq/app/aws/s3/S3.scala
```diff
@@ -19,7 +19,6 @@ object S3 {
           "unable to get S3 object for the unrecognised user job",
           "I haven't been able to get the S3 object for the unrecognised user job, which contains the Janus data",
           500,
-          context = Some(e.getMessage),
```
Nirvikalpa108, line 22: This is my mistake. I can't think why I would want to remove context, but keep throwable! Sorry about this! I'll note to add this back in another PR.

hq/test/notifications/AnghammaradNotificationsTest.scala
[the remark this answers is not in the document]
adamnfish: This is the structure, which let us split out cards to fill in the gaps. Some of these are done in open PRs now, some are yet to be written. Those open PRs will be fixed and re-opened against main when this is merged.

hq/test/schedule/SendTestNotifications.scala
```diff
@@ -1,78 +0,0 @@
...
-
-/*
-  WARNING: This test suite is designed to only be run locally in order to trigger real email notifications to the
-  anghammarad.test.alerts Google group. This can be useful for viewing changes to the email format, for example.
-
-  In order to utilise this you can remove the 'DoNotDiscover` annotation on the class and you may
-  choose to replace 'ignore' with 'in' in the test cases below, as appropriate, and then run the tests as usual.
- */
-@DoNotDiscover
-class SendTestNotifications extends FreeSpec with OneAppPerSuiteWithComponents with Matchers with AttemptValues {
```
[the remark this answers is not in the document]
adamnfish: Yeah, I can see that this would be useful! I'm not crazy about the approach though, so let's talk about what is useful here. Generally for this kind of irregular manually triggered task we're better having a CLI endpoint to trigger it somewhere, rather than putting it in our tests.

hq/app/logic/CredentialsReportDisplay.scala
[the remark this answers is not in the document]
Nirvikalpa108: Yes you're right @jfsoul, but the location of this code has changed. It used to be in `IamUnrecognisedUsers.scala`.

hq/app/aws/s3/S3.scala
adamnfish: No, this shouldn't be here and is something I removed early on in the Remediation branch work. `context` has a special meaning in Failure, in that it can tell you the location that caused the error.

For example,
If we're taking a form submission from the user and they've given an incorrect field, the context can be used to identify which field was broken. This is important if we need to line up error message on the client. If the submission is invalid and we get several errors, the context lets us put the error message next to the right input box.
It's not just for UIs though. As another example, if we're trying to deserialise a JSON string, the context of a failure might tell you that the problem was in the `date` field.

The exception's message doesn't give "context" in this sense. It's just the description of the failure, and that should be in the message itself, or included in the log entry directly as part of the (optional) `throwable` field.

Let's chat about it a little more, but no need to restore this.

hq/app/logic/VulnerableAccessKeys.scala
```diff
@@ -1,33 +1,42 @@
 package logic
 
 import config.Config.{iamHumanUserRotationCadence, iamMachineUserRotationCadence}
-import model.{AccessKey, AccessKeyEnabled, VulnerableAccessKey}
+import model.{AccessKey, AccessKeyEnabled}
 
 object VulnerableAccessKeys {
```
Nirvikalpa108, line 6: Only a couple of methods in this object are used and they're used for dashboard in `CredentialReportDisplay`. If we don't remove this unused code now, let's make sure to do it later.

hq/app/aws/s3/S3.scala
Nirvikalpa108: Makes sense. Thanks @adamnfish! 👍🏽

hq/app/logic/VulnerableAccessKeys.scala
```diff
@@ -1,33 +1,42 @@
...
-
+  /**
+    * Flags active outdated keys for human users. Disabled keys are ignored.
+    */
   def hasOutdatedHumanKey(keys: List[AccessKey]): Boolean =
```
Nirvikalpa108, line 10: Are we planning to remove `hasOutdatedHumanKey` and `hasOutdatedMachineKey`? I can't see where they're used.

hq/test/logic/VulnerableAccessKeysTest.scala
```diff
@@ -0,0 +1,80 @@
...
+
+class VulnerableAccessKeysTest extends FreeSpec with Matchers {
+  def accessKey(daysOld: Int, active: Boolean) = {
+    AccessKey(
+      if (active) AccessKeyEnabled else AccessKeyDisabled,
+      Some(DateTime.now().minusDays(daysOld))
+    )
+  }
+
+  "hasOutdatedHumanKey" - {
```
Nirvikalpa108, line 17: I wonder if the `hasOutdatedHumanKey` and the `hasOutdatedMachineKey` tests are required? I couldn't see where these methods were being used.

hq/test/notifications/AnghammaradNotificationsTest.scala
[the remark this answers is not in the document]
Nirvikalpa108: You're right @jfsoul I didn't write any tests for these methods,  because all they do is string interpolation. Here's the offending PR: https://github.com/guardian/security-hq/pull/358. 

I really should have deleted this test file but must have forgotten! I think this file is not required.

Nirvikalpa108 (approved):
Nice one @adamnfish 💪🏽💯👏🏽😀

hq/app/logic/VulnerableAccessKeys.scala
adamnfish: Yeah, if these aren't used then we can definitely get rid of them as well. I have a feeling they were still used when I did the original prune of unused code, but it looks like that's changed now 👍 

We can do another pass of tidying up straight after this, or as part of tidying up the unrecognised user job, whatever suits us best.

hq/test/logic/VulnerableAccessKeysTest.scala
adamnfish: For sure, if the functions themselves aren't used we can remove these at the same time 👍

PR #371, opened 2021-12-01
IAM Remediation: calculateOutstandingOperations

Nirvikalpa108:
## What does this change?
Adds tests and implementations for the logic required by the IAM Remediation job to determine which operations, if any,
are required for old credentials.

## What is the value of this?
Determines which work SHQ should do on vulnerable access keys.

No.

## Any additional notes?
I have made some changes to the alert cadence and I would like to check that everyone in the team is happy with this. 

Also, I want to check what we would like to do in the case that an access key is enabled, vulnerable and its most recent operation type is `Remediation`. 

Ideally this case should never happen because all vulnerable credentials would be disabled if denoted as ready for Remediation, but in the case that it does I suggest the key is flagged for remediation again so that SHQ can attempt to disable it again in case something has gone wrong with the automated process when the CRON runs. This means that hopefully the next time the CRON runs the key would be disabled. It would be great to get thoughts on this.

### DIFF
--- a/hq/app/logic/IamRemediation.scala
+++ b/hq/app/logic/IamRemediation.scala
@@ -1,10 +1,10 @@
 package logic
 
 import config.Config
+import config.Config.{daysBetweenFinalNotificationAndRemediation, daysBetweenWarningAndFinalNotification}
 import db.IamRemediationDb
-import model.{CredentialMetadata, IamUserRemediationHistory, PartitionedRemediationOperations, RemediationOperation}
-import model.{AccessKey, AccessKeyEnabled, AwsAccount, CredentialReportDisplay, IAMUser}
-import org.joda.time.DateTime
+import model._
+import org.joda.time.{DateTime, Days}
 import play.api.Logging
 import utils.attempt.{Attempt, FailedAttempt}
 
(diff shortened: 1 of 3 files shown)


hq/app/logic/IamRemediation.scala
```diff
@@ -79,14 +83,76 @@ object IamRemediation extends Logging {
...
+          // the required operation is a FinalWarning.
+          Some(RemediationOperation(userRemediationHistory, FinalWarning, mostRecentActivity.iamProblem, mostRecentActivity.problemCreationDate))
+        case FinalWarning if now.isAfter(mostRecentActivity.dateNotificationSent.plusDays(daysBetweenFinalNotificationAndRemediation - 1)) =>
+          // If the most recent activity is a FinalWarning and the last notification was sent at least `Config.daysBetweenFinalNotificationAndRemediation` ago,
+          // the required operation is Remediation.
+          Some(RemediationOperation(userRemediationHistory, Remediation, mostRecentActivity.iamProblem, mostRecentActivity.problemCreationDate))
+        case Remediation =>
+          val name = userRemediationHistory.iamUser.username
+          val account = userRemediationHistory.awsAccount.name
+          logger.warn(s"$name in $account has an access key of recent activity type Remediation, but the key is enabled. Please investigate as I will continue to attempt key disablement until rotated.")
```
adamnfish, line 150: Great message!

hq/app/logic/IamRemediation.scala
```diff
@@ -45,19 +45,23 @@ object IamRemediation extends Logging {
...
-      key.lastRotated.exists { date =>
-        // using minus 1 so that we return true if the last rotated date is exactly on the cadence date
-        date.isBefore(now.minusDays(Config.iamMachineUserRotationCadence.toInt - 1))
-      } && key.keyStatus == AccessKeyEnabled
-    }
+  private def hasOutdatedMachineKey(keys: List[AccessKey], now: DateTime): Boolean = keys.exists(isOutdatedMachineKey(_, now))
+
+  private def isOutdatedHumanKey(key: AccessKey, now: DateTime): Boolean = {
+    key.lastRotated.exists { date =>
+      // using minus 1 so that we return true if the last rotated date is exactly on the cadence date
```
adamnfish, line 54: These datetimes have second resolution, so this might not quite behave as expected. It's not a big deal though, if everything happens one day earlier or later than expected for a specific key it's still fine. Let's consider if we want to be more precise later on, we're absolutely fine here for now!

adamnfish (approved):
Looks fantastic, thank you!

PR #372, opened 2021-12-01
IAM remediation lookupCredentialId implementation

Nirvikalpa108:
## What does this change?
Adds implementation for the lookupCredentialId function.

## What is the value of this?
This function attempts to find matching access key metadata from an AWS list-access-keys API call, with the problem access key that SHQ has identified.

If this fails, a FailedAttempt is returned, because this means that the access key cannot be disabled as an access key id has not been identified.

No love.

No.

## Any additional notes?
This change includes an important edge case.

Currently we differentiate access keys by date, but what if both a user's access keys were created on exactly the same date to the second?

Yes, this is unlikely, but an edge case nevertheless. I think this would be impossible if the keys were created in the console, but could potentially be possible if the CLI were used.

This has been accounted for by returning a FailedAttempt and logging this out as an error for developers to investigate. Having written this down, I realise that it would be helpful for Security developers to be alerted if this occurs. It would be good to get thoughts on this.

A potential solution to this problem, to be implemented in the future, would be to:
- enrich the credentials report with access key ids, similar to how tags are added to the report
- add access key id to the access key case class in the model
- add access key id to a field on the `OutdatedCredential` `IamProblem` type.

### DIFF
--- a/hq/app/logic/IamRemediation.scala
+++ b/hq/app/logic/IamRemediation.scala
@@ -6,7 +6,7 @@ import db.IamRemediationDb
 import model._
 import org.joda.time.{DateTime, Days}
 import play.api.Logging
-import utils.attempt.{Attempt, FailedAttempt}
+import utils.attempt.{Attempt, FailedAttempt, Failure}
 
 import scala.concurrent.ExecutionContext
 
@@ -172,7 +172,26 @@ object IamRemediation extends Logging {
     * This might fail, because it may be that no matching key exists.
     */
   def lookupCredentialId(badKeyCreationDate: DateTime, userCredentials: List[CredentialMetadata]): Attempt[CredentialMetadata] = {
-    ???
+    val username = userCredentials.map(_.username).headOption.getOrElse("unknown username")
+    userCredentials.filter { credentialMetadata =>
+      credentialMetadata.creationDate.withMillisOfSecond(0) == badKeyCreationDate.withMillisOfSecond(0)
+    } match {
+      case singleMatchingKey :: Nil => Attempt.Right(singleMatchingKey)
+      case Nil =>
+        Attempt.Left(FailedAttempt(Failure(
+          "unable to identify matching access key in user's metadata",
+          s"I've made a list-access-keys AWS API call for $username, but I have not found a matching key in the response.",
+          500
+        )))
+      case _ =>
+        // This is an edge case where both the user's access keys both have the same creation date.
+        // This should be unlikely given the Credentials Reports creation dates are defined up to the second.
+        Attempt.Left(FailedAttempt(Failure(
+          s"both of $username's access keys have the exact same creation date - cannot decide which one to select for disablement",
+          s"I've hit an edge case for $username's access keys where both have the same creation date, please investigate as I can't decide which one to disable.",
+          500
+        )))
+    }
   }
 
   def formatRemediationOperation(remediationOperation: RemediationOperation): String = {
--- a/hq/test/logic/IamRemediationTest.scala
+++ b/hq/test/logic/IamRemediationTest.scala
@@ -7,10 +7,11 @@ import model._
 import org.joda.time.DateTime
 import org.scalatest.Inside.inside
 import org.scalatest.{FreeSpec, Matchers, OptionValues}
-import utils.attempt.{FailedAttempt, Failure}
+import utils.attempt.{AttemptValues, FailedAttempt, Failure}
 
+import scala.concurrent.ExecutionContext.Implicits.global
 
-class IamRemediationTest extends FreeSpec with Matchers with OptionValues {
+class IamRemediationTest extends FreeSpec with Matchers with OptionValues with AttemptValues {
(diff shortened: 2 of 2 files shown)


hq/app/logic/IamRemediation.scala
```diff
@@ -106,7 +106,24 @@ object IamRemediation extends Logging {
     * This might fail, because it may be that no matching key exists.
     */
   def lookupCredentialId(badKeyCreationDate: DateTime, userCredentials: List[CredentialMetadata]): Attempt[CredentialMetadata] = {
-    ???
+    userCredentials.filter { credentialMetadata =>
+      credentialMetadata.creationDate.withMillisOfSecond(0) == badKeyCreationDate.withMillisOfSecond(0)
```
adamnfish, line 110: Great idea to explicitly set the millis here, it nicely documents what we've learned about the source data 👍

hq/app/logic/IamRemediation.scala
```diff
@@ -106,7 +106,24 @@ object IamRemediation extends Logging {
...
+    } match {
+      case singleMatchingKey :: Nil => Attempt.Right(singleMatchingKey)
+      case Nil => Attempt.Left(FailedAttempt(Failure(
```
adamnfish, line 113: I love these messages!

Bear in mind the userCredentials field is a list, so I'd expect this to write the username out multiple times into these fields. Perhaps we can get a single username once at the top of this function so we can use that in these failure strings?

Very minor point, this is great.

hq/app/logic/IamRemediation.scala
```diff
@@ -106,7 +106,24 @@ object IamRemediation extends Logging {
...
+        "unable to identify matching access key in user's metadata",
+        s"I've made a list-access-keys AWS API call for ${userCredentials.map(_.username)}, but I have not found a matching key in the response.",
+        500
+      )))
+      case _ =>
+        // This is an edge case where both the user's access keys both have the same creation date.
+        // This should be unlikely given the Credentials Reports creation dates are defined up to the second.
+        Attempt.Left(FailedAttempt(Failure(
+        s"both of ${userCredentials.map(_.username)}'s access keys have the exact same creation date - cannot decide which one to select for disablement",
```
adamnfish, line 122: It looks like the indentation is a bit off here, can we indent lines 122 - 125? The arguments to Failure and the closing parens are one chunk too far left. No harm, though.

adamnfish: True above as well, we want the body of a case statement to be indented, generally.

hq/test/logic/IamRemediationTest.scala
```diff
@@ -144,7 +145,39 @@ class IamRemediationTest extends FreeSpec with Matchers {
...
+    // and it's unlikely both keys would be created at exactly the same time, but could happen, especially if created using the CLI.
+    // If this happens then we won't know how to identify the keys' id, which is required to disable it, so we return a Failure.
+    "given a key creation date matches two dates in the metadata, return a failure" in {
+      val result = lookupCredentialId(date, List(matchingAccessKey2, matchingAccessKey))
+      result.isFailedAttempt() shouldBe true 
+    }
+    // AWS's credential's report shows that a key's last rotated date is recorded to the second,
+    // so this function must seek to match to the dates up to the second, but not the millisecond.
+    "given a key creation date matches to the minute, but not the second, return a failure" in {
+      val keyCreationDate = new DateTime(1,1,1,1,1,1)
```
adamnfish, line 170: Careful about using DateTimes with really funny years, like this! It'll be fine here (I think) but depending how the date gets used / represented as it flows through the program, this could break things! If the Date is ever converted to a (32-bit) timestamp, it won't be possible to represent the year `1` and this will either get truncated to the first of January 1970, or something even more surprising than that will happen!

Best to use a "realistic" date, even for test data (just like the one at the top of the file). Any time in the decades around now is fine.

hq/test/logic/IamRemediationTest.scala
```diff
@@ -144,7 +145,39 @@ class IamRemediationTest extends FreeSpec with Matchers {
...
+      val metaDataCreationDate = new DateTime(1,1,1,1,1,2)
+      val result = lookupCredentialId(keyCreationDate, List(matchingAccessKey.copy(creationDate = metaDataCreationDate)))
+      result.isFailedAttempt() shouldBe true
+    }
+    "given a key creation date matches to the second, but not the millisecond, return the metadata" in {
```
adamnfish, line 175: I think I'd consider saying a little more here in the test message? We know the input data only has second resolution, so we're testing an edge case where the milliseconds unexpectedly don't match. I don't feel strongly here though, see what you think!

adamnfish (approved):
Looks great! There are a couple of little things you might choose to tweak, but see how feel.

Thank you!

hq/test/logic/IamRemediationTest.scala
Nirvikalpa108: Sorry, this was a mistake! Thanks!

PR #373, opened 2021-12-03
Unrecognised user notification handling improvements

jfsoul:
## What does this change?
On testing of the unrecognised user notification in production, 2 things became apparent:
1) The notification was not being sent. The logs indicated an attempt was made by the program to put a message on the necessary Anghammarad SNS topic, but the lambda was not invoked, indicating the topic was not published to.
2) There was no indication in the SHQ logs that this failure happened.

To address these this PR does the following:
1) Adds the necessary Anghammarad SNS publish policy to the instance role via CDK. It looks like this was omitted when we migrated to CDK, but it wasn't important until now.
2) Updates the error handling in the part of the job that interprets the result of a notification `send`. We were previously `fold`ing an `Attempt` (to a `Future`) and then transforming back to an `Attempt`. If I understand correctly the `fold` is transforming the `Left` of our attempted notification send into a `None`, and `None` is not treated as a failure case, so the error is swallowed. Instead we let the `Attempt` from the notification sending bubble up to the edge of the program. 

## What is the value of this?
Notifications from this job (and others in the future) will send successfully (or fail with a visible error log)

CDK update included

No

## Any additional notes?
No

### DIFF
--- a/cdk/lib/__snapshots__/security-hq.test.ts.snap
+++ b/cdk/lib/__snapshots__/security-hq.test.ts.snap
@@ -34,6 +34,11 @@ Object {
       "Description": "Amazon Machine Image ID for the app security-hq. Use this in conjunction with AMIgo to keep AMIs up to date.",
       "Type": "AWS::EC2::Image::Id",
     },
+    "AnghammaradSnsArn": Object {
+      "Default": "/account/services/anghammarad.topic.arn",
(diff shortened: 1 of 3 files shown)


Nirvikalpa108 (approved):
Thank you for fixing all the bugs in getting this feature into PROD @jfsoul!

PR #378, opened 2021-12-07
Update DB schema for IAM Remediation jobs

Nirvikalpa108:
## What does this change?
The refactor of the outdated credentials job has changed the database structure. It now includes a range key as well as a primary key, which this PR adds.

## What is the value of this?
The database is now up to date with the refactored IAM remediation job.

Yes. 

No.

## Any additional notes?
Before this PR is deployed, the current db table should be deleted. This should be safe as the data in it has only been used for test purposes only and it's not a problem if we send another email about the same credential if one has already been sent in the past.

### DIFF
--- a/cdk/lib/security-hq.ts
+++ b/cdk/lib/security-hq.ts
@@ -55,6 +55,10 @@ export class SecurityHQ extends GuStack {
         name: 'id',
         type: AttributeType.STRING,
       },
+      sortKey: {
+        name: 'dateNotificationSent',
+        type: AttributeType.NUMBER,
+      },
     });
     const defaultChild = table.node.defaultChild as unknown as CfnElement;
     defaultChild.overrideLogicalId('SecurityHqIamDynamoTable');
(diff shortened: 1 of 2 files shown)


cdk/lib/security-hq.ts
```diff
@@ -55,6 +55,10 @@ export class SecurityHQ extends GuStack {
         name: 'id',
         type: AttributeType.STRING,
       },
+      sortKey: {
+        name: 'dateNotificationSent',
+        type: AttributeType.STRING,
```
jfsoul, line 60: Should this be `.NUMBER`?

jfsoul (commented):
Maybe we should rename the table just to keep the separation clear, and deal with the clearing up of the old data separately?

cdk/lib/security-hq.ts
Nirvikalpa108: Nice one @jfsoul thanks for catching this 👍🏽

jfsoul (approved):

Nirvikalpa108:
> Maybe we should rename the table just to keep the separation clear, and deal with the clearing up of the old data separately?

I'll put a note on this PR that the old table needs deleting before this change gets deployed. Thanks Jon 👍🏽

Nirvikalpa108:
The old db table has been deleted so going to merge this PR now.

PR #380, opened 2021-12-08
Use DynamoDbLocal for local application development

jorgeazevedo:
## What does this change?

This PR makes security hq's dynamo db client connect to `localhost:8000` instead of aws when running locally. To that effect this PR patches the sbt configuration to have `DynamoDbLocal` automatically started, and introduces an initialisation script for the database table and the directory that will hold it between applications runs.

No

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,14 +55,15 @@ There are two other files that you will need to run security HQ locally. These a
 1.  `~/.gu/security-hq.local.conf` (Please update the value of key `GOOGLE_SERVICE_ACCOUNT_CERT_PATH`)
 1. `~/.gu/security-hq-service-account-cert.json`
 
-Both of these files can be found in the Security AWS account in S3 here `s3://security-dist/security/PROD/security-hq/`.
+Both of these files can be found in the Security AWS account in S3 here `s3://security-dist/security/DEV/security-hq/`.
 
 To access the Security AWS account you will need to raise a PR in [Janus](https://github.com/guardian/janus/blob/main/guData/src/main/scala/com/gu/janus/data/Access.scala) 
 to get access to the Security account (`Security.dev` should be appropriate).
 
 Once you have Janus credentials for the AWS Security account, you can copy the files from S3 by using the following:
 ```
- aws s3 cp s3://security-dist/security/PROD/security-hq/security-hq.local.conf ~/.gu --profile security
+ aws s3 cp s3://security-dist/security/DEV/security-hq/security-hq.local.conf ~/.gu --profile security
+ aws s3 cp s3://security-dist/security/DEV/security-hq/security-hq-service-account-cert.json ~/.gu --profile security
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -138,6 +138,21 @@ From the root of the project:
 
 Once the sever has started, the webapp is accessible at [https://security-hq.local.dev-gutools.co.uk/](https://security-hq.local.dev-gutools.co.uk/)
 
+### DynamoDB local setup
+
+
```
adamnfish, line 143: I'd maybe delete the doubled up newlines here and below? No harm though!

adamnfish (approved):
Great!

README.md
jorgeazevedo: Ah, I missed that, good catch!

script/setup
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# Creates the directory where DynamoDbLocal will store its state as a file
+# Creates a table with the the right schema
+
+mkdir -p ~/.gu/security-hq
+
+aws dynamodb create-table --table-name security-hq-iam-TEST \
```
jfsoul, line 7: shall we name it `-DEV`? to distinguish it from any tables we might create as part of an integration test project?

README.md
```diff
@@ -138,6 +138,19 @@ From the root of the project:
...
+Running Security HQ locally requires a local instance of DynamoDb to connect
+to. The `run` tasks handles most of the work automatically by starting and
+stopping `DynamoDBLocal` as appropriate. It does not, however, create a table
+for the application to use.
+
+To do that, the following script needs to be executed _while_ the application
+is running. This only needs to be done once, as the database is stored on disk
+and persistent between application runs.
+
+1. `$ ./script/setup`
```
jfsoul, line 152: I do find this workflow a little unusual, but we can review again once we've taken a look at integration tests and/or other ways to run dynamo locally.

script/setup
jorgeazevedo: Yes, great point. I forgot about the table name situation. The application picks up the table name from the  `IAM_DYNAMO_TABLE_NAME` in config, which we all currently set to production. I'll add a note to the README

README.md
jorgeazevedo: Yes, I agree. I think the way this has been setup allows us to iterate on the approach fairly easily.

script/setup
jorgeazevedo: What do you think of using just `security-hq-iam`? Given it's the name of the table in prod now and the value of `IAM_DYNAMO_TABLE_NAME`, it would simplify the setup a bit.

jfsoul: Is it adding a little bit of risk whereby if someone accidentally sets their stage to `PROD` locally (or there's a coding error) then we're immediately talking to the real table?

jorgeazevedo: Oh it totally does, great point!

PR #384, opened 2021-12-16
Upgrade SBT to 1.5.7 to protect from log4j exploit

Nirvikalpa108:
## What does this change?
Due to the log4j exploit, this PR upgrades sbt to 1.5.7 where log4j has been updated to a safe version. 

## What is the value of this?

This means that it should not be possible for SHQ to be compromised using log4j in sbt. This possibility is very much an edge case, but this change seems prudent nevertheless.

No.

No.

## Any additional notes?
The sbt upgrade also required a play upgrade, which triggered binary incompatibility warnings with scala-java8-compat. This has been fixed using a `libraryDependencySchemes` setting as described here: https://github.com/guardian/status-app/pull/94/files/23687b0a44a0b0ad3afad39a19d903b4b22e6160#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R17.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,21 +3,26 @@ import com.gu.riffraff.artifact.RiffRaffArtifact.autoImport._
 import com.typesafe.sbt.packager.archetypes.systemloader.ServerLoader.Systemd
 import play.sbt.PlayImport.PlayKeys._
 import sbt.Keys.libraryDependencies
+import sbt.util
 
 import scala.concurrent.duration.DurationInt
 
 // common settings (apply to all projects)
-organization in ThisBuild := "com.gu"
-version in ThisBuild := "0.2.0"
-scalaVersion in ThisBuild := "2.12.15"
-scalacOptions in ThisBuild ++= Seq("-deprecation", "-feature", "-unchecked", "-target:jvm-1.8", "-Xfatal-warnings")
+ThisBuild / organization := "com.gu"
+ThisBuild / version := "0.2.0"
+ThisBuild / scalaVersion := "2.12.15"
+ThisBuild / scalacOptions ++= Seq("-deprecation", "-feature", "-unchecked", "-Xfatal-warnings")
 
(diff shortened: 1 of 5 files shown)


project/plugins.sbt
```diff
@@ -7,7 +7,7 @@ addSbtPlugin("com.localytics" % "sbt-dynamodb" % "2.0.2")
 libraryDependencies += "org.vafer" % "jdeb" % "1.3" artifacts Artifact("jdeb", "jar", "jar")
 
 // The Play plugin
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.1")
+addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.11")
 
 // web plugins
```
jfsoul, line 12: Maybe we can remove the dependency tree plugin now? (below)

build.sbt
```diff
@@ -3,22 +3,27 @@ import com.gu.riffraff.artifact.RiffRaffArtifact.autoImport._
...
 import sbt.Keys.libraryDependencies
+import sbt.util
 
 import scala.concurrent.duration.DurationInt
 
 // common settings (apply to all projects)
-organization in ThisBuild := "com.gu"
-version in ThisBuild := "0.2.0"
-scalaVersion in ThisBuild := "2.12.15"
-scalacOptions in ThisBuild ++= Seq("-deprecation", "-feature", "-unchecked", "-target:jvm-1.8", "-Xfatal-warnings")
```
jfsoul, line 13: Did I accidentally remove the target jvm argument?

jfsoul (approved):

build.sbt
Nirvikalpa108: My mistake! I've re-added now, thanks!

Nirvikalpa108: No, I did this on purpose and I found it's required for the project to compile.

jfsoul: Ah, it works for me with the flag still there. I wonder if it's something to do with your local java version. Happy to pair on it?

PR #385, opened 2022-01-04
Treat missing login profile as successful password removal

jfsoul:
## What does this change?
This adds error recovery during attempts to remove a login profile (i.e. password) from an IAM user. Currently we handle errors from AWS by transforming them into more helpful `FailedAttempt` types, but generally there hasn't yet been a need to recover from an error by treating it as a successful attempt. However, we don't actually care if a password we try to remove no longer exists (very possible due to the current 4 hour lag in source data from the credentials report), so in this instance we treat it as a success.

## What is the value of this?
If the jobs that utilise this feature are run more frequently, they will not fail when attempting to remove passwords multiple times.

No

No

## Any additional notes?
If interested, there were alternative, more generic, approaches explored in #381 and #382. Since we're only needing to do error recovery in one place at the moment, this is probably the better approach to avoid unnecessary abstraction.

### DIFF
--- a/hq/app/schedule/vulnerable/IamRemovePassword.scala
+++ b/hq/app/schedule/vulnerable/IamRemovePassword.scala
@@ -1,17 +1,17 @@
 package schedule.vulnerable
 
-import aws.AwsAsyncHandler.{awsToScala, handleAWSErrs}
-import aws.AwsClients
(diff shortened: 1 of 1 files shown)


Nirvikalpa108 (approved):
Looks great! Thanks Jon!

PR #387, opened 2022-01-11
Identify human users without MFA

jfsoul:
## What does this change?
Adds an implementation and tests for the method we use to identify human users (those with passwords) but with no MFA enabled.

## What is the value of this?
Starts filling in the implementation of a job to disable users who have a password without MFA enabled

No

No

## Any additional notes?
No


hq/test/logic/IamRemediationTest.scala
```diff
@@ -98,6 +98,26 @@ class IamRemediationTest extends FreeSpec with Matchers with OptionValues with A
     }
   }
 
+  "identifyUsersWithPasswordNoMFA" - {
+    "given a user with a password and No MFA, should return that user" in {
```
akash1810, line 102: minor: could this be `given a human user with a password and No MFA, should return that user` for consistency with `given a machine user...` below?

akash1810 (approved):

PR #391, opened 2022-01-17
Schedule and instantiate IamRemediationService

Nirvikalpa108:
## What does this change?
This PR adds scheduling and instantiates the `IamRemediationService` so that it will be triggered when the app starts and will run every 5 hours.

In addition, it adds an additional config property; a list of AWS account Ids for which the remediation service is run on. Also, this PR changes the cadence of the remediation service, so that it can be tested more easily. 

## What is the value of this?
The additional config value means that the feature can be rolled out more incrementally, because not all accounts are ready for this service to be turned on.

No.

Yes! Here is the new config property, `ACCOUNT_IDS_FOR_IAM_REMEDIATION_SERVICE`, which is a list of AWS Account Ids.

## Any additional notes?
Here is the rollout plan:
1. Add Deploy Tools and Domains to the list of account Ids in config
2. Deploy this branch to PROD and check that the expected operations take place everyday.
3. Revert the cadence change back to 7 days from the 1 day that this PR changed it to (`daysBetweenWarningAndFinalNotification` and `daysBetweenFinalNotificationAndRemediation`).
3. Add other accounts to the list in config which don't currently have any outdated credentials
4. Discuss a roadmap with teams that do have outdated credentials to determine when the service can be run in their account.

### DIFF
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -11,6 +11,7 @@ import com.amazonaws.services.sns.AmazonSNSAsyncClientBuilder
 import com.google.cloud.securitycenter.v1.{SecurityCenterClient, SecurityCenterSettings}
 import config.Config
 import controllers._
+import db.IamRemediationDb
 import filters.HstsFilter
 import model.AwsAccount
 import org.quartz.impl.StdSchedulerFactory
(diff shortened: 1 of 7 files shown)


Nirvikalpa108:
I've now updated SHQ's config (both PROD and DEV) to include `ACCOUNT_IDS_FOR_IAM_REMEDIATION_SERVICE` @jfsoul @jorgeazevedo.

jfsoul (approved):

PR #392, opened 2022-01-18
Outdated credentials job bug fixes

jfsoul:
## What does this change?
Adds a few small bug fixes to the outdated credentials job which we have been testing in the `Domains` account. Namely:
- (Main fix) the `problemCreationDate` is correctly initialised to the last rotated date of the vulnerable access key, rather than the time the job ran, which will correct both the notification message and importantly allow the job to find previous warnings
- Simplifies the filtering of the activity history and removes the day-level-granularity comparison, preferring exact equality. This should prevent mix ups between keys created on the same day.
- Small correction to the rectification date on the initial warning email
- Adjusts the conditions for when a final warning or remediation should be performed, giving at least a day's grace between stages of remediation (unless `daysBetweenWarningAndFinalNotification` is configured to be zero or less)

The first commit in this PR can be ignored for now, I will revert it before merging. It just helps with testing.

## What is the value of this?
Correctly functioning outdated credentials job.

No

No

## Any additional notes?
No

### DIFF
--- a/hq/app/logic/IamRemediation.scala
+++ b/hq/app/logic/IamRemediation.scala
@@ -105,7 +105,7 @@ object IamRemediation extends Logging {
       userRemediationHistory <- remediationHistories
       vulnerableKey <- identifyVulnerableKeys(userRemediationHistory, now)
       keyPreviousAlert = identifyMostRecentActivity(userRemediationHistory, vulnerableKey)
-      keyNextActivity <- identifyRemediationOperation(keyPreviousAlert, now, userRemediationHistory)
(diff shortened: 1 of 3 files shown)


hq/app/services/IamRemediationService.scala
```diff
@@ -220,7 +220,7 @@ class IamRemediationService(
   }
 
   if (environment.mode != Mode.Test) {
-    val disableCredentials: Observable[DateTime] = Observable.interval(5.hours)
+    val disableCredentials: Observable[DateTime] = Observable.interval(10.seconds, 5.minutes)
```
jfsoul, line 223: This isn't a permanent change

jorgeazevedo (approved):

PR #394, opened 2022-01-20
Schedule the outdated job during working hours and allow cache to populate

jfsoul:
## What does this change?
Updates the scheduling of the outdated credentials job, in 2 ways:
- Adds an explicit initial delay of 10 minutes, giving the cache service time to populate the credentials report we require for this job. We could also introduce a dependency on the cache service `Observable`, but I've left that for later. Right now the initial delay is implicitly set to 5 hours, making it hard to observe changes we deploy to the job.
- Restricts execution of the job to working hours, as it may take dangerous actions in accounts if warnings are not heeded.

## What is the value of this?
Job runs at times that are easier for us to observe, and at times where teams can respond to actions taken by it (even if that's just a warning email).

No

No

## Any additional notes?
No

### DIFF
--- a/hq/app/services/IamRemediationService.scala
+++ b/hq/app/services/IamRemediationService.scala
@@ -220,15 +220,21 @@ class IamRemediationService(
   }
 
   if (environment.mode != Mode.Test) {
-    val disableCredentials: Observable[DateTime] = Observable.interval(5.hours)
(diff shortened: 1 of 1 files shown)


jorgeazevedo (approved):
Looks sensible to me!

PR #395, opened 2022-01-21
Clarify remediation date in notification and remove extra quotes

jfsoul:
## What does this change?
Adds "...at the next opportunity" to the end of the notification messages to indicate that the remediation may not happen on that day, because the job doesn't run at the weekend currently. I've kept it generic sounding so we don't have to worry too much about changing the scheduling and keeping this message up to date.

Also removes an extra quote from the messages.


Nirvikalpa108 (approved):
This is a good improvement, thanks @jfsoul 👍🏽

PR #397, opened 2022-01-25
Deletes MFA logic due to new Service Control Policy

Nirvikalpa108:
## What does this change?
Deletes code related to the IAM **Remove Password Without MFA** job.

## What is the value of this?
There is no longer any need for an IAM Remediation job that disables human permanent credentials which are missing MFA, due to the introduction of a Service Control Policy (SCP), which does not allow access to AWS without MFA:
https://github.com/guardian/service-control-policies/pull/11

No.

No.

## Any additional notes?
No.


hq/app/schedule/IamMessages.scala
```diff
@@ -22,8 +22,8 @@ object IamMessages {
   object VulnerableCredentials {
     private def message(account: AwsAccount): String = {
       s"Please check the following permanent credentials in AWS Account ${account.name}, " +
-      s"which have been flagged as either needing to be rotated or requiring multi-factor authentication (if you're " +
-      s"already planning on doing this, please ignore this message). If this is not rectified before the deadline, " +
+      s"which have been flagged as needing to be rotated (if you're already planning on doing this," +
```
jfsoul, line 25: Hopefully we'll delete this file and move the relevant code to live with the other notifications soon!

hq/test/notifications/AnghammaradNotificationsTest.scala
```diff
@@ -1,30 +0,0 @@
...
-
-  "outdatedCredentialFinalWarning" - {
-    "TODO" ignore {}
-  }
-
-  "passwordWithoutMfaFinalWarning" - {
-    "TODO" ignore {}
-  }
-
-  "outdatedCredentialRemediation" - {
```
jfsoul, line 23: Oh I guess we missed these tests when we wrote the notifications? What is a useful thing to test here I wonder? 🤔

jfsoul (approved):

hq/test/notifications/AnghammaradNotificationsTest.scala
Nirvikalpa108: There wasn't anything useful to test about the message strings, so these tests were never completed. I think it's fine to delete this empty test file.

PR #402, opened 2022-02-03
Create per-user opt-out for Outdated Credentials job

jfsoul:
## What does this change?
Adds a tag that can be applied to IAM users which will opt them out of identification by the Outdated Credentials job. This is to deal with the (thankfully rare) scenario where an account contains a key that cannot simply be rotated e.g. it is embedded in a large number of places. This opt-out tag would allow us to enable the job on those accounts even while the team is working on rotating it. *It should be a last resort only*

The upside to this approach is that when we are able to rotate the user, all that needs to be done is to remove the tag. The downside is that this is a publicly visible key that can be used to bypass a useful security measure. If it is abused then we won't be alerted immediately, but we will start to see keys in the dashboard over 365 days old in accounts where the job is enabled. I think this should be enough to spot overuse eventually (plus we trust our colleagues!)

## What is the value of this?
Allows us to enable the job in more accounts

No

We could put the tag key in config, but we trust our colleagues so I don't see the value

## Any additional notes?
No

### DIFF
--- a/hq/app/config/Config.scala
+++ b/hq/app/config/Config.scala
@@ -18,6 +18,7 @@ import scala.util.Try
 object Config {
   val iamHumanUserRotationCadence: Long = 90
   val iamMachineUserRotationCadence: Long = 365
+  val outdatedCredentialOptOutUserTag = "SecurityHQ::OutdatedCredentialOptOut"
   val daysBetweenWarningAndFinalNotification = 7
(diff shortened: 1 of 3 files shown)


jorgeazevedo (approved):
Having this granularity available instead of having to exclude the whole account I think is a solid improvement!

PR #404, opened 2022-02-04
Update deprecated JVM heap size flags

jfsoul:
## What does this change?
Updates the deprecated JVM options `XX:MaxRAMFraction` and `XX:InitialRAMFraction` which have not been supported since Java 8 update 190 (we are on update 312). These have been replaced by the similar percentage versions I've used. I've done this in an attempt to combat some stability problems we've been observing with SHQ.

My understanding is that without these flags being recognised we are using the default heap size values, which for SHQ at the moment are:
```
> java -XX:+PrintFlagsFinal -version | grep HeapSize
    uintx InitialHeapSize                          := 130023424                           {product}
    uintx MaxHeapSize                              := 2057306112                          {product}
```
Which is about 124MB and 1962MB. With the new settings those should both go up to around 4GB. N.B. initial heap size is recommended to be set to the same value as max.

## What is the value of this?
Hopefully addresses instability of the app, but even so it's best to use settings that actually work as intended! 

No

No

## Any additional notes?
We should tell people about this deprecation because there are 48(!) usages of this flag in the guardian github org 😱


Nirvikalpa108 (approved):

jfsoul:
I've been looking at the GC logs for the last couple of hours and it looks pretty healthy memory-wise, so this may not solve our stability problems. I've paused `ReplaceUnhealthy` on the ASG, which might give us some more logs to interrogate on Monday.

jfsoul:
I'm less convinced that these flags were actually _removed_ in update 191 or if the replacements were just enabled from then on...

PR #406, opened 2022-02-09
Disable snyk dashboard

jorgeazevedo:
## What does this change?

This leaves the UI itself up, but disables the fetching of snyk data in CacheService, and displays a deliberate error message.

Do you think we can get away with this @jfsoul ? I'm being a little basic here, wasn't sure where to draw the line between doing almost nothing and dismantling the whole thing, like routes templates models etc

Security_HQ

## What is the value of this?

Snyk is going to decommission the API we need soon, so this is getting ahead of that

No

No

### DIFF
--- a/hq/app/utils/attempt/Failure.scala
+++ b/hq/app/utils/attempt/Failure.scala
@@ -64,6 +64,12 @@ object Failure {
     Failure(details, friendlyMessage, 500)
   }
 
+  def snykDashboardDecomission(cacheContent: String): Failure = {
+    val details = s"Cache service error; unable to retrieve $cacheContent"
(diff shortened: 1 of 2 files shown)


hq/app/utils/attempt/Failure.scala
```diff
@@ -64,6 +64,12 @@ object Failure {
     Failure(details, friendlyMessage, 500)
   }
 
+  def snykDashboardDecomission(cacheContent: String): Failure = {
+    val details = s"Cache service error; unable to retrieve $cacheContent"
+    val friendlyMessage = s"This dashboard has been temporarily closed due to snyk disabling the API it relied on. Please go to snyk.io directly for information about vulnerabilities in guardian services."
```
jfsoul, line 69: This is true, although we could instead word it as `...closed while we evaluate the best format for this information`

jfsoul (approved):

PR #407, opened 2022-02-09
Use cached thread pool for most AWS client operations

jfsoul:
## What does this change?
This updates nearly all of our AWS SDK clients to use a _cached_ rather than _fixed_ thread pool. Why is that desirable? We spotted that the application was falling over after approximately 24 hours. We eventually tracked this down to an OutOfMemoryException triggered by a lack of available threads. 

After much examination of the threading behaviour, we believe this happens because we are using the default `FixedThreadPool` that AWS provides in each of our clients, which has a maximum size of 10, and does not terminate idle threads whatsoever. Due to the behaviour of our application (large batches of network requests on a predictable schedule) we were observing that the thread pools slowly filled up over time and were never emptied of idle threads. Some of these threads were re-used, but it was not enough to stay under the limit we currently have on `MaxTasks` imposed by systemd (about 4900). This sounds like an awful lot, but we are creating thousands of SDK clients due to all the regions, accounts and services we work with.

`CachedThreadPool` on the other hand has no maximum size, but importantly it terminates idle threads after 60 seconds. This means that between scheduled tasks, we are able to remove idle threads, and stay at a much lower range of threads:

image

## What is the value of this?
It will keep the app running more consistently

No

No

## Any additional notes?
Big thanks for all the help from @jorgeazevedo on this one.

### DIFF
--- a/hq/app/aws/AWS.scala
+++ b/hq/app/aws/AWS.scala
@@ -3,36 +3,36 @@ package aws
 import com.amazonaws.ClientConfiguration
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
 import com.amazonaws.auth.{AWSCredentialsProvider, AWSCredentialsProviderChain, STSAssumeRoleSessionCredentialsProvider}
-import com.amazonaws.client.builder.AwsClientBuilder
 import com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration
+import com.amazonaws.client.builder.{AwsAsyncClientBuilder, AwsClientBuilder}
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.cloudformation.{AmazonCloudFormationAsync, AmazonCloudFormationAsyncClientBuilder}
 import com.amazonaws.services.dynamodbv2.{AmazonDynamoDB, AmazonDynamoDBClientBuilder}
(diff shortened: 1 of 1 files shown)


hq/app/aws/AWS.scala
```diff
@@ -52,37 +52,40 @@ object AWS {
     } yield AwsClient(client, account, region)
   }
 
-  def dynamoDbClient(securityCredentialsProvider: AWSCredentialsProvider, region: Regions, stage: Stage): AmazonDynamoDB = {
```
jfsoul, line 55: I just moved a few methods around so the logic in this file was easier to follow (`clients` next to where it was called from, for example)

jfsoul:
This change was running in production over the weekend and the app was stable 🎉

jorgeazevedo (approved):
Great work on this, I think it's a neat and tidy solution!

jorgeazevedo:
@jfsoul before you merge this, I was thinking of deploy the version current in main to gather some data and produce an identical chart, for a sort of before/after. I figure it might come in handy! I'd leave it collecting data from say 2pm today, and then collect it tomorrow morning (hopefully before the instance gets cycled).

jfsoul:
Ok sure, I will hold off merging @jorgeazevedo

jorgeazevedo:
@jfsoul I've disabled CD and will deploy an older main build in a couple of hours. Feel free to merge!

PR #408, opened 2022-02-10
Switch to guardian org snyk workflow

jorgeazevedo:
## What does this change?

This rewrites our snyk workflow to use the new guardian org level snyk workflow rather than a custom implementation, which is in line with our current guidelines but actually doesn't work for us.

## What is the value of this?

Testing this new global snyk workflow, which has the potential to be adopted widely across the department.

Finally having reliable data about which dependencies the project uses.

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -7,26 +7,15 @@ on:
   schedule:
     - cron: '0 6 * * *'
   push:
+    branches:
+      - main
   workflow_dispatch:
 
 jobs:
   security:
-    runs-on: ubuntu-latest
-    env:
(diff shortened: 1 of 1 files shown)


jfsoul (approved):
Change looks good, but how does it look on snyk more importantly!

jorgeazevedo:
It looks good! It catches the cdk project as well. What it doesn't square with is the results dependabot finds on our package.json, which flags lots of vulnerabilities, but that is also the case for the snyk github integration, so it's not getting any worse!

Projects___Snyk

jfsoul:
>  it doesn't square with is the results dependabot finds

Oh that'll be interesting to take a closer look at

PR #412, opened 2022-02-22
Update Play Google Auth

jfsoul:
## What does this change?
Updates the version of play google auth we're using to `2.2.2` which introduces play secret rotation for the first time. To begin with this will use a static secret from SSM, and later we can introduce a rotation strategy.

## What is the value of this?
This work is a precursor to moving to scala 2.13 and to addressing some vulnerabilities identified by snyk.

No

Yes, I've already added DEV and PROD secrets to SSM.

## Any additional notes?
No

### DIFF
--- a/hq/app/AppComponents.scala
+++ b/hq/app/AppComponents.scala
@@ -3,16 +3,19 @@ import aws.{AWS, AwsClient}
 import com.amazonaws.ClientConfiguration
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
 import com.amazonaws.auth.{AWSCredentialsProviderChain, DefaultAWSCredentialsProviderChain}
+import com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration
 import com.amazonaws.regions.Regions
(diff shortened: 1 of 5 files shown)


hq/app/AppComponents.scala
```diff
@@ -71,28 +74,48 @@ class AppComponents(context: Context)
...
+      AmazonDynamoDBClientBuilder.standard()
+        .withCredentials(securityCredentialsProvider)
+        .withRegion(Config.region)
+        .build()
+    case DEV =>
+      AmazonDynamoDBClientBuilder.standard()
+        .withCredentials(securityCredentialsProvider)
+        .withEndpointConfiguration(new EndpointConfiguration("http://localhost:8000", Config.region.name))
+        .build()
+  }
```
jfsoul, line 110: I've moved a few things around to avoid forward references and to group conceptually similar things together. I've also moved single-client initialisations out of the separate `AWS` object, as I don't think the indirection there was adding much value and there are already initialisations done inline in this file.

PR #413, opened 2022-02-22
Move to Java 11

jfsoul:
## What does this change?
Updates to Java 11, which the Guardian's Play Secret Rotation library ultimately depends on (via https://github.com/guardian/play-secret-rotation/blob/main/build.sbt#L18) on https://github.com/ben-manes/caffeine/releases/tag/v3.0.0.


jacobwinch (approved):
It's probably worth mentioning this in the `README` too, as we currently have some apps which still use Java 8 and others which require Java 11.

PR #414, opened 2022-02-23
Refactor use (and lack of use) of IamClient for consistency

jfsoul:
## What does this change?
This refactor aims to address some inconsistencies and duplication in the functionality that removes passwords and disables access keys in AWS IAM. Namely:
- Disabling of access keys was done in 2 places
- Password removals were not using the IamClient at all
- Existing methods had inconsistent signatures

We aim for the IamClient to be a fairly thin wrapper around the AWS SDK, with the application specifics of how those are used housed elsewhere. We could probably go further than this does, but that can come later and is less justified anyway.

There were a couple of other minor changes included, like removing a bit more unrelated duplication, and some moving of methods.

## What is the value of this?
Clearer separation of application logic from AWS SDK interaction and less duplication of functionality.

No

No

## Any additional notes?
I'll need to do some testing of both jobs before this can be released.

### DIFF
--- a/hq/app/logging/Cloudwatch.scala
+++ b/hq/app/logging/Cloudwatch.scala
@@ -76,8 +76,8 @@ object Cloudwatch extends Logging {
     putMetric(defaultNamespace, "Vulnerabilities", Seq(("GcpProject", project),("DataType", dataType.toString)), value)
   }
 
-  def putIamRemovePasswordMetric(reaperExecutionStatus: ReaperExecutionStatus.Value): Unit = {
-    putMetric(defaultNamespace, "IamRemovePassword", Seq(("ReaperExecutionStatus", reaperExecutionStatus.toString)), 1)
+  def putIamRemovePasswordMetric(reaperExecutionStatus: ReaperExecutionStatus.Value, value: Int = 1): Unit = {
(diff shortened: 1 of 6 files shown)


hq/app/logging/Cloudwatch.scala
```diff
@@ -76,8 +76,8 @@ object Cloudwatch extends Logging {
     putMetric(defaultNamespace, "Vulnerabilities", Seq(("GcpProject", project),("DataType", dataType.toString)), value)
   }
 
-  def putIamRemovePasswordMetric(reaperExecutionStatus: ReaperExecutionStatus.Value): Unit = {
-    putMetric(defaultNamespace, "IamRemovePassword", Seq(("ReaperExecutionStatus", reaperExecutionStatus.toString)), 1)
+  def putIamRemovePasswordMetric(reaperExecutionStatus: ReaperExecutionStatus.Value, value: Int = 1): Unit = {
```
jacobwinch, line 79: Minor: as far as I can tell this is only used in a couple of places (and value is different both times), so I probably wouldn't use a default parameter value here.

jacobwinch (approved):
Looks good to me 👍

PR #418, opened 2022-02-28
Consolidate functionality that lists IAM user access keys

jfsoul:
## What does this change?
This PR is primarily aimed at removing duplication in code that lists IAM user access keys. That currently exists because the "unrecognised user" job does not use the functionality in the IAM client.

In removing the duplicate code entirely from `IamListAccessKeys` we are also able to bring some of the deeply-nested functions of the job closer to the edge (the outer for comprehension of the job). This makes it clearer in a single place all of the actions that the job performs. In consolidating method calls, we have also consolidated types, so there are some that are no longer needed (e.g. `VulnerableAccessKey`)

There are a couple of new types added to the model, which are not strictly needed, but are for convenience when working with tuples like `(Account, Users)` or `(Account, Keys)`.

This PR is probably best reviewed commit by commit.

## What is the value of this?
Less duplication, cleaner and clearer code.

No

No

## Any additional notes?
No


hq/app/logic/IamUnrecognisedUsers.scala
```diff
@@ -49,10 +47,13 @@ object IamUnrecognisedUsers extends Logging {
...
-    val unrecognisedUsers = accountCredsReports.map { case (acc, crd) => (acc, filterUnrecognisedIamUsers(crd.humanUsers, janusUsernames)) }
-    val unrecognisedUsersInAllowedAccounts = filterAllowedAccounts(unrecognisedUsers, allowedAccountIds)
-    unrecognisedUsersInAllowedAccounts.filter { case (_, users) => users.nonEmpty } // remove accounts with an empty list of users
+  ): List[AccountUnrecognisedUsers] = {
+    accountCredsReports.map {
+      case (acc, crd) => AccountUnrecognisedUsers(acc, filterUnrecognisedIamUsers(crd.humanUsers, janusUsernames))
+    }.filter { accountUsers =>
+      accountUsers.unrecognisedUsers.nonEmpty && // remove accounts with an empty list of users
+        allowedAccountIds.contains(accountUsers.account.id)
+    }
```
jfsoul, line 56: Changed the return type to our convenience type `AccountUnrecognisedUsers` instead of a list of tuples. Also inlines the filtering logic to remove the unnecessary misdirection of `filterAllowedAccounts`.

hq/app/logic/IamUnrecognisedUsers.scala
```diff
@@ -112,39 +112,23 @@ object IamUnrecognisedUsers extends Logging {
...
         logger.error(s"failed to delete at least one password: ${failure.logMessage}")
         Cloudwatch.putIamRemovePasswordMetric(ReaperExecutionStatus.failure, 1)
       case Right(success) =>
-        logger.info(s"passwords deleted for ${vulnerableUsers.map(_.username).mkString(",")}")
+        logger.info(s"passwords deleted for ${accountUnrecognisedUsers.unrecognisedUsers.map(_.username).mkString(",")}")
         Cloudwatch.putIamRemovePasswordMetric(ReaperExecutionStatus.success, success.flatten.length)
     }
   }
 
-  def disableAccountUsers(
```
jfsoul, line 131: this method is no longer needed because we do this work at the edge of the job in a for comprehension

hq/app/services/IamRemediationService.scala
```diff
@@ -96,12 +96,21 @@ class IamRemediationService(
...
       config <- getIamUnrecognisedUserConfig(config)
+      // fetch and parse our stored Janus config to use the canonical source of "recognised" usernames
       s3Object <- getS3Object(securityS3Client, config.janusUserBucket, config.janusDataFileKey)
       janusData = JanusConfig.load(makeFile(s3Object.mkString))
       janusUsernames = getJanusUsernames(janusData)
+      // look up the credentials report from the cache service as our source of current IAM users
       accountCredsReports = getCredsReportDisplayForAccount(cacheService.getAllCredentials)
+      // determine the unrecognised users by comparing Janus usernames to the IAM users (and filter to allowed accounts)
       allowedAccountsUnrecognisedUsers = unrecognisedUsersForAllowedAccounts(accountCredsReports, janusUsernames, config.allowedAccounts)
-      _ <- Attempt.traverse(allowedAccountsUnrecognisedUsers)(disableAccountUsers(_, iamClients))
```
jfsoul, line 104: instead of deeply nesting listing, disabling keys and removing passwords - we orchestrate it all from here

nicl (approved):
Not too familiar with the subject matter, but the changes seem sensible to me. Comments are around readability really - not specific to this PR, but combining several functional transformations in a row can be difficult to read.

jorgeazevedo (approved):

hq/app/logic/IamUnrecognisedUsers.scala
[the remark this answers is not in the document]
jfsoul: Nice suggestion, I blindly stuck with what was there and just chained it, but this is better.

PR #419, opened 2022-02-28
Report all error statuses to cloudwatch for IAM

jfsoul:
## What does this change?
We're removing the restriction on which types of IAM credential statuses get reported to cloudwatch (for consumption in our external dashboards.) We are currently excluding outdated credentials from being included in reporting, but since that change we have been working with teams to clean up accounts. We are also not currently actively using the external dashboard as a benchmark for progress, so a sudden increase in critical vulnerabilities reported there is acceptable.

## What is the value of this?
A step toward better visibility of IAM problems, with an eye on introducing an SLO in this area.

No

No

## Any additional notes?
No


jorgeazevedo (approved):
Agreed, thanks Jon!

PR #420, opened 2022-03-02
Upgrade test dependencies for scala 2.13 compatibility

jfsoul:
## What does this change?
Upgrades the testing-related dependencies that are currently on versions that were not built for scala 2.13.

This is not a trivial change, mainly due to the movement of entities in scalatest 3.1.0 described here: https://www.scalatest.org/release_notes/3.1.0.

I have split it into 3 commits:
1. Bump the existing dependencies in 673df1a9a7bad489822f0b8b752efa80cb07bed7
2. Autofix the renames/moves using [this procedure](https://github.com/scalatest/autofix/tree/master/3.1.x) in d4a1585fe98ddddebe51c880af07215ada2a174c. As this was automated, it can be largely ignored for review purposes.
3. Introduce scalatestplus scalacheck in 62e45598feae1423b802bc93de5d809f977737a6, which is the new separate library for combining scalatest and scalacheck. Previously this functionality was in scalatest itself.

## What is the value of this?
Prepares us for scala 2.13 and keeps deps up to date.

No

No

## Any additional notes?
I had some trouble with my IDE recognising these changes or displaying unrelated errors during development, but that might've been a local problem to me. (It appears to be ok now FWIW)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -54,9 +54,9 @@ lazy val hq = (project in file("hq"))
       "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion,
       "com.fasterxml.jackson.module" %% "jackson-module-scala" % jacksonVersion,
       "com.google.cloud" % "google-cloud-securitycenter" % "1.3.6",
-      "org.scalatestplus.play" %% "scalatestplus-play" % "3.1.2" % Test,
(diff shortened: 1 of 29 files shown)


jorgeazevedo (approved):
Was able to run the tests and everything looks ok, so I think this is good to go

PR #422, opened 2022-03-03
Update to Scala 2.13.x

jfsoul:
## What does this change?
This PR updates SHQ to Scala 2.13.x. We have already taken care of all the dependencies in previous PRs (except for Janus config tools, which is bumped here because it is not cross-compiled). Here we do the actual upgrade and address the syntax changes. The changes which impacted us (which should be split into separate commits) were:
- Auto application of `()` deprecated
- Moving of converters for Java types 
- A single pattern match was identified as non-exhaustive (technically it already was exhaustive, but maybe scala 2.13 can't tell that) 🤷‍♂️ 
- `mapValues` returning a `MapView`
- Procedure syntax deprecation i.e. `def run {` -> `def run: Unit = {`
- Deprecation of `'` tick for Symbols
- `replaceAllLiterally` deprecated
- Either is now right-biased

I think most of these were non-controversial. For the right-biased Eithers change I also identified a dependency we could remove (scalatestplus-play), which was bringing in an older version of scalatest, which didn't have the appropriate EitherValues helper.

## What is the value of this?
Keeping up to date with language version and deps, with all the benefit that usually brings

No

No

## Any additional notes?
Everything passes locally, and I've tested the web app. I have not manually tested the lambda(s), but I think those need addressing separately as they haven't been updated for a while.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,7 @@ import scala.concurrent.duration.DurationInt
 // common settings (apply to all projects)
 ThisBuild / organization := "com.gu"
 ThisBuild / version := "0.2.0"
-ThisBuild / scalaVersion := "2.12.15"
+ThisBuild / scalaVersion := "2.13.8"
 ThisBuild / scalacOptions ++= Seq("-deprecation", "-feature", "-unchecked", "-Xfatal-warnings")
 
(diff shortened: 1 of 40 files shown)


akash1810 (approved):
🎉

PR #423, opened 2022-03-10
Add workflow to test PRs for new vulnerabilities

jorgeazevedo:
## What does this change?

This workflow uses the new snyk PR workflow (see https://github.com/guardian/.github/tree/ja-snyk-pr) to test PRs that change dependencies for new vulnerabilities.

## What is the value of this?

Preventing vulnerabilities from being committed to main!

No

No

### DIFF
--- a/.github/workflows/snyk-pr.yml
+++ b/.github/workflows/snyk-pr.yml
@@ -14,5 +14,6 @@ jobs:
     uses: guardian/.github/.github/workflows/sbt-node-snyk-pr.yml@ja-snyk-pr
     with:
       ORG: guardian-security
+      SEVERITY_THRESHOLD: critical
     secrets:
        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/snyk-pr.yml
```diff
@@ -0,0 +1,18 @@
+# This action runs every day at 6 AM and on every push
+# If the branch it's running on is main then it will run snyk monitor (reports vulnerabilities to snyk.io)
+# Otherwise it will run snyk test
+name: Snyk
+
+on:
+  pull_request:
+    branches:    
+      - main
```
jfsoul, line 9: Does this restriction add much? I sometimes (initially) target PRs against other branches if they build on top of each other, before merging them in sequence - this would mean the check doesn't run at first. It's probably only a minor inconvenience.

jfsoul, line 9: Does this restriction add much? I sometimes (initially) target PRs against other branches if they build on top of each other, before merging them in sequence - this would mean the check doesn't run at first. It's probably only a minor inconvenience.

jfsoul: and as discussed it'd be good to check out a `paths` restriction to see if we can restrict to when manifest files are changed

jfsoul: and as discussed it'd be good to check out a `paths` restriction to see if we can restrict to when manifest files are changed

jorgeazevedo: Yeah, you're right I don't think there's a strong argument for this only running on PRs against main. I've removed it now

jfsoul (approved):

PR #425, opened 2022-03-21
Switch to using node 16 for cdk and frontend build

jorgeazevedo:
## What does this change?

We've tracked down some cdk CI build failures due to it using the wrong node version.

After some investigation it came to light that the cdk folder was set up with its own `.nvmrc`, the intention being that that would be picked up by the GHA.

In reality, the setup node action was ignoring it. Instead, it was picking up on the `.nvmrc` at the root of the project, which uses a different version than the one the cdk needs.

This solves the problem by setting the version of node the cdk requries in the `.nvmrc` at the project root instead. The frontend build and npm run commands appear to run successfully.

## What is the value of this?

Working builds!

No

No


jfsoul (approved):
👍 👍  This is a great spot. I'm going to try rebasing the dependabot PRs against this to see if it helps with those too.

PR #428, opened 2022-03-22
Restore/revamp Snyk dashboard with vulnerability age indicator

jfsoul:
## What does this change?
This PR restores the previously disabled Snyk dashboard. The main motivation for this is to experiment with introducing an SLO for 3rd party vulnerabilities, which requires providing relevant metrics to users.  This PR implements a basic warning in that area that an organisation contains long-lived high/critical severity vulnerabilities.

We have deviated significantly from the previous functionality/API usage.  Previously we would work on a per-project basis and make API calls for each of those (all 400+ 😱 ). Now we are utilising the reporting API at the organisation level (i.e. per team) to gather the information we need, which makes the density of requests much more manageable.

There are certainly improvements to be made still, but this PR is already large. Namely we could:
- Surface more information about specific vulnerabilities (and which ones are breaching the SLO!)
- Provide a time-series graph view of vulnerabilities and number of projects
- Preferably indicate how many projects are integrated in the recommended manner (i.e. CLI) 
- Add paging to the API usage to work around the current 1000 result limit
- Fix the generated links by using the `slug` field from the API instead of `name`
- Add some explanation or specifics regarding the SLO
-  Link more things to Snyk!

image
Redacted parts look like this, but these numbers are made up:
image
Accordion contents look like:
image

(Happy to screen share the full changes, if the redaction is too disruptive)

## What is the value of this?
We start to surface relevant metrics on age of critical/high severity vulnerabilities to teams, which will be the basis of an SLO.

No

No

## Any additional notes?
No

### DIFF
--- a/hq/app/logic/SnykDisplay.scala
+++ b/hq/app/logic/SnykDisplay.scala
@@ -59,7 +59,6 @@ object SnykDisplay extends Logging {
 
   def parseOrganisationVulnerabilitiesBody(s: String): Attempt[List[SnykProjectIssues]] =
     parseJsonToObject("project vulnerabilities", s, body => {
-      //logger.info(s"parsing project vulns for ${s}")
       val orgIssuesResult = (Json.parse(body) \ "results").validate[List[SnykProjectIssue]]
       orgIssuesResult.map(groupProjectIssues)
     })
--- a/hq/app/logic/SnykDisplay.scala
+++ b/hq/app/logic/SnykDisplay.scala
@@ -79,5 +79,23 @@ object SnykDisplay extends Logging {
     orgs.sortBy(soi => (-soi.critical, -soi.high, -soi.medium, -soi.low, soi.organisation.name))
 
   def sortProjects(projects: List[SnykProjectIssues]): List[SnykProjectIssues] =
-    projects.sortBy(spi => (-spi.high, -spi.medium, -spi.low, spi.project.get.name))
+    projects.sortBy(spi => (-spi.critical, -spi.high, -spi.medium, -spi.low, spi.project.fold("")(_.name)))
+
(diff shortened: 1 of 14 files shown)


hq/app/logic/SnykDisplay.scala
```diff
@@ -48,28 +52,49 @@ object SnykDisplay {
...
+    projectsAndIssues.groupBy(_.project).map { case (project, issues) =>
+      SnykProjectIssues(project, issues)
+    }.toList
 
-  def linkToSnykProject(snykProjectIssues: SnykProjectIssues, queryString: Option[String]): String = snykProjectIssues.project match {
-    case Some(project) if project.organisation.nonEmpty =>
-      s"https://snyk.io/org/${project.organisation.get.name}/project/${project.id}/${queryString.getOrElse("")}"
+  def linkToSnykProject(snykProjectIssues: SnykProjectIssues, org: SnykOrganisation, queryString: Option[String]): String = snykProjectIssues.project match {
+    case Some(project) =>
+      s"https://snyk.io/org/${org.name}/project/${project.id}/${queryString.getOrElse("")}"
```
jorgeazevedo, line 73: Something strange is happening with these links. It works for most cases, but not for ed tools and frontend?

jorgeazevedo (approved):
Really solid! The only issue I noticed is the one with the links, which is fairly minor.

hq/app/logic/SnykDisplay.scala
jfsoul: Sorry, missed this. I spotted this as well and put a bullet in the description. We need to use `slug` instead of `name` because you can rename orgs but the URLs are static.

I'm proposing we do this in a follow up because this PR has gotten pretty big already.

jorgeazevedo: ohh sorry, you did mention it in the description. Yep, follow up is fine!

PR #434, opened 2022-03-25
Add table containing top "old" high/critical vulnerabilities identified by Snyk

jfsoul:
## What does this change?
Adds a table to the Snyk dashboard summary of an org, which displays the top 5 oldest critical (or high if there are no critical) vulnerabilities. This helps users navigate from this dashboard to the most important-to-fix issues in Snyk when they are not meeting an SLO.

While there I've also:
- Renamed the dashboard because the name could be confused with Snyk's product for static code analysis "Snyk Code".
- Added the number of "old" vulnerabilities to the warning message t
- Fixed/simplified links by using the project url value returned by the API

image

## What is the value of this?
As above, directs users to the most important-to-fix issues.

No

No

## Any additional notes?
We could still do with adding a tooltip explaining the calculation of "old"


hq/app/views/snyk/snyk.scala.html
```diff
@@ -7,7 +7,7 @@
     <div class="hq-sub-header">
         <div class="container hq-sub-header__row">
             <div class="hq-sub-header__name">
-                <h4 class="header light grey-text text-lighten-5">Snyk Code Scanning</h4>
+                <h4 class="header light grey-text text-lighten-5">Snyk - Dependency Vulnerabilities</h4>
```
jorgeazevedo, line 10: 👍

hq/app/views/snyk/snykOrgProjects.scala.html
```diff
@@ -20,26 +21,64 @@ <h5>Top 5 vulnerable projects</h5>
...
+
 <div class="row">
     <div class="col s12">
         @defining(longLivedHighVulnerabilities(orgIssues), longLivedCriticalVulnerabilities(orgIssues)) { case (longLivedHighVulnerabilities, longLivedCriticalVulnerabilities) =>
           @if(longLivedHighVulnerabilities.size > 0 || longLivedCriticalVulnerabilities.size > 0) {
               <div class="card-panel">
                   <i class="material-icons left red-text">warning</i>
-                  <span>This organisation contains critical or high severity vulnerabilities older than the maximum permitted age</span>
+                  <span>This organisation contains @{longLivedCriticalVulnerabilities.size} critical and @{longLivedHighVulnerabilities.size} high severity vulnerabilities older than the maximum permitted age</span>
                   <a target="_blank" rel="noopener noreferrer" href="@{s"https://snyk.io/org/${orgIssues.organisation.name}/reports/issues"}">
```
jorgeazevedo, line 82: I think these links are suffering from the same issue that was causing the other ones to break for ed tools and frontend

hq/app/views/snyk/snykOrgProjects.scala.html
```diff
@@ -20,26 +21,64 @@ <h5>Top 5 vulnerable projects</h5>
...
         </table>
     </div>
 }
 
+@topOldVulnerabilities() = {
+    @defining(top5OldCriticalOrHighVulnerabilities(orgIssues)) { oldIssues =>
+        @if(oldIssues.nonEmpty) {
+            <div class="row">
+                <div class="col s12">
+                    <h5>Top 5 old critical/high vulnerabilities</h5>
```
jorgeazevedo, line 42: Minor point, but while the context for what old means here is right at the top ("older than the maximum permitted age"), at first glance I wasn't sure if this meant that it's just sorting them by age. I wonder if it would make it a bit clearer to make explicit that these are the top 5 vulns that are "breaching the SLO" or something to that effect.

jorgeazevedo (approved):
These are helpful tweaks, great work!

hq/app/views/snyk/snykOrgProjects.scala.html
jfsoul: Good spot!

hq/app/views/snyk/snykOrgProjects.scala.html
jfsoul: Yeah, it would be nice to add a detailed explainer to this dashboard like we have for IAM/SGs. Tempted to address separately?

jorgeazevedo: Yep, totally!

PR #445, opened 2022-04-19
Skip riff raff upload on dependabot PRs

jorgeazevedo:
## What does this change?

With this patch, the ci workflow only uploads builds to riff raff (and, more importantly, only has _permissions_ to upload) on PRs opened by devs. For PRs opened by dependabot, the riff raff steps are skipped.

## What is the value of this?

This will make builds for dependabot PRs green. They currently fail because they do not have access to the necessary permissions. This is due to a security measure implemented by github on march 2021, where the ci build environment offered to dependabot no longer has access to secrets to prevent malicious dependencies from syphoning them off or using them as an attack vector. See https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544

No

No

### DIFF
--- a/.github/workflows/ci-sbt.yml
+++ b/.github/workflows/ci-sbt.yml
@@ -19,12 +19,15 @@ jobs:
       # required by aws-actions/configure-aws-credentials
       id-token: write
       contents: read
+    env:
+      SHOULD_UPLOAD_TO_RIFRAFF: ${{ github.actor != 'dependabot[bot]' }}
     steps:
       - uses: actions/checkout@v2
 
       # Setup AWS credentials to enable uploading to S3 for Riff-Raff.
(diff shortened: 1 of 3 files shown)


jfsoul (approved):
Maybe add some information to the README regarding dependabot PRs? In particular I think it would be useful to highlight how to deploy and test dependabot PRs if necessary (I am assuming this would simply require a manual re-run of the action)

akash1810:
Have we documented the reasoning against using the [Dependabot secret approach](https://github.com/aws-actions/configure-aws-credentials/issues/188#issuecomment-812086459) anywhere? I think that's a more scalable/simple solution as we'd create one org secret and done, whereas in this solution, each repo needs updating.

PR #452, opened 2022-05-06
Improve java 11 story

nicl:
Improve CI build:

* use Corretto to mirror PROD
* cache to improve build times

Also adds an sdk(man) configuration file to make it easier to develop with the right Java version. (It acts equivalent to an .nvmrc and is automatically applied when cd'ing into the directory when the user has sdkman installed.)

Note, we suggest the Microsoft (openjdk-based) version of Java 11 as Corretto 11 is not availble for M1 unfortunately at the moment.

See this (WIP) doc for more context: https://docs.google.com/document/d/1ZR-YnaXCT5_gLVmTCeGs0mWd3KPaAozPjQK8uUzHZ9w/edit#.


.sdkmanrc
```diff
@@ -0,0 +1 @@
+java=11.0.14.1-ms
```
akash1810, line 1: Is this Microsoft's flavour of java? Should `Corretto-11.0.15.9.1` be used instead to match CI and PROD?

nicl: @akash1810 unfortunately I don't think AWS publish Corretto 11 for M1 (only 17+). I'll double check this though.

nicl: Update: https://docs.aws.amazon.com/corretto/latest/corretto-8-ug/downloads-list.html shows an M1-friendly version! For some reason, sdkman does not support it. I might look into raising a quick PR on sdkman to see if this can be supported.

akash1810: `asdf` supports it 😉

jfsoul (approved):
Looks good, but we can wait to resolve the asdf and corretto questions

.tool-versions
```diff
@@ -0,0 +1 @@
+java corretto-11.0.15.9.1
```
jfsoul, line 1: Should we add a quick note to the readme to use asdf?

PR #455, opened 2022-05-31
chore: Remove the VPC stack from this repository

akash1810:
## What does this change?

We're moving the stack into it's own repository - [guardian/security-platform](https://github.com/guardian/security-platform).

This is because:
  - The stack is account wide
  - The CDK definition needs some private information (account ID), so a public repository is no longer suitable


jorgeazevedo (approved):

jfsoul:
💎

PR #458, opened 2022-07-01
Update getting started docs

kenoir:
## What does this change?

Re-order steps and clarify instructions in some places.

## What is the value of this?

Make it a little easier to get started.


jfsoul (approved):
🏆

PR #484, opened 2022-07-20
Dependency updates 2022-07-20

jfsoul:
## What does this change?
Updates all scala dependencies as suggested by Scala Steward. I've only had to remove the kinesis logback appender update, because it requires AWS SDK v2.


akash1810 (approved):

jfsoul:
> Approved on the basis that CI has passed...
Thanks! I've given it a quick test locally too and everything looks good.

PR #495, opened 2022-07-26
simplify login instructions. no special link required

NovemberTang:
## What does this change?
Updates instructions for accessing snyk

## What is the value of this?
Simplifies the process, and removes the need for a special link that needs to be kept secret

no

### DIFF
--- a/hq/markdown/snyk.md
+++ b/hq/markdown/snyk.md
@@ -7,7 +7,8 @@ vulnerabilities in your application.
 
 ## Getting set up
 
-To access snyk, go to [their website](https://app.snyk.io/) and sign up via Enterprise SSO. Any other method won't work.
+To access snyk, go to [their website](https://app.snyk.io/login/sso) to sign up via Enterprise SSO. Any other method won't
+work and might have adverse consequences.
 This also may not work if you are already signed into another Snyk account, so you may need to logout first.
 
 We have an enterprise group called 'The Guardian'. Everyone will get added to the organisation `guardian-people`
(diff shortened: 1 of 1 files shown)


hq/markdown/snyk.md
```diff
@@ -7,12 +7,8 @@ vulnerabilities in your application.
 
 ## Getting set up
 
-We use Enterprise SSO to get into Snyk, tied to our office email accounts. This requires entering via the following
-special link.
-
-[%SNYK_SSO_LINK%](%SNYK_SSO_LINK%)
-
-This may not work if you are already signed into another Snyk account, so you may need to logout first.
+To access snyk, go to [their website](https://app.snyk.io/) and sign up via Enterprise SSO. Any other method won't work.
```
akash1810, line 10: I'm a bit nervous about this. If you accidentally opt to login w/GitHub, Snyk will automatically create an account for you and create a project for every repository you have access to, regardless of their parent organisation. This is 2000+ repos, if you're a member of the Guardian GH org, and you very quickly bump up against the free quota in Snyk, and PRs in the aforementioned repos will start failing because Snyk cannot run (as you're out of quota). This causes great confusion!

I say all this having logged in to Snyk this way before! It wasn't fun!

If we want to remove the deep-link, could we replace it with https://app.snyk.io/login/sso, which is the result of clocking `sign up via Enterprise SSO`/

NovemberTang: interesting, thanks for the explanation and workaround!

akash1810 (approved):

PR #675, opened 2022-11-07
docs: Simplify instructions to run locally

akash1810:
## What does this change?

Creates a script to automate previously manual steps.

Now, to run the app locally, one should:
  1. Ensure dependencies are met.
  2. Run `script/setup`.
  3. Run `script/start`.

Included in this change is a replacement of `sbt-dynamodb` with Docker. This is for two reasons:
  1. https://github.com/localytics/sbt-dynamodb is deprecated (last commit 2018)
  2. Docker is simpler. Previously, the README made reference to running SBT, then running a script just before the app starts. Now, the setup script performs all the required steps at once.

See the revised README [here](https://github.com/guardian/security-hq/blob/aa/setup/README.md).

## What is the value of this?
Simpler local DX.

No.

No.

## Any additional notes?
N/A.


kenoir (approved):
👌 this is great.

PR #783, opened 2023-01-13
Updated the new public URLs for security reports

aaronp:
## What does this change?

Updated security report links

## What is the value of this?

The links are no longer broken (point to public resources)

no

no

## Any additional notes?

no


jacobwinch (approved):
Thanks for fixing this 👍

PR #845, opened 2023-02-15
refactor: Remove root `package.json` file

akash1810:
## What does this change?
The `package.json` file at the root enables prettier and eslint on the JavaScript in the repository. It also does some Snyk checks, which have been superseded by GitHub Actions.

Whilst this is good, we're not checking these as part of the build. Additionally, we don't have a standard client-side setup in this repository, with `jquery` and `materialize` committed to VCS.

That is, in it's current state, the root `package.json` isn't providing much utility in this repository, so remove it.

## What is the value of this?
Dependency management becomes easier.


.github/workflows/ci-sbt.yml
```diff
@@ -41,11 +41,12 @@ jobs:
           distribution: 'corretto'
           cache: 'sbt'
 
-      # (Respects .nvmrc)
       - name: Setup Node
-        uses: guardian/actions-setup-node@v2.4.1
```
akash1810, line 46: This action has been deprecated, in favour of the first party `actions/setup-node`.

PR #847, opened 2023-02-15
Remove Dependabot from `dependency-updates` process

akash1810:
## What does this change?
In #445 and #552 we configured Dependabot to run weekly, and bundle changes into the `dependency-updates` branch, which gets PR'd against `main` once a month.

Whilst this process has been working, there are a couple of drawbacks:
  - The repository is set up differently from all other repositories
  - The revision history of the `dependency-updates` branch is [quite noisy](https://github.com/guardian/security-hq/pull/816), as Scala Steward is also adding commits
  - A single branch makes it difficult to isolate infrastructure changes from application changes

In this change, we configure Dependabot in the same way as other repositories:
  - Run once a month
  - Raise PRs against `main`, one PR per dependency bump

The downside of this is we loose the Riff-Raff upload protections that #445 added, but I'm not sure they are necessary as this behaviour hasn't been replicated elsewhere, even on more sensitive repositories (e.g. Janus, or Riff-Raff). It might be they were added here purely as a prototype?

## What is the value of this?
The shape of this repository matches others, making it more accessible.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,15 +6,9 @@
 version: 2
 updates:
   - package-ecosystem: "npm"
-    directory: "/"
-    schedule:
-      interval: "weekly"
-    target-branch: "dependency-updates"
-  - package-ecosystem: "npm"
     directory: "/cdk"
     schedule:
-      interval: "weekly"
-    target-branch: "dependency-updates"
+      interval: "monthly"
     ignore:
       - dependency-name: "aws-cdk"
       - dependency-name: "aws-cdk-lib"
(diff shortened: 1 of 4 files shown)


kenoir:
> Raise PRs against main, one PR per dependency bump
I wonder if the [queued merging GitHub recently released](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) would help alleviate some of the problems with having many PRs created by Dependabot.

kenoir (approved):
I'm ok with this, I'd like to enable the merge queue and see if that improves the experience of delaing with the unbatched Dependabot PRs.

PR #876, opened 2023-02-21
chore: Remove Snyk on PR workflow

akash1810:
## What does this change?
Removes the GH Workflow that runs Snyk on PR (added in #423), as we have one that runs [on push to `main`](https://github.com/guardian/security-hq/blob/48c8f07514608f4e7f7e26f62bede3709dda5b13/.github/workflows/snyk.yml) already.

This makes the repository's shape more similar to others, simplifying the DX.

## What is the value of this?
Increased consistency = easier to maintain?


kenoir (approved):
I think there may be a use case for this more widely in the future (folk would like to know about vulnerabilities before they're merged), but I don't think it's adding value in its current form.

PR #900, opened 2023-03-13
Update doc to make clear where the gathered information comes from

JuliaBrigitte:
Update doc to make clear the information displayed is only as up to date as the last run for AWS Trusted Advisor. Also put links in README to other doc available in repo, because that would have made my life easier. Some info about how to debug added too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ presents the data collected by those processes.
 It also provides an interface on some markers of a watched AWS
 account's health from a security point of view.
 
+## Trusted Advisor
+Security HQ uses information from AWS Trusted Advisor.
+This might not be as up to date as one might wish and because noticeable for S3 buckets.
+
 ## Local development
 ### Requirements
 1. Java 11. See [.tool-versions](.tool-versions) for the exact version. [asdf](https://asdf-vm.com/) is the recommended Java version manager.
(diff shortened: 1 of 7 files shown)


script/debug
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
akash1810, line 1: In the README, we've said:

```bash
./script/start --debug
```

Which implies this extra debug script isn't needed?

README.md
```diff
@@ -11,6 +11,10 @@ presents the data collected by those processes.
 It also provides an interface on some markers of a watched AWS
 account's health from a security point of view.
 
+## Trusted Advisor
+Security HQ uses information from AWS Trusted Advisor.
+This might not be as up to date as one might wish and because noticeable for S3 buckets.
```
kenoir, line 16: Can we clarify what the update cadence is, whether it's possible to refresh this data and what the restrictions are around that?

kenoir, line 16:
```suggestion
This might not be as up to date as one might wish and may be noticeable for S3 buckets.
```

JuliaBrigitte: I put the comment more it to give further onlookers a hint, not because I yet know how it works.

akash1810 (approved):
Not a blocker, can be done in a future PR: I wonder if the AWS SDK returns a last updated timestamp that we can show?

PR #996, opened 2023-06-29
Upgrade security group notifier lambda to java 11, fix build.

NovemberTang:
## What does this change?

We have an objective to upgrade our java runtimes from 8 to 11. While carrying out this upgrade we discovered that the lambda no longer built, so there are some changes to make that process work again, improve the documentation, and automate the building of the jar.

## What is the value of this?

Java 8 is deprecated, and it's important that the lambda has more recent, less vulnerable library versions in production.

**Have you committed your changes to the CloudFormation templates?**

Yes

**Has the CloudFormation or StackSet update been completed?**

Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ Security HQ holds a Lambda, which checks for security groups that are open to th
 
 It is deployed as a stack set and is defined in  `cloudformation/watched-account.template.yaml`.
 
-#### Deployment
+#### Deployment (currently manual)
 1. Update the version in `build.sbt`.
 2. Build the lambda by running `.script/package-lambda`.
 2. Upload it to S3 (ask DevX where the bucket is located).
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ account's health from a security point of view.
 
 ## Trusted Advisor
 Security HQ uses information from AWS Trusted Advisor.
-This might not be as up to date as one might wish and may be noticeable for S3 buckets.
+This might not be as up-to-date as one might wish and may be noticeable for S3 buckets.
 
 ## Local development
 ### Requirements
@@ -96,7 +96,7 @@ or 365 days for users without a password (machine users).
 It also disables permanent users who have left the Guardian.
 
 The reaper sends email notifications to the AWS account the user is in, before disabling a user.
-The emails are sent via Anghammarad and uses it's AWS Account to email address mappings.
+The emails are sent via Anghammarad and uses its AWS Account to email address mappings.
 
 You can also find the dynamo table in the Security AWS Account.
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -96,18 +96,22 @@ or 365 days for users without a password (machine users).
...
 
 It is deployed as a stack set and is defined in  `cloudformation/watched-account.template.yaml`.
-This lambda is deployed manually by creating a JAR file locally and uploading it to S3: `s3://guardian-dist/guardian/PROD/securitygroups-lambda/`. The version name is important, because
-the cloudformation has a paramter, `version`, which is used to locate the correct S3 file. The stackset is then deployed manually again to all accounts from the root account.
 
-The build and deploy process is manual at present. The lambda was deployed once in 2018 and hasn't been updated since. Health tickets have been put onto the backlog to try to improve this process.
+#### Deployment
+1. Update the version in `build.sbt`.
+2. Build the lambda by running `.script/package-lambda`.
+2. Upload it to S3 (ask DevX where the bucket is located).
```
akash1810, line 111: Could this step be done by Riff-Raff?

README.md
```diff
@@ -96,18 +96,22 @@ or 365 days for users without a password (machine users).
...
 
 ### Lambda
 Security HQ holds a Lambda, which checks for security groups that are open to the world, except ELB groups. This data is used for `https://security-hq.gutools.co.uk/security-groups`.
 
 It is deployed as a stack set and is defined in  `cloudformation/watched-account.template.yaml`.
-This lambda is deployed manually by creating a JAR file locally and uploading it to S3: `s3://guardian-dist/guardian/PROD/securitygroups-lambda/`. The version name is important, because
-the cloudformation has a paramter, `version`, which is used to locate the correct S3 file. The stackset is then deployed manually again to all accounts from the root account.
 
-The build and deploy process is manual at present. The lambda was deployed once in 2018 and hasn't been updated since. Health tickets have been put onto the backlog to try to improve this process.
+#### Deployment
```
akash1810, line 108: Worth explicitly noting that the process is currently manual?

akash1810 (approved):
LGTM! Couple of minor non-blocking comments.

README.md
NovemberTang: It could, but there is a risk of people forgetting to bump the `version`, and writing over an old version of the lambda, which would be a disaster if we needed to debug or roll back as there's no explicit link between a version and a commit. To get around this we'd need to introduce an automatic versioning system like a build number. This seemed like scope creep, so we can address it in another PR

README.md
NovemberTang: done

PR #1006, opened 2023-07-11
feat: Add Stack, Stage, App to notification message

akash1810:
## What does this change, and what is the value?
The `Stack`, `Stage`, `App`, and `gu:repo` tags can be used to identify a service, and where it is defined.

Add these to the notification message to make it easier to triage at a glance.

Here's what the current chat notification looks like:

image


jacobwinch (approved):
Great idea 💯

akash1810:
Example of message with this change:

image

PR #1007, opened 2023-07-12
fix: Display all tags on a security group 

akash1810:
## What does this change?
Improves on https://github.com/guardian/security-hq/pull/1006, to better handle those cases where the `Stack`, `Stage`, `App`, or `gu:repo` tags are not present.

Printing all the tags offers better UX, as when the specific tags are all `unknown`, the message isn't very helpful:

image


NovemberTang (approved):

akash1810:
Here's the message now:

image

https://github.com/guardian/security-hq/pull/1008 changes the markup.

PR #1008, opened 2023-07-12
fix: Bold the tag name

akash1810:
## What does this change?
A slight tweak to #1007, correcting the markdown(?) to bold the tag name, instead of italicise.

The current message italicises the tag name, which looks a bit odd!

image


jacobwinch (approved):

akash1810:
Following this change, the message is now:

image

PR #1080, opened 2023-12-14
Fix: bump logback to fix high Snyk vulnerabilities

tjsilver:
## What does this change?

Bumps logback-classic (and core) to versions that don't have high vulnerabilities. Because Play ships with a vulnerable version of logback (even Play 2.9), I have added the dependency explicitly so that it overrides Play's version.

Note: I could have updated Play to v.2.9 and sbt to v1.9.7, but it wasn't necessary to fix the vulnerabilities. Please comment if you think this should be done as part of this change too.

I don't think so. 

I don't think so.

I tested this by running `./script/ci` `sbt test` `sbt compile` and `snyk test` to ensure the high vulnerabilities had been removed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -62,6 +62,8 @@ lazy val hq = (project in file("hq"))
       "org.scalacheck" %% "scalacheck" % "1.17.0" % Test,
       "com.github.alexarchambault" %% "scalacheck-shapeless_1.15" % "1.3.0" % Test,
       "com.gu" %% "anghammarad-client" % "1.8.1",
+      "ch.qos.logback" % "logback-classic" % "1.4.14",
+
 
       // logstash-logback-encoder brings in version 2.11.0
       // exclude transitive dependency to avoid a runtime exception:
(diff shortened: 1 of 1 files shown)


akash1810 (commented):
Any reason to use two different versions of the same library?

tjsilver:
> Any reason to use two different versions of the same library?

I did this because I was keeping as close to the original versions used by com.gu and Play, as I wasn't sure what the differences in versions might cause in terms of functionality.

tjsilver:
> > Any reason to use two different versions of the same library?
> 
> I did this because I was keeping as close to the original versions used by com.gu and Play, as I wasn't sure what the differences in versions might cause in terms of functionality.



> > Any reason to use two different versions of the same library?
> 
> I did this because I was keeping as close to the original versions used by com.gu and Play, as I wasn't sure what the differences in versions might cause in terms of functionality.

I have now made both versions the same by bumping v1.2 to v1.4 @akash1810 - it looks like 1.4 supports Jakarta EE (not sure about 1.2). https://logback.qos.ch/dependencies.html

akash1810 (approved):

PR #1087, opened 2024-01-12
Make open security group notification more readable

NovemberTang:
## What does this change?

Target tags are presented as one key value pair per line.

## What is the value of this?

Easier reading

Requires stackset update, but not urgently, it can wait until next week.

No

## Any additional notes?

No

### DIFF
--- a/lambda/security-groups/src/main/scala/com/gu/hq/Notifier.scala
+++ b/lambda/security-groups/src/main/scala/com/gu/hq/Notifier.scala
@@ -40,7 +40,7 @@ object Notifier extends StrictLogging {
          |Warning: Security group **$groupId** in account **$accountName** is open to the world.
          |
          |The security group has ${targetTags.length} tags.
-         |${targetTags.map(t => s"**${t.key}**: ${t.value}").mkString(", ")}
+         |${targetTags.map(t => s"**${t.key}**: ${t.value}").mkString("\n")}
          |
          |""".stripMargin
     val targets = getTargetsFromTags(targetTags, accountId)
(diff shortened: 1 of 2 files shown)


lambda/security-groups/src/main/scala/com/gu/hq/Notifier.scala
```diff
@@ -40,7 +40,7 @@ object Notifier extends StrictLogging {
          |Warning: Security group **$groupId** in account **$accountName** is open to the world.
          |
          |The security group has ${targetTags.length} tags.
-         |${targetTags.map(t => s"**${t.key}**: ${t.value}").mkString(", ")}
+         |${targetTags.map(t => s"**${t.key}**: ${t.value}").mkString("\n")}
```
adamnfish, line 43: Does it need to be "|\n"? I feel like the answer to this should be obvious but it doesn't sound it to me this afternoon 🤔

NovemberTang: Possibly! I was mildly surprised that no tests failed as a result of this change so I'll stick one in and see what happens.

NovemberTang: Turns out it's totally happy with `\n`. `\n|` would also have been fine, but `|\n` causes the test to fail.

adamnfish (approved):
Great idea 👍

adamnfish:
It occurred to me late that this probably won't work. For emails, the raw text is used for the email message. However, for "hangouts" messages (as Google chat was called when Anghammarad was created) Anghammarad will parse the content as markdown and use that to generate the JSON datastructure for a chat "Card". Single newlines are ignored, and just treated as a single parapgraph spread over multiple lines.

https://github.com/guardian/anghammarad/blob/main/anghammarad/src/main/scala/com/gu/anghammarad/messages/Messages.scala

PR #1089, opened 2024-01-16
chore: upgrade to Play v3

tjsilver:
## What does this change?

- Upgrades Play to v3.0.1.

- Upgrades various other dependencies to enable this to work. 

- Bumps the version of Security HQ to 0.5.0. 

- One change to Scala config has been made to remove an import for `GoogleServiceAccount` that doesn't exist in the later version of Google Auth.

## What is the value of this?

This has been done to  remove our dependence on Akka. Play v3 uses Pekko instead.

I don't think so.

I have updated the local DEV config in S3 with a longer application secret, as required by Play v3. I don't  think PROD will be affected.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,11 +10,6 @@ ThisBuild / version := "0.5.0"
 ThisBuild / scalaVersion := "2.13.10"
 ThisBuild / scalacOptions ++= Seq("-deprecation", "-feature", "-unchecked", "-Xfatal-warnings")
 
-// We need to avoid a binary-incompatible version conflict - because janus-config-tools v.0.0.5 requires com.typesafe.play:twirl-api_2.13:1.4.2
-ThisBuild / libraryDependencySchemes ++= Seq(
-  "org.scala-lang.modules" %% "scala-xml" % VersionScheme.Always
-)
-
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -6,20 +6,21 @@ import scala.concurrent.duration.DurationInt
 
 // common settings (apply to all projects)
 ThisBuild / organization := "com.gu"
-ThisBuild / version := "0.4.2"
+ThisBuild / version := "0.5.0"
 ThisBuild / scalaVersion := "2.13.10"
 ThisBuild / scalacOptions ++= Seq("-deprecation", "-feature", "-unchecked", "-Xfatal-warnings")
 
+// We need to avoid a binary-incompatible version conflict - because janus-config-tools v.0.0.5 requires com.typesafe.play:twirl-api_2.13:1.4.2
```
adamnfish, line 13: Perhaps we can remove this line once we bump janus-config-tools to 0.0.6, which we can do straight after this.

adamnfish (approved):
LGTM, if it works!

adamnfish:
With the caveat that we need to check if the tests are running!

PR #1107, opened 2024-03-15
chore(deps): Update `play-googleauth` to v4.0.0

akash1810:
## What does this change?
This fixes an issue with the `locale` field in Google's response during the login process. See https://github.com/guardian/play-googleauth/pull/218.

`play-secret-rotation` is also updated to v7.1.0 due to binary compatibility.

## What is the value of this?
More reliable login experience.


adamnfish (approved):
🙏

PR #1118, opened 2024-05-01
refactor: remove GCP functionality

tjsilver:
## What does this change?

Removes the GCP functionality from the app/UI.

Ran the app locally and GCP findings and links to them are now gone.

## What is the value of this?

This functionality is currently broken so it shouldn't make any difference to users.
We are paring down Security HQ and migrating functionality to other repos where possible. This will allow us to keep this repo up-to-date more easily.

No.

Only the app config - where GCP functionality has been deleted.


hq/app/config/Config.scala
```diff
@@ -67,20 +67,6 @@ object Config {
     )
   }
 
-  def gcpSccAuthentication(implicit config: Configuration): GcpSccConfig = {
-    val gcpOrgId = requiredString(config, "gcp.orgId")
-    val gcpSccSourceId = requiredString(config, "gcp.sccSourceId")
-    GcpSccConfig(gcpOrgId, gcpSccSourceId)
-  }
-
-  def gcpCredentialsProvider(implicit config: Configuration): FixedCredentialsProvider = {
```
tjsilver, line 76: @mchv we will need to remove the credentials from the Google Developer Console - I believe you have the permissions to do this?

NovemberTang (commented):
I think as a result of this we can remove some dependencies. Namely the `google.cloud.securitycenter` one, but there may be one or two others

NovemberTang (approved):
Hooray!!!

hq/app/config/Config.scala
tjsilver: @mchv The credentials can now be deleted - thank you!

PR #1124, opened 2024-05-10
Remove Security Group functionality

marsavar:
## What does this change?

Removes the open security group functionality from security-hq.
NOTE: do not merge yet until we are sure the dashboard matches the current report.

## What is the value of this?

It's been replaced by this dashboard.

No

No

## Any additional notes?

Tested locally

image


akash1810:
Worth chatting with Ophan before removing this page in Security HQ as it looks like [they link to it](https://github.com/search?q=org%3Aguardian%20https%3A%2F%2Fsecurity-hq.gutools.co.uk%2Fsecurity-groups&type=code).

tjsilver (approved):
Can probably also remove references to `finding-header__badge` from `main.css` as it looks like it's only used for SGs.

PR #1129, opened 2024-06-04
Remove IAM credential reports controller

marsavar:
## What does this change?

As per title. This has now moved to Grafana.

## What is the value of this?

Less maintenance burden. One source of truth.

## Any additional notes?

1. I left a lot of the logic in place because it's used by the credentials reaper scheduled task, which we want to keep.
2. I've slightly changed the index controller so that the panel now links to the Grafana dashboard. 

## Tests
Ran this locally!


hq/app/views/index.scala.html
```diff
@@ -17,7 +17,7 @@ <h4 class="header light grey-text text-lighten-5">Centralised AWS Account securi
         <div class="row hp-reports">
             <div class="card">
                 <div class="card-content">
-                    <a class="hp__link" href="/iam">
+                    <a class="hp__link" href="@iamReportLink">
```
marsavar, line 20: Can possibly hardcode the URL here instead of passing it as a variable to the index controller, but I thought it was neat to have all the dashboard links as vals in the controller. There's only the IAM report link now, but more will be added soon.

hq/app/controllers/HQController.scala
```diff
@@ -13,12 +13,14 @@ import scala.concurrent.ExecutionContext
...
-  extends BaseController  with SecurityHQAuthActions {
+  extends BaseController with SecurityHQAuthActions {
 
   private val accounts = Config.getAwsAccounts(config)
+  private val iamReportLink = "https://metrics.gutools.co.uk/d/bdn97cui5rbi8f/iam-credentials-report"
 
   def index = authAction {
-    Ok(views.html.index(accounts))
+
+    Ok(views.html.index(accounts, iamReportLink))
```
tjsilver, line 23: Shall we link to the Security HQ list of dashboards instead? Then we only have to add one link (but it does obviously mean an extra click for users): https://metrics.gutools.co.uk/d/fdnykwvzsgt8gf

tjsilver (approved):
There be some CSS that can be removed? https://github.com/guardian/security-hq/blob/5d277f705f4bb6905cdf308d3454c64dbb6a3555/hq/public/stylesheets/main.css#L347

tjsilver (approved):
Nice!

PR #1132, opened 2024-06-11
Ignore some typescript and eslint bumps

tjsilver:
## What does this change?

Tells Dependabot to ignore version bumps for certain versions of `typescript` and `eslint`, so they don't get bumped and break the peer dep requirements of `@guardian/eslint-config-typescript`.

`version = '10.0.1'
peerDependencies = { eslint: '^8.56.0', tslib: '^2.6.2', typescript: '~5.3.3' }`

## What is the value of this?

Fewer Dependabot PRs that we have to fix manually, hopefully.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,12 +18,21 @@ updates:
       interval: "monthly"
     commit-message:
       prefix: "chore(deps): "
-    # The version of AWS CDK libraries must match those from @guardian/cdk.
-    # We'd never be able to update them here independently, so just ignore them.
     ignore:
+      # The version of AWS CDK libraries must match those from @guardian/cdk.
+      # We'd never be able to update them here independently, so just ignore them.
       - dependency-name: "aws-cdk"
       - dependency-name: "aws-cdk-lib"
       - dependency-name: "constructs"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -18,12 +18,21 @@ updates:
...
     ignore:
+      # The version of AWS CDK libraries must match those from @guardian/cdk.
+      # We'd never be able to update them here independently, so just ignore them.
       - dependency-name: "aws-cdk"
       - dependency-name: "aws-cdk-lib"
       - dependency-name: "constructs"
+      # The version of typescript must match the versions required by @guardian/eslint-config-typescript which
+      # currently only allows patch updates, so ignore the minor and major version bumps.
+      - dependency-name: "typescript"
+        update-types: ["version-update:semver-major", "version-update:semver-minor"]
```
akash1810, line 30: Personally, I don't think this is correct as the further behind `latest` we're tracking, the more difficult it becomes to perform a "big bang" update.

We're ignoring some AWS CDK dependencies above, however these are [updated on a schedule, on the 10th of every month](https://github.com/guardian/cdk/blob/main/.github/workflows/update-aws-cdk.yaml).

I'd be keen to chat with @guardian/client-side-infra to see if we can commit to a schedule of updates.

PR #1164, opened 2024-10-15
Explicitly install sbt in dependency submission workflow

tjsilver:
## What does this change?

- Installs sbt in the dependency submission workflow.

## What is the value of this?

The workflow is currently broken because the latest Ubuntu doesn't come with sbt bundled.

## How to test

This is the workflow run [before](https://github.com/guardian/security-hq/actions/runs/11209895261) and [after ](https://github.com/guardian/security-hq/actions/runs/11347765650)this change.

- [x] TODO: remove this branch from the workflow

### DIFF
--- a/.github/workflows/dependency-graph.yml
+++ b/.github/workflows/dependency-graph.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - main
-      - ts/fix-dependency-submission-workflow
   workflow_dispatch:
 jobs:
   dependency-graph:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-graph.yml
```diff
@@ -1,14 +1,24 @@
-name: Update Dependency Graph
+name: Update Dependency Graph for SBT
 on:
   push:
     branches:
-      - main # default branch of the project
+      - main
+  workflow_dispatch:
```
NovemberTang, line 6: non-blocking: was this left in intentionally?

NovemberTang (approved):

.github/workflows/dependency-graph.yml
tjsilver: No, it just matches what dep graph integrator pr generator produces now (i.e. without the comment)

PR #1183, opened 2024-12-05
feat: remove security groups lambda

tjsilver:
## What does this change?

Removes the security groups lambda and all code relating to lambdas as this is the only remaining lambda.

## What is the value of this?

We are deprecating Security HQ, and the team has concluded that this lambda is not providing enough value to justify its continued existence and maintenance cost.

Yes.

## Tests

- [x] Run security hq app locally
- [ ] Redeploy stackset and check that IAM remediation functionality is still working


build.sbt
```diff
@@ -112,48 +112,12 @@ lazy val hq = (project in file("hq"))
...
-    name := """securitygroups-lambda""",
-    assembly / assemblyJarName := s"${name.value}-${version.value}.jar",
-    libraryDependencies ++= Seq(
-      "com.gu" %% "anghammarad-client" % "4.0.0"
-    )
-)
-
 lazy val root = (project in file(".")).
-  aggregate(hq, lambdaCommon, lambdaSecurityGroups).
+  aggregate(hq).
```
adamnfish, line 120: At the point where there's only one project the `root` / `hq` split is mostly just indirection and obfuscation! 

Nice to keep this PR focused, and since we're planning to split out the imprtant features later anyway, 100% fine to leave as-is for now 👍

hq/public/javascripts/app.js
```diff
@@ -66,16 +66,6 @@ $(document).ready(function() {
     });
   });
 
-  // Extra interactions for the dropdowns on the Security Groups page
```
adamnfish, line 69: The security groups page is separate to the security groups lambda, but I guess no harm doing this here!

The security groups page was a SHQ display of the state of all our security groups (now duplicated [here](https://metrics.gutools.co.uk/d/ddkehpibugpogd/security-groups?orgId=1)). The security-groups Lambda was the automated real-time alerting that gets sent whenever a security group gets opened up to the world.

In retrospect, the naming of these two concepts wanted some more thought!

hq/app/model/models.scala
```diff
@@ -18,14 +18,6 @@ case class AwsStack(
   region: String
 )
 
-case class StackResource(
```
adamnfish, line 21: Great catch!

cloudformation/watched-account.template.yaml
```diff
@@ -50,18 +38,6 @@ Resources:
             Resource: "*"
             Action:
             - s3:GetEncryptionConfiguration
-            # Analyse security groups
```
adamnfish, line 53: These changes are to the role SHQ assumes for watching other AWS accounts. This is separate to the Lambdas that run in each account to alert on configuration changes.

I think this ought to be a valid change now that we've removed the dashboards that required these permissions, but I think I'd want to double check that the actual services behind those dashboards in SHQ have been disabled. At first glance this doesn't seem to be the case, which means that SHQ would be trying (and failing) to read this data after it starts.

My suggestion would be for us to leave this policy as-is here, and tackle a wider change that removes deprecated functionality from SHQ (with a corresponding reducuction in its policy) as another piece of work.

adamnfish (changes requested):
Mostly LGTM, but I think we need to leave the "watched account access policy" as-is for now, ahead of actually disabling the data collection in the Security HQ app itself.

adamnfish (approved):
LGTM! Thank you

PR #1192, opened 2025-01-21
Improve error logging on IAM credentials report

adamnfish:
We add a cause to unexpected AWS SDK client errors, and use this newly-populated field (if present) from the IAM credentials report refresh code.

This should give us insights into the current failures.


tjsilver (approved):
👍

PR #1193, opened 2025-01-22
Improve error handling / logging

adamnfish:
Log messages on Failures now include a summary of the underlying cause, if present. This should improve error visibility across the board in Security HQ.

We also add additional error handling to the cache service. On every cache update we include a summary of which accounts have failures in their cache data.

This is motivated by the IAM remediation issues, and will give us more information for digging into the problem.


tjsilver (approved):
Paired on this and changes make sense.

PR #1194, opened 2025-01-22
fix: role assumption in production

tjsilver:
## What does this change?

Since the v2 SDK migration, we have been incorrectly authenticatinng the STS Assume Role Credentials Provider used by security-hq when integrating with other AWS accounts.

Instead of using the named Janus account profile (which will not exist in production), we omit the argument in order to use the default provider chain.


adamnfish (approved):
LGTM

PR #1196, opened 2025-01-30
Fixes detection of report generation after the v2 SDK update

adamnfish:
Since the AWS v2 Java SDK update, the test of whether the report has finished being generated was broken. We correct this here so that we will detect a finished report.

This will stop us getting rate limit errors, and restore IAM remediation functionality.

## What does this change?

The v1 SDK used a String for the status field, but in v2 this is an enum called `ReportType`. We had a type error warning, but because this projefct does not have fatal warnings enabled it was easy to miss. This incorrect equality check means we will be constantly retrying the report generation check instead of proceeding to load the report.


tjsilver (approved):
Good spot!!

PR #1205, opened 2025-02-03
fix: delete deprecated snyk workflow

tjsilver:
## What does this change?

Deletes the Snyk workflow.

## What is the value of this?

Snyk has been deprecated and the workflow fails.


kelvin-chappell (approved):
👍

PR #1213, opened 2025-03-04
Remove unused code that depends on the v1 AWS SDK

adamnfish:
## What does this change?

Remove unused code that depends on the v1 AWS SDK

This dead code was previously still compiling because the v1 AWS Java SDK was being brought in as a transitive dependency via the janus-config-tools library.

With the bump to v2 of janus-config-tools this is no longer the case, which highlighted this problem. Removing the unused code fixes the issue.

## What is the value of this?

This lets us update to the latest version of the janus-config-tools library, removes unused code, and eliminates an implicit dependency on the end-of-life v1 AWS Java SDK.

Closes https://github.com/guardian/security-hq/pull/1212


kelvin-chappell (approved):
👍

PR #1221, opened 2025-03-28
Bump janus-config-tools

adamnfish:
## What does this change?

This version bump adds a new feature to the config tools library, which means we need to update the library to support new config files. Otherwise, this doesn't affect how Security HQ is consuming the parsed file.

## What is the value of this?

We'll continue to be able to parse the Janus config file.

## Any additional notes?

See the following related PRs:

- https://github.com/guardian/janus-app/pull/539
- https://github.com/guardian/janus/pull/4697
- https://github.com/guardian/janus/pull/4698


kelvin-chappell (approved):
👍

PR #1261, opened 2025-08-20
fix: Increase resource signal timeout for ASG

kelvin-chappell:
## What does this change?
This change does a small thing.
Specifically, it increases the timeout for the addition of the single instance to the stack's ASG from 3 minutes to 5 minutes.

## What is the value of this?
We're seeing Cloudformation deployments failing occasionally because the instance is timing out before it's fully healthy.     
But repeating the deploy always works, which makes us think we aren't giving the instance long enough to sort itself out.
For example, in this [failed deploy](https://riffraff.gutools.co.uk/deployment/view/973e39e3-34b0-411a-af02-c520f6f1223a):
> [11:45:39] New instance(s) added to autoscaling group - Waiting on 1 resource signal(s) with a timeout of PT3M.
> [11:48:36] AutoScalingGroupSecurityhqASG8DD277A5 (AWS::AutoScaling::AutoScalingGroup): UPDATE_IN_PROGRESS
> [11:48:36] Failed to receive 1 resource signal(s) for the current batch. Each resource signal timeout is counted as a FAILURE.
> [11:48:41] AutoScalingGroupSecurityhqASG8DD277A5 (AWS::AutoScaling::AutoScalingGroup): UPDATE_FAILED
> [11:48:41] Received 0 SUCCESS signal(s) out of 1. Unable to satisfy 100% MinSuccessfulInstancesPercent requirement
> [11:48:41] security-hq-PROD (AWS::CloudFormation::Stack): UPDATE_ROLLBACK_IN_PROGRESS

## Any additional notes?
Perhaps this is a naive change but we've decided it isn't worth spending a lot of time on it as this stack will soon be redundant.

`npm run format` has been run over the files so there are some formatting changes.
The only substantive change is 
```
cfnAsg.addOverride(
      "CreationPolicy.ResourceSignal.Timeout",
      Duration.minutes(5).toIsoString(),
    );
```

PR #1265, opened 2025-09-01
fix: Increase wait time for new instance in ASG (attempt 2)

kelvin-chappell:
This was previously attempted in #1261.
In that PR I [chose the wrong timeout](https://github.com/guardian/security-hq/pull/1261#issuecomment-3232580144).

## What does this change?
This increases the timeout for a new instance being added to an ASG from 3 minutes to 5 minutes.

## What is the value of this?
Should reduce number of deployment failures we have seen.
For example, in this [failed deploy](https://riffraff.gutools.co.uk/deployment/view/973e39e3-34b0-411a-af02-c520f6f1223a):
> [11:45:39] New instance(s) added to autoscaling group - Waiting on 1 resource signal(s) with a timeout of PT3M.
> [11:48:36] AutoScalingGroupSecurityhqASG8DD277A5 (AWS::AutoScaling::AutoScalingGroup): UPDATE_IN_PROGRESS
> [11:48:36] Failed to receive 1 resource signal(s) for the current batch. Each resource signal timeout is counted as a FAILURE.
> [11:48:41] AutoScalingGroupSecurityhqASG8DD277A5 (AWS::AutoScaling::AutoScalingGroup): UPDATE_FAILED
> [11:48:41] Received 0 SUCCESS signal(s) out of 1. Unable to satisfy 100% MinSuccessfulInstancesPercent requirement
> [11:48:41] security-hq-PROD (AWS::CloudFormation::Stack): UPDATE_ROLLBACK_IN_PROGRESS

### DIFF
--- a/cdk/lib/security-hq.ts
+++ b/cdk/lib/security-hq.ts
@@ -161,8 +161,9 @@ export class SecurityHQ extends GuStack {
     });
 
     /*
-     * Increase the resource signal timeout for a new instance added to the ASG
-     * from the default value of 3 mins to 5 mins.
+     * Increase the health-check grace period for a new instance added to the ASG
+     * from the default value of 2 mins to 4 mins.
+     * This is in response to occasional failed deploys because of timeouts.
      *
      * To do this, we have to access the underlying ASG node.
      */


jacobwinch:
There are a few different timeouts involved here.

I would recommend increasing your [health check grace period](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-autoscaling-autoscalinggroup.html#cfn-autoscaling-autoscalinggroup-healthcheckgraceperiod) instead of the [pause time](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate-pausetime). There's an example of that [here](https://github.com/guardian/prism/blob/51d1f0574edaff31a6b4264dba6c089d69105ba2/cdk/lib/prism.ts#L124). The pattern will [automatically figure out a sensible pause time based on the health check grace period](https://github.com/guardian/cdk/pull/2462).

If you increase the pause time but not the health check grace period then I think the AutoScaling Group might terminate a new instance while CloudFormation's rolling update process is still waiting for it.

PR #1379, opened 2026-05-29
chore: Limit Dependabot Scala updates to LTS line

kelvin-chappell:
In #1376 we saw Scala being bumped to the Next line.  This change should keep it on the [current LTS line](https://www.scala-lang.org/development/#scala-3-lts) (v3.3.x).


tjsilver (approved):
👍

PR #1382, opened 2026-06-02
fix: bump awssdk version to latest to remediate netty-codec vuln

tjsilver:
## What does this change?

Upgrades the version of aws sdk to latest (v2.46.0) which brings in the patched version of io.netty:netty-codec (v4.1.133).

## What is the value of this?

Remediates the vulnerability in [this security alert](https://github.com/guardian/security-hq/security/dependabot/165).

## Any additional notes?

- [x] Tested locally


kelvin-chappell (approved):
👍

PR #1392, opened 2026-06-26
Create SSM Developer Policy demo

aug24:
## What is the value of this?
I'm introducing a dev policy for ssm console access so we have an exemplar for access privileges.


kelvin-chappell (approved):
Looks like a good starting point 👍

PR #1393, opened 2026-06-29
refactor: Move all non-Play components to a core SBT project

kelvin-chappell:
## What does this change?
This creates a new SBT project, `security-hq-core`, with standard Scala app structure and no reliance on the Play framework.

It moves the following packages, and corresponding test packages, from the Play app SBT project `security-hq` to the new project:
* aws
* config (partial)
* db
* logging
* logic (leaving `DocumentUtil` behind)
* model
* notifications
* utils

The `Config` object now calls on `CoreConfig` for values shared between the projects.

To remove dependence on the Play framework there are changes to the signature of some methods.  Primarily, the `AWS.clients` method now takes List[AwsAccount] as an argument rather than deriving this from `play.api.Configuration`.

Similarly there's a new util.Logging trait so that logging calls in `security-hq-core` have no dependence on the Play framework.

## What is the value of this?
This is enabling work for the extraction of lambdas for each of the background Play services, which will enable us to decommission the Play app.

No.

## Any additional notes?
See [Asana ticket](https://app.asana.com/1/1210045093164357/project/1213214650045516/task/1216038883987056?focus=true) and [parent ticket](https://app.asana.com/1/1210045093164357/project/1213214650045516/task/1214109339123725?focus=true).


core/src/main/scala/aws/AWS.scala
```diff
@@ -50,11 +49,11 @@ object AWS {
 
   private[aws] def clients[A, B <: AwsClientBuilder[B, A]](
     builder:  AwsClientBuilder[B, A],
-    configuration: Configuration,
+    accounts: List[AwsAccount],
```
adamnfish-gu, line 52: Nice decoupling, makes sense 👍

core/src/main/scala/config/CoreConfig.scala
```diff
@@ -0,0 +1,15 @@
+package config
+
+import software.amazon.awssdk.regions.Region
+
+/** Play-free configuration values shared by `core` and `hq`. */
```
adamnfish-gu: Do we need to mention "play-free" here? This is only important to us now while we migrate off Play. I'm not sure that context will matter after this migration is complete, in which case it's a commit message / PR concern rather than a source code comment one?

core/src/main/scala/utils/Logging.scala
```diff
@@ -0,0 +1,8 @@
+package utils
+
+import org.slf4j.{Logger, LoggerFactory}
+
+/** A lightweight logging trait backed by SLF4J so that `core` doesn't depend on Play. */
+trait Logging {
```
adamnfish-gu: Is this required? If we use something like [scala-logging](https://github.com/lightbend-labs/scala-logging) we'd have a maintained version of this (with other benefits).

hq/app/config/Config.scala
```diff
@@ -28,15 +28,14 @@ import software.amazon.awssdk.services.ssm.SsmClient
 
 
 object Config {
-  val iamHumanUserRotationCadence: Long = 90
-  val iamMachineUserRotationCadence: Long = 365
-  val outdatedCredentialOptOutUserTag = "SecurityHQ::OutdatedCredentialOptOut"
-  val daysBetweenWarningAndFinalNotification = 7
-  val daysBetweenFinalNotificationAndRemediation = 7
+  val iamHumanUserRotationCadence: Long = CoreConfig.iamHumanUserRotationCadence
```
adamnfish-gu: These feel like redundant duplication now, could we just write `import CoreConfig.*` here instead?

adamnfish-gu (approved):
Looks sensible, but we'll want to be careful to check the functionality is all still as expected.

core/src/main/scala/config/CoreConfig.scala
kelvin-chappell: Removed in 41b848020f4603bb51828ba7019b09341fbbdb08

core/src/main/scala/utils/Logging.scala
kelvin-chappell: Introduced the scala-logging lib in [8f28789](https://github.com/guardian/security-hq/pull/1393/commits/8f28789a05a853d7bda884fa0bc87db215718430)

hq/app/config/Config.scala
kelvin-chappell: Took out the repeated values and changed refs to point to `CoreConfig` in [84b1a71](https://github.com/guardian/security-hq/pull/1393/commits/84b1a716d956ebed17b97f71b591d1737c82ac43)

PR #1403, opened 2026-07-02
chore: Create setup-host script

kelvin-chappell:
We've decided that the standard process for working on a repo should be:
1. Run an optional `setup-host` script.
2. Start a dev container.
3. Get AWS creds.
4. Run a `setup` script.
5. Run a `start` script.

This adds the `setup-host` script for the repo.  It will soon become redundant but just establishing a pattern.

The current `setup` script won't work in a dev container as-is because it sets up dev-nginx.


.devcontainer/README.md
```diff
@@ -5,10 +5,10 @@ an isolated, reproducible development environment that runs inside a container,
 
 The configuration skeleton is generated by the Guardian's [devenv](https://github.com/guardian/devenv) config tool.
 
-Before creating a dev container, run the setup script locally:
+Before creating a dev container, run the `setupHost` script locally:
 
 ```bash
-./script/setup
+./script/setupHost.sh
```
adamnfish-gu: Minor point, but would `setup-host` be a more canonical name?

`camelCase` is more of a Java convention, shell scripts more often use use `kebab-case`. Also, the `.sh` suffix is inconsistent with the rest of the directory (and redundant)?

script/setup-host
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
adamnfish-gu, line 1: Should this script also execute `devenv generate`? I'm not sure, but we would want to make sure people run this to include their own user customisations.

kelvin-chappell: Done in [64b4330](https://github.com/guardian/security-hq/pull/1403/commits/64b433005cdb192a95efeb092d25a167d8e980d8)

.devcontainer/README.md
kelvin-chappell: Done in [b60f8e3](https://github.com/guardian/security-hq/pull/1403/commits/b60f8e39b9bd61427fef75e377c567a34da974d2)

script/setup-host
```diff
@@ -0,0 +1,33 @@
...
+  exit 1
+fi
+
+dev-nginx setup-app "$MAPPING_FILE"
+
+# See https://github.com/guardian/devenv
+if [[ -f "$DEVENV_CONFIG" ]]; then
+  devenv generate
+else
+  devenv init
```
adamnfish-gu: Should we do this? We know SHQ has a devenv file so it's kind of redundant - if we're thinking this will be a generic setup-host script then even more so, we wouldn't want to  leave them with a half-setup devenv config (and dirty their worktree) if they've decided not to use devenv?

It might be enough to mention that there's no devenv configuration and point them to devenv's docs if they want to set it up?

kelvin-chappell: That's true.  There's no need for a generic solution here.

script/setup-host
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+# The nginx setup will become redundant when the Play app has been decommissioned.
+
+set -euo pipefail
+
+DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+ROOT_DIR="${DIR}/.."
+MAPPING_FILE="$ROOT_DIR/nginx/nginx-mapping.yml"
+DEVENV_CONFIG="$ROOT_DIR/.devcontainer/devenv.yaml"
```
adamnfish-gu: I think this can go now that we aren't running a check for it?

adamnfish-gu (approved):

PR #1406, opened 2026-07-03
Separate out Iam Outdated Credentials lambda

aug24:
## What does this change?

Introduces a main and lambda handler entry point to iam-outdated-credentials.

Re-organises to simplify those.  ~More simplification will come later because they share too much code at this point.~

(Almost) all config is expected to be available in S3.  Main parameters are now
```
stack :: stageString :: dryRunString :: configBucket :: configKey
```
and all values are found in the config file at s3://configBucket/configKey

NB No changes to tests other than moving one to the subproject.

## What is the value of this?

We will eventually remove the functionality from the SHQ app.

*This set of changes does not change the functionality within security hq. It is a first step toward moving this functionality elsewhere.  It delivers a build artifact which _we do not deploy_, but which can be downloaded and tested.*


core/src/main/scala/aws/iam/IAMClient.scala
```diff
@@ -86,23 +85,33 @@ object IAMClient extends LazyLogging {
...
-          CredentialsReport.isComplete,
-          "Failed to generate credentials report",
-          delay
-        )
-        report <- getCredentialsReport(client)
-        stacks <- CloudFormation.getStacksFromAllRegions(account, cfnClients, regions)
-        reportWithTags <- enrichReportWithTags(report, client)
-        reportWithStacks = CredentialsReport.enrichReportWithStackDetails(reportWithTags, stacks)
-      } yield CredentialsReportDisplay.toCredentialReportDisplay(reportWithStacks)
+      getUpdatedCredentialsReport(account, cfnClients, iamClients, regions, delay)
```
aug24: Extracted so that it can be called without the cache.

core/src/main/scala/config/CoreConfig.scala
```diff
@@ -12,4 +24,54 @@ object CoreConfig {
 
   // TODO fetch the region dynamically from the instance
   val region: Region = Region.of("eu-west-1")
+
```
aug24, line 35: Moved from AppComponents.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsLambda.scala
```diff
@@ -0,0 +1,83 @@
...
+  extends RequestHandler[Map[String, String], String] {
+
+  implicit val ec: ExecutionContext = ExecutionContext.global
+
+  override def handleRequest(
+                              event: Map[String, String],
+                              context: Context
+                            ): String = {
+
+    val settings = Settings.fromEnvironment()
```
adamnfish-gu, line 18: I wonder if much of this static logic should be in the class rather than the handler def. We don't necessarily want to re-run all of it on every execution, and it'll help us focus on the actual invocation logic.

It depends though, maybe some of this needs to be more dynamic than that.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsLambda.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    val cfnClients = AWS.cfnClients(awsAccounts, availableRegions)
+    val delay = 3.seconds
+    val credentialReportFutures: Seq[Future[(AwsAccount, Either[FailedAttempt, CredentialReportDisplay])]] = awsAccounts
+      .map(account =>
+        IAMClient.getUpdatedCredentialsReport(account, cfnClients, iamClients, availableRegions, delay).asFuture
+          .map(result => account -> result))
+
+    val timeout = 3.minutes
+    val eventualList: Future[Seq[(AwsAccount, Either[FailedAttempt, CredentialReportDisplay])]] = Future.sequence(credentialReportFutures)
```
adamnfish-gu: Can we use `traverse` rather than `map` and `sequence`?

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsLambda.scala
```diff
@@ -0,0 +1,83 @@
...
+    val result =
+      job.disableOutdatedCredentials(
+        notificationTopicArn = settings.anghammaradSnsArn,
+        tableName = settings.iamDynamoTableName,
+        serviceAccountIds = settings.accountIdsForIamRemediationService,
+        rawCredsReports = listOfCredentialReports,
+        allowedAwsAccountIds = settings.allowedAccountIds
+      ).asFuture
+
+    Await.result(result, 10.minutes) match {
```
adamnfish-gu: We shouldn't have multiple awaits in a single program.

Can we refactor this to use a for comprehension, which already encodes dependent asynchronous steps. We can then await the program's result once at the edge of the program.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsLambda.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    val job = new IamOutdatedCredentials(
+      snsClient = snsClient,
+      iamClients = iamClients,
+      dynamo = dynamo,
+      dryRun = settings.dryRun
+    )
+
+    val cfnClients = AWS.cfnClients(awsAccounts, availableRegions)
+    val delay = 3.seconds
```
adamnfish-gu: Is this delay the same as the old version? It seems like quite a low number for polling something we expect to take minutes?

No worries! I don't think this is necessarily wrong, just checking the thinking.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsLambda.scala
```diff
@@ -0,0 +1,83 @@
...
+    val credentialReportFutures: Seq[Future[(AwsAccount, Either[FailedAttempt, CredentialReportDisplay])]] = awsAccounts
+      .map(account =>
+        IAMClient.getUpdatedCredentialsReport(account, cfnClients, iamClients, availableRegions, delay).asFuture
+          .map(result => account -> result))
+
+    val timeout = 3.minutes
```
adamnfish-gu: We have a 10 minute overall timeout, but only 3 minutes for the bit that we expect to take a long time. Do we even need this 3 minute limit?

As below, I think we might consider running this as a single composed Future and clamping the overall execution time at the edge of the program.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsLambda.scala
```diff
@@ -0,0 +1,83 @@
...
+import software.amazon.awssdk.services.s3.S3Client
+import software.amazon.awssdk.services.sns.SnsAsyncClient
+import utils.attempt.FailedAttempt
+
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.concurrent.duration.*
+import scala.jdk.CollectionConverters.ListHasAsScala
+
+class IamOutdatedCredentialsLambda
+  extends RequestHandler[Map[String, String], String] {
```
adamnfish-gu: I think this handler should roughly be:

0. maybe some setup?
1. a slim Future for comprehension
2. a result handler that awaits and handles the Either result

Most of the static setup should be in the class, I don't think it should be a concern of the invocation unless it's something we expect to change. For example I'm guessing in the case of accounts / regions we might want these to be dynamic so that new accounts / regions are picked up on each run. Let's be deliberate about this though!

A lot of these apply to the Main version as well, and that itself makes me think these aren't sharing enough code.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsMain.scala
```diff
@@ -0,0 +1,126 @@
...
+        case Right(_) =>
+          println("Completed successfully")
+
+        case Left(err) =>
+          Console.err.println(s"Job failed: $err")
+          sys.exit(1)
+      }
+    case _ =>
+      println(
+        "Usage: sbt 'project iam-outdated-credentials; runMain logic/IamOutdatedCredentialsMain " +
```
adamnfish-gu: simpler to write this as:

```bash
sbt 'iam-outdated-credentials/runMain logic.IamOutdatedCredentialsMain ...'
```

using `projectName/task` rather than `project projectName; task` shows what we're doing in one place and doesn't make batched tasks "stateful"

I'm also surprised this works as written, I thought runMain's argument would want a `.` rather than a `/`?

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsMain.scala
```diff
@@ -0,0 +1,126 @@
...
+          "   <stack>" +
```
adamnfish-gu: Is there a reason for splitting this (single line) docstring over so many lines?

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsMain.scala
```diff
@@ -0,0 +1,126 @@
...
+import utils.attempt.FailedAttempt
+
+import scala.concurrent.duration.*
+import scala.concurrent.{Await, ExecutionContext, Future}
+
+object IamOutdatedCredentialsMain {
+
+  implicit val ec: ExecutionContext = ExecutionContext.global
+
+  def main(args: Array[String]): Unit = args.toList match {
```
adamnfish-gu: I'm drawn to question how usable this main method is..! It has so many parameters and they all need to be correctly passed in the right order.

Are people going to use this? If not can it go, and if so, is there a way to make this easier? Should this instead read from a file we can provision into `~/.gu/` or something like that?

I'm just trying to imaging anyone in the future ever calling this 😅

build.sbt
```diff
@@ -158,7 +159,7 @@ lazy val hq = (project in file("hq"))
 lazy val iamOutdatedCredentials = (project in file("iam-outdated-credentials"))
   .enablePlugins(JDebPackaging)
   .disablePlugins(sbtassembly.AssemblyPlugin)
-  .dependsOn(core)
+  .dependsOn(core % "compile->compile;test->test")
```
adamnfish-gu, line 161: Do we need this test `dependsOn`? I don't spot an obvious place where we depend on test code from `core`?

iam-outdated-credentials/src/main/scala/settings/Settings.scala
```diff
@@ -0,0 +1,48 @@
+package settings
+
+import model.{DEV, PROD, Stage}
+import software.amazon.awssdk.regions.Region
+
+/** Runtime configuration for the iam-outdated-credentials lambda, sourced from environment variables set by the CDK
+  * stack.
+  *
+  */
+case class Settings(
```
adamnfish-gu, line 8: This all feels super closely related to the new functionality in `CoreConfig`. Is it worth giving this a bit more big-picture thought, to see if the config / settings / parameters concepts we have can be simplified?

adamnfish-gu (commented):
The Lambda / Main are currently listing loads of steps in a lot of detail. It's nice for the edge of the program to describe the steps so people can immediately see what's happening, and then they can dig into each piece of they need those details.

There are a few things that could help with this!

Firstly, it might be nice to move the static setup to the class so that we don't have to reload everything on every invocation, and to move noise out the handler. Then for anything that really is "dynamic" rather than just putting tens and tens of lines of code straight into the handler, is there a way to have the handler ask for the config, parametrised as needed? This would make the handler much easier to read - we currently have to wade through a lot of boiler plate to see what it actually does.

It would also allow the Main and Lambda to more meaningfully be "the same program". Right now they look more like two roughly duplicate definitions, and in this case I begin to question the value of the Main version. If purpose is to "run the program locally", would we want them to be a bit more similar? I realise the bulk of the shared code is in the `disableOutdatedCredentials` call, but it's still useful to have these share as much as possible to be able to locally reproduce issues.

On this point, we could also look at one of the many solutions for running lambda functions locally, especially if we don't care about the structure of the invocation event. No need to do tackle researching that here, but if that sounded good then maybe that would tell us the CLI version might not be needed for now?

If we're writing `Future` programs, we want a single `Await` at the program's "edge", and we use the various tools for composing Futures in the program itself (primarily flatMap via a for-comprehension, map, and traverse). Each `Await` is a tricky source of issues and the introduction of a new type of side-effect, especially when considering timeouts.

I get we don't want to put loads of effort into all this, but doing a pass-over for simplicity is still in order. I think moving the effects to the edge (one Await), and uncluttering the handler/main might be worth a look.

build.sbt
aug24: `AttemptValues`

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsLambda.scala
aug24: This is from 
```
 def getCredentialReportDisplay(
<snip>
  )(implicit ec: ExecutionContext): Attempt[CredentialReportDisplay] = {
    val delay = 3.seconds
```

It's the delay between fetches, not the timeout value.  It's not easy to read!

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsLambda.scala
aug24: No, we didn't need this.  It was introduced as part of the refactoring and goes away now that the traverses etc are all in place.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsMain.scala
aug24: I've reworked to use all the values from the existing config files in S3.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsMain.scala
aug24: It suddenly dawns on me that `run` should be sufficient anyway!  We're only going to have one.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsLambda.scala
aug24: Absolutely!  See comment on PR
```
More simplification will come later because they share too much code at this point.
```

I have simplified further in this PR now.

iam-outdated-credentials/src/main/scala/settings/Settings.scala
aug24: They certainly are related, but the CoreConfig concepts go across both lambdas and the existing app, where the Settings object is specific to this sub-project.

core/src/main/scala/config/CoreConfig.scala
```diff
@@ -12,4 +24,54 @@ object CoreConfig {
...
+  }
+
+  val securityCredentialsProvider: AwsCredentialsProviderChain = AwsCredentialsProviderChain.of(
+    ProfileCredentialsProvider.create("security"),
+    DefaultCredentialsProvider.builder.build()
+  )
+
+
+  def getSecurityDynamoDbClient(stage: Stage) = {
```
adamnfish-gu: Is `def` right for this? This means we'd be re-creating the client on every call. It looks like both uses of this are a `val` so all good for now, but I mention this as something that this design means needs to be handled carefully in the future.

hq/app/services/IamRemediationService.scala
```diff
@@ -101,26 +101,27 @@ class IamRemediationService(
...
     // this tells us which AWS accounts we are allowed to make changes to
     allowedAwsAccountIds <- Config.getAllowedAccountsForStage(config)
-  } yield new IamOutdatedCredentials(snsClient, iamClients, dynamo, dryRun = false).disableOutdatedCredentials(
-    notificationTopicArn,
-    tableName,
-    serviceAccountIds,
-    rawCredsReports,
-    allowedAwsAccountIds
-  )
+    _ = IamOutdatedCredentials(snsClient, iamClients, dynamo, dryRun = false).disableOutdatedCredentials(
```
adamnfish-gu: This looks wrong to me.

`_ = ...` means the is a pure statement (it's an assignment `=` not a chained operation `<-`

That means either this is a pure statement that is redundant (since we're assigning it to `_`), or this is a side effecting statement that should be a `<-`.

Looking at the signature for `disableOutdatedCredentials` I can see that it returns an `Attempt`, so this looks like the latter.

Maybe let's write this as something like the following:

```scala
for {
  ...
  outdatedCredentials = IamOutdatedCredentials(snsClient, iamClients, dynamo, dryRun = false)
  _ <- outdatedCredentials.disableOutdatedCredentials( ... )
} yield ()
```

But certainly, making sure we have a `<-` for our effect.

Previously we were returning the result (instead of `Unit`), which prevented this bug. Now that we return `Unit` the compiler accepts implementations that don't properly sequence their side effects. With a return type that comes from the last step we can only meet the contract by executing the final side effect, but if we return `Unit` we can do so anywhere.

This suggests an alternative, which would be for this to return the value we get at the end of our execution. Doing this would mean the compiler then enforces that we handle this execution properly.

Note: because we're using `Future` (rather than a functional effect type) this *might* work as is! It's "undefined" behaviour in general because it'll depend on how the underlying threads are managed in a lot of technical detail (including the Lambda execution environment itself). When this Future gets created it'll start immediately, but we're just firing and forgetting it because it isn't sequenced into our program's control flow. This means it might complete in the background, or it might not. Not what we want!

iam-outdated-credentials/src/main/resources/logback.xml
```diff
@@ -0,0 +1,38 @@
...
+
+  <appender name="ASYNCFILE" class="ch.qos.logback.classic.AsyncAppender">
+    <appender-ref ref="FILE" />
+  </appender>
+
+  <appender name="ASYNCSTDOUT" class="ch.qos.logback.classic.AsyncAppender">
+    <appender-ref ref="STDOUT" />
+  </appender>
+
+  <logger name="play" level="INFO" />
```
adamnfish-gu: Is this needed, it isn't a Play app?

iam-outdated-credentials/src/main/resources/logback.xml
```diff
@@ -0,0 +1,38 @@
...
+
+  <!-- Off these ones as they are annoying, and anyway we manage configuration ourselves -->
+  <logger name="com.avaje.ebean.config.PropertyMapLoader" level="OFF" />
+  <logger name="com.avaje.ebeaninternal.server.core.XmlConfigLoader" level="OFF" />
+  <logger name="com.avaje.ebeaninternal.server.lib.BackgroundThread" level="OFF" />
+  <logger name="com.gargoylesoftware.htmlunit.javascript" level="OFF" />
```
adamnfish-gu: I'd say the same here, I can't see why any of these would be present in the Lambda jar?

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -168,6 +180,53 @@ class IamOutdatedCredentials(
 }
 object IamOutdatedCredentials extends LazyLogging {
 
+  private val ACCOUNT_IDS_FOR_IAM_REMEDIATION_SERVICE = "ACCOUNT_IDS_FOR_IAM_REMEDIATION_SERVICE"
```
adamnfish-gu: I'm not sure about the value of these. It doesn't look like they're all used, and those that are used appear to be used once. If we statically know the String value, does duplicating that text as a value name help us?

I don't mind heaps, it's nice to encode assumptions clearly. Maybe we at least make sure these are all used?

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -412,11 +471,40 @@ object IamOutdatedCredentials extends LazyLogging {
...
+      .builder()
+      .bucket(bucket)
+      .key(key)
+      .build()
+
+    val inputStream: InputStream = s3.getObject(request)
+
+    val configString =
+      Using.resource(inputStream) { is =>
+        scala.io.Source.fromInputStream(is).mkString
```
adamnfish-gu: It looks like we're handling the stream properly (with `Using.resource`), but the `Source` inside it isn't closed after use. Shall we handle that as well?

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -412,11 +471,40 @@ object IamOutdatedCredentials extends LazyLogging {
...
+      }
+
+    ConfigFactory.parseString(configString).resolve()
+  }
+
+  private def parseAccounts(config: Config): List[AwsAccount] = {
+    config.getConfigList(AWS_ACCOUNTS).asScala.toList.map { accountConfig =>
+      AwsAccount(
+        id = accountConfig.getString("id"),
```
adamnfish-gu: Does this need error handling if these keys don't exist?

What behaviour would we want in these cases? The current behaviour is "this will crash and we won't do any IamOutdatedCredentials work", but maybe we would want to try our best with the accounts that are properly configured?

This doesn't have to be done here, but it's worth calling out the behaviour we have. Perhaps it's worth a comment, and/or a ticket to follow up?

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsLambda.scala
```diff
@@ -0,0 +1,30 @@
...
+    val settings = Settings.fromEnvironment()
+
+    val result = IamOutdatedCredentials.disableOutdatedCredentials(settings)
+
+    Await.result(result.underlying, 10.minutes) match {
+      case Right(_) =>
+        "Success"
+
+      case Left(err) =>
+        throw new RuntimeException(err.toString)
```
adamnfish-gu: Should we say something about this beyond the error string? We also will certainly want to populate the "cause" of this RuntimeException, if we have one from our failure.

Let's rename `err` to `failedAttempt` to highlight its actual type. We shouldn't be using "toString" on this type, it exposes a `logMessage` field that is probably a better fit. We could then use its `firstException` for the exception's cause, using `firstException.orNull`

```scala
case Left(failedAttempt) =>
  throw new RuntimeException(
    s"IamOutdatedCredentials execution failed: ${failedAttempt.logMessage}",
    failedAttempt.firstException.orNull
  )
```

This will give our logs a helpful (and detailed) error message, and a stacktrace.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsMain.scala
```diff
@@ -0,0 +1,40 @@
...
+  def main(args: Array[String]): Unit = try {
+    val settings = Settings.fromArgs(args)
+
+    val result: Attempt[Unit] = IamOutdatedCredentials.disableOutdatedCredentials(settings)
+
+    Await.result(result.underlying, 10.minutes) match {
+      case Right(_) =>
+        println("Completed successfully")
+
+      case Left(err) =>
```
adamnfish-gu: Here again, this isn't an `err` it is a `FailedAttempt` (see above).

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentialsMain.scala
```diff
@@ -0,0 +1,40 @@
...
+        Console.err.println(s"Job failed: $err")
+        sys.exit(1)
+    }
+  } catch {
+    case _: IllegalArgumentException =>
+      println(
+        """
+        |Usage: sbt 'iamOutdatedCredentials / run
```
adamnfish-gu: We normally write project scoped tasks without spaces `project/task`, which helps a little with the visual parsing. This is a single task (scoped to the project), and subsequent spaces are used to separate arguments. At the moment the spaces are doing two different jobs - delineating the project / task for the first three items, and then separating argument for the rest.

If we write it as `iamOutdatedCredentials/run arg1 arg2` it's a bit more consistent, and looks more like a standard CLI invocation.

I've written this here, but if you agree there are a few other places this is written

iam-outdated-credentials/src/main/scala/settings/Settings.scala
```diff
@@ -0,0 +1,63 @@
...
+    Settings(
+      dryRun = sys.env.getOrElse(DRY_RUN, "true").toBoolean,
+      stack = required(STACK),
+      stage = if (required(STAGE).equalsIgnoreCase("prod")) PROD else DEV,
+      configBucket = required(CONFIG_BUCKET),
+      configKey = required(CONFIG_KEY)
+    )
+  }
+
+  private val DRY_RUN = "DRY_RUN"
```
adamnfish-gu, line 74: I like that these are at the bottom, since they aren't the main purpose of this object 👍 

Maybe we could do this for the IamOutdatedCredentials settings keys as well?

adamnfish-gu (commented):
There's an important point about the main side effect not being properly threaded into our for comp that certainly wants a look!

Some other observations as well, not all for this PR.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
aug24: I prefer this sort of thing because I avoid "magic" text strings without context, but they definitely should be used.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
aug24: In the lambda definition (the cdk ticket), there's an automatic alert for failed runs.  As this is "bad config" I'd probably prefer to fail.

core/src/main/scala/config/CoreConfig.scala
aug24: This is really a temporary measure because the IamRemediationService (which uses it in the app) is going to be removed eventually.  At that point we can unwrap the method (which is just to avoid duplication and guarantee identical behaviour) and make it just a val in the IamOutdatedCredentials lambda.  The disableUnrecognisedUsers method does not use it.

core/src/main/scala/config/CoreConfig.scala
```diff
@@ -12,4 +24,54 @@ object CoreConfig {
...
+        .build(),
+      AwsAccount(stack, stack, stack, stack),
+      CoreConfig.region
+    )
+    try {
+      val availableRegionsAttempt: Attempt[List[Region]] = for {
+        ec2RegionList <- EC2.getAvailableRegions(ec2Client)
+        regionList = ec2RegionList.map(ec2Region => Region.of(ec2Region.regionName))
+      } yield regionList
+      Await
```
adamnfish-gu: We generally want a single Await at the edge of our program. I'm super nervous to see an Await hidden in the middle of a utility function, here. Can we have this return a Future and ask our "edges" to make a decision about its execution?

I can see two consumers of this:

`IamOutdatedCredentials.disableOutdatedCredentials` is already a function that returns a Future, so this should definitely be composing this async operation with its other futures via flatMap (`<-`).

The Play app's AppComponents setup in the class constructor is the other place this is used, and this is the source of the function that's been extracted. This was the "edge" of that Play server, which is why the Await was previously ok here. With the refactor in this PR, we need to keep the Await in AppComponents and have this function return a Future (AKA `Attempt`, in this codebase).

adamnfish-gu (commented):
Giving you one point before I get time to re-review the entire diff!

build.sbt
```diff
@@ -17,7 +17,8 @@ ThisBuild / scalacOptions ++= Seq(
 
 resolvers += DefaultMavenRepository
 
-val awsSdkVersion = "2.47.0"
+val awsLambdaVersion = "1.2.3"
+val awsSdkVersion = "2.46.2"
```
adamnfish-gu: Is there a reason why we're downgrading the AWS SDK here?

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -168,6 +179,46 @@ class IamOutdatedCredentials(
 }
 object IamOutdatedCredentials extends LazyLogging {
 
+  def disableOutdatedCredentials(settings: Settings)(implicit executionContext: ExecutionContext): Attempt[Unit] = {
+    val snsClient = SnsAsyncClient.builder().build()
+    val s3Client = S3Client.builder.build()
+    val availableRegions: List[Region] = calculateAvailableRegions(settings.stack, settings.stage)
```
adamnfish-gu: It's a bit confusing that we're importing some functions from other objects so we can use them unqualified here (e.g. `CoreConfig.calculateAvailableRegions`), but for functions that are actually in scope here we're calling them with this object as a qualifier (e.g. `IamOutdatedCredentials.loadConfigFromS3`).

Not a huge deal, but maybe for more consistency we'd leave the `CoreConfig` prefix on here and leave the unqualified references for the local vals?

adamnfish-gu (commented):
Just a couple of questions to add to the quick check from earlier.

core/src/main/scala/config/CoreConfig.scala
aug24: Reverted to original for AppComponents, with a TODO comment saying we will be tearing this out.  New version returning Attempt used in lambda code.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
aug24: Ticket raised to discuss.

core/src/main/scala/config/CoreConfig.scala
aug24: Following other changes, this is no longer altered.  A new implementation for the lambdas is provided instead.

core/src/main/scala/config/CoreConfig.scala
aug24: Done

build.sbt
aug24: Merge error

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -168,6 +170,45 @@ class IamOutdatedCredentials(
...
+        dryRun = settings.dryRun
+      ).disableOutdatedCredentials(
+        notificationTopicArn = anghammaradSnsArn,
+        tableName = iamDynamoTableName,
+        serviceAccountIds = accountIdsForIamRemediationService,
+        rawCredsReports = listOfCredentialReports,
+        allowedAwsAccountIds = allowedAccountIds
+      )
+    } yield disableResult
+  }
```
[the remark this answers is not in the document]
aug24: Rewritten in my own commit because it also needed to include iamClients and cfnClients.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -168,6 +170,95 @@ class IamOutdatedCredentials(
...
+          awsAccountsConfig.getStringList(ALLOWED_ACCOUNT_IDS_CONFIG_ITEM).asScala.toList
+        val accountIdsForIamRemediationService =
+          awsAccountsConfig.getStringList(REMEDIATION_ACCOUNT_IDS_CONFIG_ITEM).asScala.toList
+
+        withResource(CoreConfig.getSecurityDynamoDbClient(settings.stage)) { dynamoDbClient =>
+          val dynamo = new IamRemediationDb(dynamoDbClient)
+
+          CoreConfig
+            .calculateAvailableRegions(settings.stack, settings.stage)
+            .flatMap { availableRegions =>
```
adamnfish-gu, line 190: This is flatMapping an attempt, and then we for more flatMapping of attempts just below (in a for comprehension).

I've dug in below and I can see that the resource/cleanup abstraction doesn't fit with our Attempts. This is the first point to see the impact of this - we're switching between resources and Attempts at each layer of indentation.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -168,6 +170,95 @@ class IamOutdatedCredentials(
...
+    }
+  }
+
+  private def withResource[R <: AutoCloseable, A](acquire: => R)(use: R => Attempt[A])(implicit
+      ec: ExecutionContext
+  ): Attempt[A] = {
+    withCleanup(acquire)(resource => closeQuietly(resource.close()))(use)
+  }
+
+  private def withCleanup[R, A](acquire: => R)(cleanup: R => Unit)(use: R => Attempt[A])(implicit
```
adamnfish-gu, line 230: I'm not crazy about this design. I think this is directly contributing to the indentation and tangle of the code above. The comment mentions this is inspired by cats-effect, provides `Resource`. While it's true the underlying abstraction looks similar to what is written here, the ergonomics are quite different!

We'd typically expect to see a "Resource" program edge to be a for comp rather than nested acquisitions. We'd generally be aiming to have a program edge that does the all resource setup in a bunch of steps, separate to the program which is the IO in the final use call. Where the resources and IO are entwined (as here), we'd still have one for-comp, but would lift those IO steps into Resource.

The design we have here is fundamentally different to Cats-Effect because it is a callback interface, instead of a Monad. If we're using callbacks we might as well write it the Java way! What we have here is sort of a hybrid where we make the Java style look a bit like Cats-Effect by having everything return Attempt. We're kind of getting the downsides of both approach - we have the nesting and entanglement of callbacks, with the problem of lining up the effect signatures like a Monadic effect system.

The key thing to keep in mind is that `Resource` (in Cats-Effect) is a different Monad to `IO`, and `Attempt` is more like the latter. Specifically, `Attempt` (which pre-dates Cats-Effect) is a like a specialised Monad Transformer for `Future` and `Either`, so it would be closest to a concrete implementation of `MonadError` in Cats-Effect parlance.

This is a roundabout way of saying, this disableOutdatedCredentials above is a significantly nested / indented function now, and I think this is because we have tangled incompatible effects.

adamnfish-gu: Ultimately, in 2026 we'd be mad to write our own implementation of `MonadError` (`Attempt`), but we have the legacy of that decade old decision here today. I think we'd now be even more mad to write our own implementation of `Resource`!

If this app's effects are getting complex enough to warrant this kind of code, I think we should do one of the following:
- migrate to Cats-Effect
- switch to a synchronous Try-backed approach
- have a cleanup step
- something else (see the next comment)

Migrating Attempt to Cats-Effect wouldn't be a huge task, but there will be a few gotchas that will take some CE expertise to navigate. We'd also then have an application with entirely new performance characteristics. We'd need to re-write it with care, and pay close attention to how the result runs. Cats-Effect has a pretty serious learning curve, and my hunch is that we'd be creating something the department would struggle to maintain.

I've found that sticking to `Try` as the program's "edge effect" and doing work synchronously strikes a decent balance for Scala in this department. We still get the design pressure to separate effectful and pure computations, structure the edge of the program into a bunch of steps, and so on, and we are able to skip most of the really tricky stuff. Using `cats` for things like `traverse` is fine. However, this would also be making a very significant change to how this code runs. We make a lot of API calls across multiple accounts and regions, and it has been useful to have some degree of parallelism.

Given both these approaches come with significant behavioural drawbacks, we should also be open minded to the approach of just writing `client.close()` after we've used it? The devil is in the details, of course.

adamnfish-gu: In the previous two comments I've addressed the code that we have here, but I think it's also worth talking a little more generally as well.

IIUC the existing behaviour was to have an AWS client cache that was populated at app start time, and used thereafter.

It's hard to duplicate this exact behaviour in AWS Lambda, because the environment lifecycles don't match. When I was first reviewing this code it looked like there was a deliberate decision here to lift some of the discovery work into the handler, so it refreshes on every execution?

I don't think this is necessarily required! If we do all our setup at setup time (akin to AppComponents for the Play app) we could copy the exact model that's there already? I don't mean to say this is perfect, but perhaps that would let us start to split up some of this work.

This PR is trying to do a lot of things at once. I'd prefer to see us close it, and reopen a bunch of little PRs that use what we've learned here to address these things a bit at a time. If we're all still thinking it is best to push forwards with this large PR first then that's fine too, but this PR is quite large! We should be careful about creating a situation where there's a lot of pressure on reviewers now just saying "eh, it's been open ages, just ship it".

I think the fact that other parts of the SHQ migration project are backing up behind this PR is a sign that we could all work together on clarifying the requirements, to prevent scope creep. With that done we can make a decision on whether to split one or more "SHQ replacement Lamdba foundations" out from the actual work to port over the Lambda logic.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -168,6 +170,95 @@ class IamOutdatedCredentials(
...
+        withResource(CoreConfig.getSecurityDynamoDbClient(settings.stage)) { dynamoDbClient =>
+          val dynamo = new IamRemediationDb(dynamoDbClient)
+
+          CoreConfig
+            .calculateAvailableRegions(settings.stack, settings.stage)
+            .flatMap { availableRegions =>
+              val iamClients = AWS.iamClients(awsAccounts, availableRegions)
+              val cfnClients = AWS.cfnClients(awsAccounts, availableRegions)
+
+              withCleanup((iamClients, cfnClients)) { clients =>
```
adamnfish-gu, line 194: With a real Resource abstraction we'd write these clients on separate lines rather than tupling them up. Even with this one (although see below) I'd encourage us to use a match here rather than using tuple indices on the `clients` value.

PR #1411, opened 2026-07-07
Switch to GuRoot to get generated riff-raff.yaml

aug24:
## What does this change?

This PR uses GuRoot instead of App in the cdk definition, which triggers the creation of a riff-raff.yaml file.

A future PR will switch to use this new riff-raff.yaml file for deployment.

Tested by running `npm run synth`.


NovemberTang (approved):

kelvin-chappell (approved):
👍 (perhaps description could say how you tested it)

PR #1412, opened 2026-07-07
Add scheduled lambda with identical role to the app

aug24:
## What does this change?

This change adds a scheduled lambda definition to the existing cdk generated cloudformation.

The jar file used is created by additional CI build steps (see https://github.com/guardian/security-hq/pull/1419) and uploaded as a riff-raff artifact.


cdk/lib/security-hq.ts
```diff
@@ -114,6 +118,30 @@ export class SecurityHQ extends GuStack {
 
     dpkg -i /tmp/installer.deb`);
 
+    const additionalPolicies = [
```
aug24: Abstracted to allow referencing by both the app definition and the lambda(s) definition(s).  We can duplicate and filter down if needed after the app schedule is removed.

cdk/lib/security-hq.ts
```diff
@@ -289,6 +294,40 @@ export class SecurityHQ extends GuStack {
...
+      handler: 'logic.IamOutdatedCredentialsLambda::handleRequest',
+      timeout: Duration.minutes(10),
+      environment: {
+        "STACK": this.stack,
+        "STAGE": this.stage,
+        "DRY_RUN": "true",
+        "CONFIG_BUCKET": "security-dist",
+        "CONFIG_KEY": `security/${this.stage}/security-hq/security-hq.conf`
+      },
+      fileName: `iam-outdated-credentials/iam-outdated-credentials-${buildIdentifier}.jar`
```
NovemberTang: Could we add some more context to the PR description for future travellers, or those (me) not super involved in the project? i.e. Are there other PRs I can look at for more information, like where this jar came from? Is there subsequent work that needs to happen after this? It's tricky to work out what's going on as it is today

kelvin-chappell (approved):
👍

PR #1413, opened 2026-07-07
Switch to new riff-raff.yaml

aug24:
## What does this change?

Once we have riff-raff.yaml being created by the cdk library, and we have checked the output in a prod build, we will want to switch over to it.

This is the existing riff raff plan for the old riff-raff.yaml

image

and this is the new plan

image

**They should be functionally identical**


kelvin-chappell (approved):
👍

PR #1416, opened 2026-07-08
feat: Add IAM unrecognised users lambda

tjsilver:
## What does this change?

- Creates a new scheduled AWS Lambda, `iam-unrecognised-users`, which automatically disables IAM users who are no longer recognised by Janus. 
- Adds a local entrypoint, hardcoded to dry-run and restricted to security account only. This outputs the notifications that would be sent.
- Adds a README with instructions on how to run locally.

## What is the value of this?

We can remove this service from the Play app. 

Yes, cloudformation has been updated.

No.


iam-unrecognised-users/src/main/scala/unrecognised/UnrecognisedUsers.scala
```diff
@@ -0,0 +1,101 @@
...
+  private val ALLOWED_ACCOUNT_IDS = "ALLOWED_ACCOUNT_IDS"
+  private val ANGHAMMARAD_SNS_ARN = "ANGHAMMARAD_SNS_TOPIC_ARN"
+
+  def run(env: Map[String, String] = sys.env, restrictToAccountId: Option[String] = None): Unit = {
+    implicit val ec: ExecutionContext = ExecutionContext.global
+    val settings = Settings.fromEnvironment(env)
+    logger.info(s"Starting iam-unrecognised-users job (dryRun=${settings.dryRun}, region=${settings.region.id})")
+    val result = disableUnrecognisedUsers(settings, restrictToAccountId)
+    Await.result(result.asFuture, 5.minutes) match {
+      case Left(failure) =>
```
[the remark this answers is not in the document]
tjsilver: Rather than hard-coding a value, the handler now derives the await timeout from the Lambda's own remaining time - fixed in [4db4add](https://github.com/guardian/security-hq/pull/1416/commits/4db4addcf4b7382552a098a22041c6d129a038f1)

core/src/main/scala/logic/IamUnrecognisedUsers.scala
```diff
@@ -101,34 +101,46 @@ object IamUnrecognisedUsers extends LazyLogging {
...
+      )
+      Attempt.Right(Nil)
+    }
   }
 
   def removeAccountPasswords(
       accountUnrecognisedUsers: AccountUnrecognisedUsers,
-      iamClients: AwsClients[IamAsyncClient]
+      iamClients: AwsClients[IamAsyncClient],
+      dryRun: Boolean = false
```
aug24, line 143: We don't appear to be using dryrun here?

aug24 (approved):

core/src/main/scala/logic/IamUnrecognisedUsers.scala
tjsilver: Good spot, will add that now!

tjsilver:
@aug24 reviewed the last change outside this PR and was happy with it.

PR #1418, opened 2026-07-09
Put template in the location expected by generated riff raff yaml

aug24:
## What does this change?

The cdk generated riff raff yaml is expected to be in its default location (cdk.out).  Our existing deploy moved it to a different location.  This PR puts it back in the default location.


kelvin-chappell (approved):
👍

PR #1419, opened 2026-07-14
Iam outdated credentials lambda

aug24:
## What does this change?

This PR extracts the functionality required by the new lambda version into its own build artifact.

This is suitable to be deployed for dry run testing, and to serve as a pattern for similar work.

There is at least one TODO in resource handling which need to be discussed as a team: clients are created and not closed (the lambda host OS will close them).  This is the same approach as used in the app, which would not have picked up new config after initial deploy.  We can improve on this once we have the lambda code split out.


iam-outdated-credentials/src/main/resources/logback.xml
```diff
@@ -0,0 +1,38 @@
+<!-- https://www.playframework.com/documentation/latest/SettingsLogger -->
+<configuration>
+
+  <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+    <file>${application.home:-.}/logs/application.log</file>
```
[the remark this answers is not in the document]
aug24: Rebased out of existence.

hq/app/services/IamRemediationService.scala
```diff
@@ -101,26 +101,27 @@ class IamRemediationService(
           (now.getHourOfDay == 14 && now.getMinuteOfHour == 0)
```
tjsilver, line 101: Since the changes to this file and appComponents.scala are to the hq app and not directly used/relevant here, could these changes be moved to a separate PR?

aug24: This is necessary solely because the previous PR ended up returning Attempt[Attempt[Unit]] which means even waiting on the (outer) Attempt doesn't guarantee the code completes.  So it's a fix to a regression.

tjsilver (approved):
👍

PR #1423, opened 2026-07-16
Change Lambda handler to ignore event

aug24:
## What does this change?

The previous implementation assumed a useful event.  We do not, in fact use the event, and worse, we don't have anything in scope that can parse it.  So: don't.


tjsilver (approved):
👍

PR #1425, opened 2026-07-17
Add config read permissions for outdated-credentials lambda

aug24:
## What does this change?

This permission appears to be implicit for the app, but needs to be explicit for a lambda.

### DIFF
--- a/cdk/lib/security-hq.ts
+++ b/cdk/lib/security-hq.ts
@@ -119,9 +119,14 @@ export class SecurityHQ extends GuStack {
     dpkg -i /tmp/installer.deb`);
 
     const guPutCloudwatchMetricsPolicy = new GuPutCloudwatchMetricsPolicy(this);
-    const guGetS3ObjectsPolicy = new GuGetS3ObjectsPolicy(this, "S3AuditRead", {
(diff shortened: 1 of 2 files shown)


cdk/lib/security-hq.ts
```diff
@@ -119,10 +119,23 @@ export class SecurityHQ extends GuStack {
...
+    const guGetS3AuditObjectsPolicy = new GuGetS3ObjectsPolicy(
+      this,
+      "S3AuditRead",
+      {
+        bucketName: auditDataS3BucketName.valueAsString,
+        paths: [auditDataS3BucketPath],
+      },
+    );
+    const configS3BucketPath = `${this.stack}/${this.stage}/${this.app}/security-hq.conf`;
+    const guGetS3ConfigObjectsPolicy = new GuGetS3ObjectsPolicy(
```
kelvin-chappell: There's a `loadAccountConfigPolicy` already defined later in this file.  Nice to use the `GuGetS3ObjectsPolicy` construct but maybe we can put these things together

kelvin-chappell (approved):
👍

PR #1426, opened 2026-07-17
Add config read permissions for outdated credentials lambda

aug24:
## What does this change?

Correct format for resource.


kelvin-chappell (approved):
👍

PR #1427, opened 2026-07-17
Fix instant conversion

aug24:
## What does this change?

We have an error in the existing app, during the outdated credentials job, at this line:

```
	at aws.iam.IAMClient$.listUserAccessKeys$$anonfun$1$$anonfun$2$$anonfun$1$$anonfun$1(IAMClient.scala:162)
```

Extraordinarily, it isn't failing during the lambda execution of the outdated credentials job.


kelvin-chappell (approved):
👍

PR #1428, opened 2026-07-17
Disable job pending discussion

aug24:
## What does this change?

The fix to ensure these jobs run has had unintended consequences in that the job ran and disable keys without warning for a week first.  Notifications were sent, but unhelpfully it was on Friday afternoon.

Jorge, Akash, Jon and Justin have been discussing and it was decided to ensure this cannot happen again until we have a plan.


jorgeazevedo (approved):
Looks good!

PR #1429, opened 2026-07-20
Slow down dependency updates

NovemberTang:
## What does this change?

Reduces the frequency of dependency updates

## What is the value of this?

Less PR pressure. Less fatigue on teams.

No

No

## Any additional notes?

No

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,4 +1,1 @@
 pullRequests.frequency = "90 days"
-updates.cooldown = {
-  minimumAge: "7 days"
-}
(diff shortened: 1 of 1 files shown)


.scala-steward.conf
```diff
@@ -1 +1,4 @@
-pullRequests.frequency = "30 days"
+pullRequests.frequency = "90 days"
+updates.cooldown = {
+  minimumAge: "7 days"
+}
```
kelvin-chappell: This section is actually redundant as the [Scala Steward workflow already has it in its config](https://github.com/guardian/scala-steward-public-repos/blob/ab43ae408d89d67675d8ed56cb97718b5c81c461/scala-steward.conf#L25-L27).

kelvin-chappell (approved):
👍

PR #1430, opened 2026-07-20
Add dryrun settings for remediation within app

aug24:
## What does this change?

We found on Friday (17/7/2026) that the remediation job was broken, and fixed it.  This resulted in the immediate deactivation of 38 keys, without prior warning.

Before turning this job back on _either in the app or the lambda_, we need to ensure that this will not be repeated.

We want to be able to run both the jobs in dry run mode (in either implementation) to output the plan without actually actioning it.

### DIFF
--- a/core/src/main/scala/logic/IamUnrecognisedUsers.scala
+++ b/core/src/main/scala/logic/IamUnrecognisedUsers.scala
@@ -102,7 +102,7 @@ object IamUnrecognisedUsers extends LazyLogging {
   def disableAccountAccessKeys(
       accountUnrecognisedKeys: AccountUnrecognisedAccessKeys,
       iamClients: AwsClients[IamAsyncClient],
-      dryRun: Boolean = false
+      dryRun: Boolean
   )(implicit ec: ExecutionContext): Attempt[List[UpdateAccessKeyResponse]] = {
     if (!dryRun) {
       val AccountUnrecognisedAccessKeys(account, accessKeys) = accountUnrecognisedKeys
@@ -140,7 +140,7 @@ object IamUnrecognisedUsers extends LazyLogging {
   def removeAccountPasswords(
       accountUnrecognisedUsers: AccountUnrecognisedUsers,
       iamClients: AwsClients[IamAsyncClient],
-      dryRun: Boolean = false
+      dryRun: Boolean
   )(implicit ec: ExecutionContext): Attempt[List[Option[DeleteLoginProfileResponse]]] = {
     if (!dryRun) {
       val results = Attempt.traverse(accountUnrecognisedUsers.unrecognisedUsers)(user =>
@@ -169,7 +169,7 @@ object IamUnrecognisedUsers extends LazyLogging {
 
   def unrecognisedUserNotifications(
       accountUsers: List[AccountUnrecognisedUsers],
-      dryRun: Boolean = false
+      dryRun: Boolean
   ): List[Notification] = {
     if (!dryRun) {
       accountUsers.flatMap { case AccountUnrecognisedUsers(account, users) =>
--- a/core/src/main/scala/model/iamremediation.scala
+++ b/core/src/main/scala/model/iamremediation.scala
@@ -81,5 +81,6 @@ case class UnrecognisedJobConfigProperties(
     janusDataFileKey: String,
     janusUserBucket: String,
     securityAccount: AwsAccount,
-    anghammaradSnsTopicArn: String
+    anghammaradSnsTopicArn: String,
+    dryRun: Boolean
 )
--- a/hq/app/config/Config.scala
+++ b/hq/app/config/Config.scala
@@ -5,7 +5,7 @@ import com.gu.googleauth.{AntiForgeryChecker, GoogleAuthConfig, GoogleGroupCheck
 import com.gu.play.secretrotation.aws.parameterstore.{AwsSdkV2, SecretSupplier}
 import com.gu.play.secretrotation.{SnapshotProvider, TransitionTiming}
 import model.*
-import play.api.Configuration
+import play.api.{Configuration, Logging}
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.ssm.SsmClient
 import utils.attempt.{Attempt, FailedAttempt, Failure}
@@ -16,7 +16,7 @@ import scala.concurrent.ExecutionContext
 import scala.jdk.CollectionConverters.*
 import scala.util.Try
 
-object Config {
+object Config extends Logging {
   val app = "security-hq"
 
   val documentationLinks: List[Documentation] = List(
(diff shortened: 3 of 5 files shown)


core/src/main/scala/logic/IamUnrecognisedUsers.scala
```diff
@@ -102,7 +102,7 @@ object IamUnrecognisedUsers extends LazyLogging {
   def disableAccountAccessKeys(
       accountUnrecognisedKeys: AccountUnrecognisedAccessKeys,
       iamClients: AwsClients[IamAsyncClient],
-      dryRun: Boolean = false
+      dryRun: Boolean
```
aug24, line 105: Do not default here.

core/src/main/scala/logic/IamUnrecognisedUsers.scala
```diff
@@ -140,7 +140,7 @@ object IamUnrecognisedUsers extends LazyLogging {
   def removeAccountPasswords(
       accountUnrecognisedUsers: AccountUnrecognisedUsers,
       iamClients: AwsClients[IamAsyncClient],
-      dryRun: Boolean = false
+      dryRun: Boolean
```
aug24, line 143: Or here

core/src/main/scala/logic/IamUnrecognisedUsers.scala
```diff
@@ -169,7 +169,7 @@ object IamUnrecognisedUsers extends LazyLogging {
 
   def unrecognisedUserNotifications(
       accountUsers: List[AccountUnrecognisedUsers],
-      dryRun: Boolean = false
+      dryRun: Boolean
```
aug24, line 172: Or here.

We need to ensure that all these are the same value, so make them explicit and pass them from a single value higher up.

core/src/main/scala/model/iamremediation.scala
```diff
@@ -81,5 +81,6 @@ case class UnrecognisedJobConfigProperties(
     janusDataFileKey: String,
     janusUserBucket: String,
     securityAccount: AwsAccount,
-    anghammaradSnsTopicArn: String
+    anghammaradSnsTopicArn: String,
+    dryRun: Boolean
```
aug24, line 85: Add that single value to the config for this job.

hq/app/config/Config.scala
```diff
@@ -122,7 +122,8 @@ object Config {
       bucket <- getIamUnrecognisedUserBucket(config)
       securityAccount <- getSecurityAccount(config)
       anghammaradSnsTopicArn <- getAnghammaradSNSTopicArn(config)
-    } yield UnrecognisedJobConfigProperties(accounts, key, bucket, securityAccount, anghammaradSnsTopicArn)
+      dryRun <- getUnrecognisedUserDryRun(config)
```
aug24: Get the value, specific to this job (unrecognised users - one half of remediation).

hq/app/services/IamRemediationService.scala
```diff
@@ -71,11 +71,11 @@ class IamRemediationService(
         listAccountAccessKeys(_, iamClients)
       )
       // disable each access key for unrecognised users
-      _ <- Attempt.traverse(unrecognisedUserAccessKeys)(disableAccountAccessKeys(_, iamClients))
+      _ <- Attempt.traverse(unrecognisedUserAccessKeys)(disableAccountAccessKeys(_, iamClients, config.dryRun))
```
aug24: Pass the same value in in all three places.

hq/app/services/IamRemediationService.scala
```diff
@@ -101,23 +101,23 @@ class IamRemediationService(
...
   private def disableOutdatedCredentials(): Attempt[Unit] = for {
+    dryRun <- Config.getOutdatedCredentialsDryRun(config)
     notificationTopicArn <- getAnghammaradSNSTopicArn(config)
     tableName <- getIamDynamoTableName(config)
     serviceAccountIds <- Config.getAccountsForIamRemediationService(config)
     rawCredsReports = cacheService.getAllCredentials
     // this tells us which AWS accounts we are allowed to make changes to
     allowedAwsAccountIds <- Config.getAllowedAccountsForStage(config)
-    result <- IamOutdatedCredentials(snsClient, iamClients, dynamo, dryRun = false).disableOutdatedCredentials(
+    result <- IamOutdatedCredentials(snsClient, iamClients, dynamo, dryRun).disableOutdatedCredentials(
```
aug24: Don't hard code to false.  Use the value from config (line 113 above).

hq/app/config/Config.scala
```diff
@@ -134,6 +135,20 @@ object Config {
     )
   }
 
+  def getUnrecognisedUserDryRun(config: Configuration): Attempt[Boolean] = {
+    Attempt.Right(
+      // Not using toBoolean because we want to default to true (do nothing) if the config is missing or invalid
+      config.getOptional[String]("unrecognisedUser.dryRun").getOrElse("true").equalsIgnoreCase("true")
+    )
+  }
+
```
aug24, line 147: Get the config items separately (^^ and vv) 

ONLY do the job if the provided dryRun string is exactly "false" (ignoring case).

hq/app/config/Config.scala
```diff
@@ -134,6 +135,26 @@ object Config {
     )
   }
 
+  def getUnrecognisedUserDryRun(config: Configuration)(implicit executionContext: ExecutionContext): Attempt[Boolean] = {
+    Attempt.Right(
+      // Default to true; only an explicit "false" disables dry-run.
+      // Not using toBoolean because we want to default to true (do nothing) if the config is missing or invalid
+      !config.getOptional[String]("unrecognisedUser.dryRun").getOrElse("true").equalsIgnoreCase("false")
```
kelvin-chappell:
```suggestion
      config.getOptional[String]("unrecognisedUser.dryRun").exists(_.equalsIgnoreCase("false"))
```

kelvin-chappell (approved):
👍

hq/app/config/Config.scala
```diff
@@ -134,6 +134,27 @@ object Config {
...
 
+  // Default to true; only an explicit "false" disables dry-run.
+  // Not using toBoolean because we want to default to true (do nothing) if the config is missing or invalid
+  private[config] def getDryRun(config: Configuration, key: String) =
+    !config.getOptional[String](key).exists(_.equalsIgnoreCase("false"))
+
+  def getUnrecognisedUserDryRun(
+      config: Configuration
+  )(implicit executionContext: ExecutionContext): Attempt[Boolean] = {
+    Attempt
```
adamnfish-gu: Why `Attempt.tap` a pure operation?

I'd remove the Attempt and just add a log line to this function, we don't need to introduce the possibility of asynchronicity here. If we're using Attempt to wrap logging everywhere then maybe that's a different story, but I think mostly we just emit logs where needed?

hq/app/config/Config.scala
```diff
@@ -134,6 +134,27 @@ object Config {
     )
   }
 
+  // Default to true; only an explicit "false" disables dry-run.
+  // Not using toBoolean because we want to default to true (do nothing) if the config is missing or invalid
+  private[config] def getDryRun(config: Configuration, key: String) =
```
adamnfish-gu, line 139: Is this `getDryRun` or `readAsBooleanWithDefault`? This is especially jarring in the tests where we're reading random keys and asserting the Boolean result.

hq/app/config/Config.scala
aug24: We are using Attempt to wrap other values, but here we don't actually need to because we effectively recover with dry-run set to true for any bad or missing value.  I've removed Attempt altogether.

hq/app/config/Config.scala
aug24: I've made it more explicit and moved the sub-key into the function.  This is specific to dry run settings.

PR #1432, opened 2026-07-20
Restrict last action to preceding two weeks

aug24:
## What does this change?

This change adds a test to cover the scenario where a key has had a final warning, but then for whatever reason the job has not run for a while.  This should not then result in an immediate de-activation of the key.  This test fails with the current code.

An additional filter has been added requiring that the last action was within fourteen days of the current action decision.  Following this addition, the new test passes.

This has required more changes to test code because all tests of this code must now pass in the nominal date (`date`) on which the action is being considered.

### DIFF
--- a/hq/app/services/IamRemediationService.scala
+++ b/hq/app/services/IamRemediationService.scala
@@ -6,8 +6,9 @@ import com.gu.janus.JanusConfig
 import config.Config
 import config.Config.{getAnghammaradSNSTopicArn, getIamDynamoTableName, getIamUnrecognisedUserConfig}
 import db.IamRemediationDb
-import logic.IamUnrecognisedUsers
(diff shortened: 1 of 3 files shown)


kelvin-chappell (approved):
Good idea! 👍 
Assuming that the wiring up of the dry-run value in CDK will come in a future PR

aug24:
> Good idea! 👍 Assuming that the wiring up of the dry-run value in CDK will come in a future PR

The dry run value for the app runs is in the config file, which is manually edited.  We didn't want to reuse it for the lambdas because we need two independent processes for controlling the app runs and the lambda runs.

Dry run app values default to true so we don't need to set it, and the code that reads it was already merged and confirmed to be correct.

PR #1433, opened 2026-07-21
Reactivate app jobs

aug24:
## What does this change?

Now we have confirmed the dry run mode is true for both jobs, we can remove the comment markers and allow the jobs to run.


tjsilver (approved):
👍

PR #1436, opened 2026-07-22
Clean up unused AWS clients

andrew-nowak:
## What does this change?

Quickly cleans up a couple of AWS clients which aren't used in practice. Includes the removal of a now unnecessary SDK module!


core/src/main/scala/aws/AWS.scala
```diff
@@ -69,9 +68,6 @@ object AWS {
         c.advancedOption(SdkAdvancedAsyncClientOption.FUTURE_COMPLETION_EXECUTOR, newCachedThreadPool())
       )
```
aug24, line 69: Do we still need to import Ec2AsyncClient and the lib?

andrew-nowak: Nope - import removed (the ec2 lib is still used elsewhere in both `hq` and `core`, so I can't remove the dependency)

NovemberTang (approved):

PR #1437, opened 2026-07-23
Update devenv and settings

aug24:
## What does this change?

Bring the devenv version up to date and add copilot to the project.


kelvin-chappell (approved):
👍

PR #1440, opened 2026-07-23
Add notification to devx security on disable

aug24:
## What does this change?

This PR adds an additional notification to DevX/Security whenever a credential is de-activated.

This provides us with a helpful heads-up that we are likely to have a team somewhere with a broken process.

## Any additional notes?

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1216793019549897

### DIFF
--- a/core/src/main/scala/aws/iam/IAMClient.scala
+++ b/core/src/main/scala/aws/iam/IAMClient.scala
@@ -19,7 +19,7 @@ import scala.jdk.CollectionConverters.*
 
 object IAMClient extends LazyLogging {
 
-  val SOLE_REGION = Region.of("us-east-1")
+  private val SOLE_REGION = Region.of("us-east-1")
 
   private def generateCredentialsReport(
       client: AwsClient[IamAsyncClient]
@@ -74,7 +74,7 @@ object IAMClient extends LazyLogging {
     }
   }
 
-  def getCredentialReportDisplay(
+  private def getCredentialReportDisplay(
       account: AwsAccount,
       currentData: Either[FailedAttempt, CredentialReportDisplay],
       cfnClients: AwsClients[CloudFormationAsyncClient],
@@ -159,7 +159,12 @@ object IAMClient extends LazyLogging {
                 )
               }
           }
-        } yield CredentialMetadata(akm.userName, akm.accessKeyId, new DateTime(akm.createDate.toEpochMilli), credentialStatus)
+        } yield CredentialMetadata(
+          akm.userName,
+          akm.accessKeyId,
+          new DateTime(akm.createDate.toEpochMilli),
+          credentialStatus
+        )
       }
     } yield credentialMetadatas
   }
--- a/core/src/main/scala/aws/iam/IAMClient.scala
+++ b/core/src/main/scala/aws/iam/IAMClient.scala
@@ -19,7 +19,7 @@ import scala.jdk.CollectionConverters.*
 
 object IAMClient extends LazyLogging {
 
-  private val SOLE_REGION = Region.of("us-east-1")
+  val SOLE_REGION = Region.of("us-east-1")
 
   private def generateCredentialsReport(
       client: AwsClient[IamAsyncClient]
@@ -74,7 +74,7 @@ object IAMClient extends LazyLogging {
     }
   }
 
-  private def getCredentialReportDisplay(
+  def getCredentialReportDisplay(
       account: AwsAccount,
       currentData: Either[FailedAttempt, CredentialReportDisplay],
       cfnClients: AwsClients[CloudFormationAsyncClient],
@@ -159,12 +159,7 @@ object IAMClient extends LazyLogging {
                 )
               }
           }
-        } yield CredentialMetadata(
-          akm.userName,
-          akm.accessKeyId,
-          new DateTime(akm.createDate.toEpochMilli),
-          credentialStatus
-        )
+        } yield CredentialMetadata(akm.userName, akm.accessKeyId, new DateTime(akm.createDate.toEpochMilli), credentialStatus)
       }
     } yield credentialMetadatas
   }
--- a/core/src/main/scala/notifications/AnghammaradNotifications.scala
+++ b/core/src/main/scala/notifications/AnghammaradNotifications.scala
@@ -36,8 +36,8 @@ object AnghammaradNotifications extends LazyLogging {
       }
   }
 
-  val channel = Preferred(Email)
-  val sourceSystem = "Security HQ Credentials Notifier"
+  private val channel = Preferred(Email)
+  private val sourceSystem = "Security HQ Credentials Notifier"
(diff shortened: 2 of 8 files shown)


core/src/main/scala/aws/iam/IAMClient.scala
```diff
@@ -19,7 +19,7 @@ import scala.jdk.CollectionConverters.*
 
 object IAMClient extends LazyLogging {
 
-  val SOLE_REGION = Region.of("us-east-1")
+  private val SOLE_REGION = Region.of("us-east-1")
```
aug24: Just clearing warnings.

hq/app/services/IamRemediationService.scala
```diff
@@ -37,7 +38,8 @@ class IamRemediationService(
     iamClients: AwsClients[IamAsyncClient],
     lifecycle: ApplicationLifecycle,
     environment: Environment,
-    securityS3Client: S3Client
+    securityS3Client: S3Client,
+    devXSecurityAccountMaybe: Option[AwsAccount]
```
aug24: This _will_ be populated if there is a configured account called "security", but we have to cope with the possibility there isn't.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -33,13 +34,13 @@ class IamOutdatedCredentials(
     *   - disable an IAM credential and send a notification that this has been done
     *   - remove an IAM password and send a notification that this has been done
     */
-  private def performRemediationOperation(
+  private[logic] def performRemediationOperation(
```
aug24, line 40: Make accessible from tests.  Don't return a single sns id string, return a list of sns id strings, because we can now have from zero to two.

I've made the dryRun option return an empty list, rather than a fake value, as I think it's more meaningful.

script/setup
```diff
@@ -44,6 +44,7 @@ downloadConfig() {
     # BSD sed (macOS)
     sed -i '' 's|"/Users/ADD_YOUR_NAME|${HOME}"|g' ~/.gu/security-hq/security-hq.local.conf
   fi
+}
```
aug24, line 48: Little bugfix I found and sneaked in.

iam-outdated-credentials/src/test/scala/logic/IamOutdatedCredentialsTest.scala
```diff
@@ -737,6 +752,152 @@ class IamOutdatedCredentialsTest extends AnyFreeSpec with Matchers with OptionVa
     }
   }
 
+  "performRemediationOperation" - {
```
aug24, line 755: This whole section was rather under tested.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -93,14 +102,30 @@ class IamOutdatedCredentials(
             iamClients
           )
           // send a notification to say this is what we have done
-          notificationId <- AnghammaradNotifications.send(notification, notificationTopicArn, snsClient)
+          userNotificationId <- AnghammaradNotifications.send(notification, notificationTopicArn, snsClient)
+          // Send a notification to devx too, if we've got an account for them.
+          securityNotificationIdMaybe <- sendSecurityNotification(notificationTopicArn, notificationDevXSecurityMaybe)
```
kelvin-chappell: If this fails, presumably we lose the status update before it's written to Dynamo.  So perhaps the write should happen first

aug24: Excellent point.  I've reconsidered throughout and re-ordered where I think it's important.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -33,13 +35,12 @@ class IamOutdatedCredentials(
     *   - disable an IAM credential and send a notification that this has been done
     *   - remove an IAM password and send a notification that this has been done
     */
-  private def performRemediationOperation(
+  private[logic] def performRemediationOperation(
```
kelvin-chappell, line 40: The logic of this method is quite difficult to follow now.  I wonder if extracting the body of each case into its own method would make the structure clearer.

aug24: Done - and also pulled up notification topic, and table name into the object as they are being passed around a lot, and removed dryrun from the method signatures as it's also not needed.

core/src/main/scala/notifications/AnghammaradNotifications.scala
```diff
@@ -36,8 +36,8 @@ object AnghammaradNotifications extends LazyLogging {
       }
   }
 
-  val channel = Preferred(Email)
-  val sourceSystem = "Security HQ Credentials Notifier"
+  private val channel = Preferred(Email)
```
aug24, line 39: Just getting rid of warnings as I'm in the file.

core/src/main/scala/notifications/AnghammaradNotifications.scala
```diff
@@ -118,6 +118,37 @@ object AnghammaradNotifications extends LazyLogging {
     )
   }
 
+  def outdatedCredentialRemediationDevXSecurity(
```
aug24, line 121: Construct the new warning for the security team.

hq/app/config/Config.scala
```diff
@@ -105,12 +105,7 @@ object Config extends Logging {
     } yield AwsAccount(id, name, roleArn, number)
   }
 
-  def getIamDynamoTableName(config: Configuration): Attempt[String] = {
-    Attempt.fromOption(
-      config.getOptional[String]("alert.iamDynamoTableName"),
-      FailedAttempt(Failure("unable to get dynamo table name", "unable to get dynamo table name for IAM jobs", 500))
-    )
-  }
+  def getIamDynamoTableName(config: Configuration): String = requiredString(config, "alert.iamDynamoTableName")
```
aug24, line 108: There's no need for this to be an attempt.  It's better if the SHQ deployment fails.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -55,52 +56,129 @@ class IamOutdatedCredentials(
 
     (remediationOperation.iamRemediationActivityType, remediationOperation.iamProblem) match {
```
aug24, line 57: This could simplify further later: we only have one type of iamProblem!  We really should either get rid of this or consider what else we might choose to detect.

kelvin-chappell (approved):
🚀

PR #1441, opened 2026-07-27
Cache and reuse constructed AWS clients

andrew-nowak:
## What does this change?

As in the title - create a cache of constructed sdk clients, so we can avoid unnecessary repeated construction.

## What is the value of this?

This should be more performant

### DIFF
--- a/core/src/main/scala/aws/AWS.scala
+++ b/core/src/main/scala/aws/AWS.scala
@@ -3,6 +3,7 @@ package aws
 import config.CoreConfig
 import model.AwsAccount
 import software.amazon.awssdk.auth.credentials.{AwsCredentialsProviderChain, ProfileCredentialsProvider}
+import software.amazon.awssdk.awscore
 import software.amazon.awssdk.awscore.client.builder.{AwsAsyncClientBuilder, AwsClientBuilder}
 import software.amazon.awssdk.core.client.config.SdkAdvancedAsyncClientOption
 import software.amazon.awssdk.regions.Region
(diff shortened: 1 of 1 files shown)


core/src/main/scala/aws/iam/IAMClient.scala
```diff
@@ -159,7 +159,12 @@ object IAMClient extends LazyLogging {
                 )
               }
           }
-        } yield CredentialMetadata(akm.userName, akm.accessKeyId, new DateTime(akm.createDate.toEpochMilli), credentialStatus)
+        } yield CredentialMetadata(
```
aug24: If not changing the file, don't format the file?

core/src/main/scala/aws/AWS.scala
```diff
@@ -46,39 +61,59 @@ object AWS {
...
+    client(withCustomThreadPool(SupportAsyncClient.builder), account, region)
+  private val taClientCache = new AwsClientCache(taClientBuilder)
 
   // Only needs Regions.US_EAST_1
   def taClients(accounts: List[AwsAccount], region: Region = Region.of("us-east-1")): AwsClients[SupportAsyncClient] =
-    clients(withCustomThreadPool(SupportAsyncClient.builder), accounts, region)
+    taClientCache.getClients(accounts, List(region))
+
+  private def s3ClientBuilder(account: AwsAccount, region: Region): S3Client =
+    client(S3Client.builder(), account, region)
```
aug24, line 101: No custom thread pool here?

core/src/main/scala/aws/AWS.scala
```diff
@@ -46,39 +61,59 @@ object AWS {
...
-    for {
-      account <- accounts
-      region <- regionList
-      client = builder
-        .credentialsProvider(credentialsProvider(account))
-        .region(region)
-        .build()
-    } yield AwsClient(client, account, region)
-  }
+  private def client[A, B <: AwsClientBuilder[B, A]](
```
aug24, line 64: We could enforce the custom thread pool by pulling it up to here, rather than referencing it in several places below.

core/src/main/scala/aws/AWS.scala
andrew-nowak: This is using the synchronous client so far, I've chosen not to change it to the async client as it's currently only used by the app, so presumably will soon be a candidate for deletion

core/src/main/scala/aws/AWS.scala
andrew-nowak: I had a try at exactly this, but I couldn't quite get it to work. The option to configure the threadpool belongs to `AwsAsyncClientBuilder`, and the options to configure the region and credentials provider belong to `AwsClientBuilder` (ie. non-async builder), and the only existing types which are subtypes of both interfaces are the concrete SDK client builders (ie. `IamAsyncClientBuilder` etc.). I also can't see a way to require `B` to be a subtype of both generic builder variants. Lacking all of that, the best alternative I was able to find was to continue the previous implementation's pattern of having a separate method for configuring the async options from the sync options.

aug24: That rings a bell!  There's no appropriate marker interface for only async clients.  I hit the same problem.

aug24 (approved):

PR #1443, opened 2026-07-27
Restrict remediation events to Tuesdays

aug24:
## What does this change?

This PR adds another decision step to the remediation pathway.  Remediations will only take place when the day of the week is Tuesday.

However, notifications to the security team will be sent even if remediation will be skipped, for maximum visibility.

### DIFF
--- a/core/src/main/scala/notifications/AnghammaradNotifications.scala
+++ b/core/src/main/scala/notifications/AnghammaradNotifications.scala
@@ -118,6 +118,37 @@ object AnghammaradNotifications extends LazyLogging {
     )
   }
 
+  def outdatedCredentialNoRemediationDevXSecurity(
+      awsAccount: AwsAccount,
+      iamUser: IAMUser,
+      problemCreationDate: DateTime,
+      devXSecurityAccount: AwsAccount
+  ): Notification = {
+    val endUserNotificationTargets = notificationTargets(awsAccount, iamUser)
+    val devxSecurityNotificationTargets = List(Account(devXSecurityAccount.accountNumber))
+    val endUserTargetsString = endUserNotificationTargets.map(_.toString).mkString(", ")
+    val message =
+      s"""
+         |The permanent credential, ${iamUser.username}, in ${awsAccount.name} was eligible for deactivation today,
+         |because it was last rotated on ${printDay(problemCreationDate)}.
+         |
+         |It wasn't deactivated because it's not Deactivation Tuesday.
+         |
+         |THIS ACTION HAS HIGH POTENTIAL TO BREAK THINGS.
+         |
+         |BE PREPARED FOR USERS TO BE UPSET!
+         |""".stripMargin
+    val subject = s"DISABLED long-lived credential in ${awsAccount.name}"
+    Notification(
+      subject,
+      message + genericOutdatedCredentialText,
+      Nil,
+      devxSecurityNotificationTargets,
+      channel,
+      sourceSystem
+    )
+  }
+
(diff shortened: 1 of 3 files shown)


core/src/main/scala/notifications/AnghammaradNotifications.scala
```diff
@@ -118,6 +118,37 @@ object AnghammaradNotifications extends LazyLogging {
     )
   }
 
+  def outdatedCredentialNoRemediationDevXSecurity(
```
aug24, line 121: Just create a new message.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -68,17 +68,34 @@ class IamOutdatedCredentials(
...
 
   private def remediation(
+      now: DateTime,
       awsAccount: AwsAccount,
       iamUser: IAMUser,
       problemCreationDate: DateTime,
       thisRemediationActivity: IamRemediationActivity
   )(implicit ec: ExecutionContext) = {
-    if (!dryRun) {
+    if (dryRun) {
```
aug24, line 82: Change logic flow for clarity, and add the Tuesday check.

iam-outdated-credentials/src/test/scala/logic/IamOutdatedCredentialsTest.scala
```diff
@@ -767,7 +766,7 @@ class IamOutdatedCredentialsTest extends AnyFreeSpec with Matchers with OptionVa
       override def close(): Unit = ()
     }
 
-    val fakeRemediationIamClient = new IamAsyncClient {
+    def getFakeRemediationIamClient(remediationCounter: AtomicInteger): IamAsyncClient = new IamAsyncClient {
```
aug24, line 776: Improve testing and prove that we only actually do the remediation on remediation tuesdays.

core/src/main/scala/notifications/AnghammaradNotifications.scala
```diff
@@ -118,6 +118,37 @@ object AnghammaradNotifications extends LazyLogging {
...
+         |The permanent credential, ${iamUser.username}, in ${awsAccount.name} was eligible for deactivation today,
+         |because it was last rotated on ${printDay(problemCreationDate)}.
+         |
+         |It wasn't deactivated because it's not Deactivation Tuesday.
+         |
+         |THIS ACTION HAS HIGH POTENTIAL TO BREAK THINGS.
+         |
+         |BE PREPARED FOR USERS TO BE UPSET!
+         |""".stripMargin
+    val subject = s"DISABLED long-lived credential in ${awsAccount.name}"
```
andrew-nowak: Subject line sounds like we _have_ disabled the credential, but actually we haven't


```suggestion
    val subject = s"Imminent disabling of long-lived credential in ${awsAccount.name}"
```

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
andrew-nowak: Why do we put the logic here, and not with the `identifyRemediationOperation` logic?

aug24:
> Why do we put the logic here, and not with the identifyRemediationOperation logic?

Because the decision making of what remediation operation applies (if any) is separate to the decision making of whether or not to actually do it.

andrew-nowak (approved):

PR #1444, opened 2026-07-27
chore: Small improvements for running iam-unrecognised-users locally

kelvin-chappell:
The small improvements are:
* Made it clearer which logging output is part of the dry run
* Removed the redundant `restrictToAccountId` param as the dev config file only contains the Security account anyway
* Passed the execution context through from `Main` and `Handler` and using Scala 3 terminology ('using' instead of 'implicit', without a value name)
* Added start script for the lambda that supplies the env vars, which are already referred to in the repo so not giving anything away
* Added `deprecation` scalac option to both lambdas as it isn't provided to them by the Play plugin
* Granted the `RunSecurityHqLocallyPolicy` dev policy permission to list the contents of the audit bucket.

To support this, I've added a dev Janus data file to the S3 Dev path.  It has a missing user so that we can see effects without breaking anything in production.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -159,6 +159,7 @@ lazy val iamOutdatedCredentials = (project in file("iam-outdated-credentials"))
   .dependsOn(core % "compile->compile;test->test")
   .settings(
     name := """iam-outdated-credentials""",
+    scalacOptions += "--deprecation",
     fileDescriptorLimit := Some("16384"), // This increases the number of open files allowed when running in AWS
     Assets / pipelineStages := Seq(digest),
     // exclude docs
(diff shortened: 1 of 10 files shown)


tjsilver (commented):
I tried testing this locally with the `Run Security HQ lambdas locally` credential but got an error. I tried also with `readonly` and `developer` permissions, and in both cases observed one line of the `DRY RUN` messaging (which was repeated by `INFO`): 
```
DRY RUN: Would send 0 notification(s).
15:14:38.091 [sbt-bg-threads-1] INFO  unrecognised.UnrecognisedUsers$ - Successfully ran unrecognised user job and sent 0 notification(s).
```
There was a lot of noisy output in the logs with the error `Unable to load credentials from any of the providers in the chain` which obscured any other `DRY RUN` messaging  I think. 

Have messaged @kelvin-chappell with more details of the exact log lines.

tjsilver (approved):
Tested locally using the start script and worked as expected. Might be good to update the [README](https://github.com/guardian/security-hq/blob/main/iam-unrecognised-users/README.md#running-locally) for local running, but that can be done in a follow-up.

PR #1445, opened 2026-07-27
Only construct a single IAM client per account

andrew-nowak:
## What does this change?

IAM is a "global" service, and only operates in the us-east-1 region. We construct IAM clients for every region, but then only select and use the us-east-1 ones. Skip constructing the unnecessary and unused clients

## What is the value of this?

Small performance improvement, hopefully less confusion


kelvin-chappell (approved):
👍

PR #1449, opened 2026-07-29
Add generate credential report policy to dev policy

aug24:
## What does this change?

Adds iam:GenerateCredentialReport to the dev policy used to run lambdas locally.

This is not required for the "real" lambda because it obtains the same privilege, per target account, via an assume role.


kelvin-chappell (approved):
👍

PR #1450, opened 2026-07-29
Remove SSM dev policy in favour of generic .ssh Janus policy

aug24:
## What does this change?

Removes an old developer policy which was used to explore ssm access to instances but is now unused.


kelvin-chappell (approved):
👍

PR #1452, opened 2026-07-29
Remove cloudformation enrichment of user data

andrew-nowak:
## What does this change?

Remove all Cloudformation API calls, client construction and data enrichment.

## What is the value of this?

I did some digging into exactly what SHQ is doing with the Cloudformation data it fetches; I've come to the conclusion that it is almost certainly doing _nothing_.

The Cloudformation data is fetched to enrich any discovered IAM users, with the stack they're associated with if one was found. The discovered stack travels with the constructed User objects, but I can't find any usage. When this functionality was added, the data was used to enrich the SHQ dashboard, with which users belonged to a stack and a helpful link to that stack (see screenshots in #99). Now that those dashboards have been decommissioned, nothing else is accessing the data, so I believe we can remove it entirely. This allows us to massively decrease the amount of API requests we're making which should decrease runtime too, and simplify the logic and should allow us to completely sidestep the problem of different regions enabled in different accounts.


kelvin-chappell (approved):
👍 
Can [calculateAvailableRegions](https://github.com/guardian/security-hq/blob/6ad7d1fdd3000c7bdf86b67e8955e98198b41b25/core/src/main/scala/config/CoreConfig.scala#L35) and code behind it also go?
And there will be some CDK permissions that are no longer needed.

andrew-nowak:
> Can [calculateAvailableRegions](https://github.com/guardian/security-hq/blob/6ad7d1fdd3000c7bdf86b67e8955e98198b41b25/core/src/main/scala/config/CoreConfig.scala#L35) and code behind it also go? 
> And there will be some CDK permissions that are no longer needed.

Great points, thank you! Yes `calculateAvailableRegions` can go; the code behind it can't quite go yet as it's still in use by the app. The unnecessary permissions are in the role created by the stack set, which I've updated

hq/app/AppComponents.scala
```diff
@@ -53,7 +53,6 @@ class AppComponents(context: Context)
   private val stage = Config.getStage(configuration)
 
   // the aim of this is to get all the regions that are available to this account
-  // TODO: Delete this and only use calculateAvailableRegions which does not await.
```
andrew-nowak, line 56: removed this TODO, as `calculateAvailableRegions` no longer exists and this problem is better addressed by removing the app wholesale, which doesn't need a TODO

PR #1453, opened 2026-07-29
Add GetCredentialReport to the dev policy

aug24:
## What does this change?

Adds another required permission to the dev policy.


kelvin-chappell (approved):
👍

PR #1454, opened 2026-07-29
Add more permissions to lambda dev-policy

aug24:
## What does this change?

Adds all the read permissions I think we need, taken from watched-account.template.yaml.
Write permissions are deliberately not included.  If you want to do writes, you can take .admin permissions.


tjsilver (approved):
👍

PR #1458, opened 2026-07-30
Clean up logging for deleted users

aug24:
## What does this change?

One known issue is that the credentials report can be slightly stale.  If we have deleted a user since the report was generated, we don't care that we can't get information on it.  Clean up the logs with an info instead of a warn+stacktrace


andrew-nowak (approved):

kelvin-chappell (approved):
👍

PR #1459, opened 2026-07-30
add iamOutdatedCredentials to the list of aggregated sbt projects

andrew-nowak:
## What does this change?

Add `iamOutdatedCredentials` to the `aggregate` list in the SBT config.

## What is the value of this?

The project will be included in any top level sbt commands, such as `compile`, `test`, `scalafmtAll` etc.


kelvin-chappell (approved):
Well spotted! 👍

PR #1463, opened 2026-07-30
Remove outdated credentials from app

aug24:
## What does this change?

We fully expect to have nothing left, but it helps to tidy up as we go.  This PR removes everything from the app that has moved to the iam outdated credentials lambda.

In order to achieve this, the error for unused code elements has been turned on, but filtered so it doesn't apply to generated code.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,6 +12,9 @@ ThisBuild / scalaVersion := "3.3.8"
 ThisBuild / scalacOptions ++= Seq(
   "-feature",
   "-no-indent", // don't support significant indentation
+  "-Wunused:all", // fail the build on unused imports, vals, params, and private members
+  "-Wconf:src=.*/twirl/.*&msg=.*unused.*:silent", // ignore dead-code warnings in generated Twirl sources
+  "-Wconf:src=.*/routes/.*&msg=.*unused.*:silent", // ignore dead-code warnings in generated Play routes sources
   "-Xfatal-warnings"
 )
 
--- a/core/src/main/scala/db/IamRemediationDb.scala
+++ b/core/src/main/scala/db/IamRemediationDb.scala
@@ -36,7 +36,7 @@ class IamRemediationDb(client: DynamoDbClient) {
     } yield result.responseMetadata.requestId
   }
 
-  private def scan(request: ScanRequest)(implicit ec: ExecutionContext): Attempt[List[Map[String, AttributeValue]]] = {
+  private def scan(request: ScanRequest): Attempt[List[Map[String, AttributeValue]]] = {
     try {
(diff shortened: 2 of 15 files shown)


build.sbt
```diff
@@ -12,6 +12,9 @@ ThisBuild / scalaVersion := "3.3.8"
 ThisBuild / scalacOptions ++= Seq(
   "-feature",
   "-no-indent", // don't support significant indentation
+  "-Wunused:all", // fail the build on unused imports, vals, params, and private members
```
adamnfish-gu, line 15: I'm not sure we want these committed? This sort of thing can be a real hassle during development, because it means you can't leave anything half-finished.

core/src/main/scala/db/IamRemediationDb.scala
```diff
@@ -52,23 +52,7 @@ class IamRemediationDb(client: DynamoDbClient) {
     }
   }
 
-  private def get(request: GetItemRequest)(implicit ec: ExecutionContext): Attempt[Map[String, AttributeValue]] = {
-    try {
```
adamnfish-gu, line 56: I don't think this is safe.

The `Attempt.Right` constructor expects a pure value, but we're giving it something that can fail. `client.getItem` is marked as being able to throw any of the following:

```scala
throws ProvisionedThroughputExceededException,
       ResourceNotFoundException, RequestLimitExceededException, InternalServerErrorException, ThrottlingException,
       AwsServiceException, SdkClientException, DynamoDbException
```

This highlights the risk of using this unused code tool without carefully considering the changes, especially where we're doing side-effecting work that involves Java libraries!

adamnfish-gu (commented):
It looks like this is catching a bunch of extra things? That's maybe fine, but given the error in it below, I think we want to be a bit careful.

Perhaps we could raise a more focused PR that uses this as a guide to make just the described change - removing the outdated credentials functionality?

core/src/main/scala/db/IamRemediationDb.scala
adamnfish-gu: My apologies, the side-by side diff made this look like a change rather than a removal.

All good if the whole method is going away!

adamnfish-gu (approved):
Apologies, I had misread the side-by-side diff.

This change is bigger than it needs to be, might have been nice to apply the unused check first as a simple tidy-up step? We could then have used this new functionality to guide us tidying away the outdated credentials code in a focused PR.

No worries though, if the team are happy with this!

build.sbt
adamnfish-gu: @aug24 pointed out this compiler option is one of the ok ones, so no harm in including it here going forwards!

aug24:
Split into PRs 1464 and 1465

PR #1464, opened 2026-07-31
Remove app call to outdated credentials

aug24:
## What does this change?

We fully expect to have nothing left, but it helps to tidy up as we go. This PR removes everything from the app that has moved to the iam outdated credentials lambda.


andrew-nowak (approved):
🥳

PR #1472, opened 2026-08-03
Reorganise metrics and add better alerts

aug24:
## What does this change?

Metrics are reorganised so that they are emitted in the lambdas.

We emit the first three metrics from the Lambdas, while the app still emits `vulnerabilities`.  Each of the three is emitted in both Success and Failure forms: the count successfully remediated and the count where remediation failed.

In addition, the lambdas already have an alert for failure of the process as a whole.

```
  private object MetricName extends Enumeration {
    val iamDisableOutdatedKeys = "IamDisableOutdatedKeys"
    val iamDisableAccessKey = "IamDisableAccessKey"
    val iamRemovePassword = "IamRemovePassword"
    val vulnerabilities = "Vulnerabilities"
  }
```

Each of the three lambda metrics is now alerted on (1) >0 remediation failure and (2) extended absence.

Future task: We need to consider what the consumer of the `vulnerabilities` metrics will be.

## What is the value of this?

We need clarity on lambda failure.

### DIFF
--- a/core/src/main/scala/aws/iam/IAMClient.scala
+++ b/core/src/main/scala/aws/iam/IAMClient.scala
@@ -254,9 +254,7 @@ object IAMClient extends LazyLogging {
             logger.info(
               s"Attempt to disable access keys was successful. ${updateAccessKeyResults.length} key(s) were disabled in ${account.name}."
             )
-            if (updateAccessKeyResults.nonEmpty) {
-              Cloudwatch.putIamDisableAccessKeyMetric(ReaperExecutionStatus.success, updateAccessKeyResults.length)
-            }
+            Cloudwatch.putIamDisableAccessKeyMetric(ReaperExecutionStatus.success, updateAccessKeyResults.length)
           }
         )
       )
--- a/core/src/main/scala/aws/iam/IAMClient.scala
+++ b/core/src/main/scala/aws/iam/IAMClient.scala
@@ -10,6 +10,8 @@ import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.iam.IamAsyncClient
 import software.amazon.awssdk.services.iam.model.*
 import utils.attempt.{Attempt, FailedAttempt, Failure}
+import logging.Cloudwatch
+import logging.Cloudwatch.ReaperExecutionStatus
 
 import scala.concurrent.duration.*
 import scala.concurrent.{ExecutionContext, Future}
@@ -17,7 +19,7 @@ import scala.jdk.CollectionConverters.*
 
 object IAMClient extends LazyLogging {
 
-  val SOLE_REGION = Region.of("us-east-1")
+  val SOLE_REGION: Region = Region.of("us-east-1")
 
   /*
    * Note: the report is actually generated a maximum of once every 4 hours.
@@ -86,7 +88,7 @@ object IAMClient extends LazyLogging {
     }
   }
 
-  def getCredentialReportDisplay(
+  private def getCredentialReportDisplay(
       account: AwsAccount,
       currentData: Either[FailedAttempt, CredentialReportDisplay],
       iamClients: AwsClients[IamAsyncClient]
@@ -191,10 +193,23 @@ object IAMClient extends LazyLogging {
       .accessKeyId(accessKeyId)
       .status("Inactive")
       .build()
-    for {
+    (for {
       client <- iamClients.get(awsAccount)
       result <- handleAWSErrs(client)(asScala(client.client.updateAccessKey(request)))
-    } yield result
+    } yield result).tap(
+      _.fold(
+        { failure =>
+          logger.error(s"Failed to disable outdated access key: ${failure.logMessage}")
+          Cloudwatch.putIamDisableOutdatedKeysMetric(ReaperExecutionStatus.failure)
+        },
+        { _ =>
+          logger.info(
+            s"Attempt to disable outdated access key $accessKeyId for IAM user $username was successful in account ${awsAccount.name}."
+          )
+          Cloudwatch.putIamDisableOutdatedKeysMetric(ReaperExecutionStatus.success)
+        }
+      )
+    )
   }
 
   private def handleDeleteLoginProfileErrs(awsClient: AwsClient[IamAsyncClient], username: String)(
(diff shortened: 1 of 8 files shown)


core/src/main/scala/aws/iam/IAMClient.scala
```diff
@@ -191,10 +193,23 @@ object IAMClient extends LazyLogging {
       .accessKeyId(accessKeyId)
       .status("Inactive")
       .build()
-    for {
+    (for {
       client <- iamClients.get(awsAccount)
       result <- handleAWSErrs(client)(asScala(client.client.updateAccessKey(request)))
-    } yield result
+    } yield result).tap(
```
aug24, line 199: Added to emit IamDisableOutdatedKeys metric so we can alert on it.

core/src/main/scala/aws/iam/IAMClient.scala
```diff
@@ -217,4 +232,70 @@ object IAMClient extends LazyLogging {
       result <- handleDeleteLoginProfileErrs(client, username)(asScala(client.client.deleteLoginProfile(request)))
     } yield result
   }
+
```
aug24, line 235: All this block is just moved from the calling code to disentangle.

core/src/main/scala/logging/Cloudwatch.scala
```diff
@@ -12,21 +12,28 @@ import scala.util.{Failure, Success, Try}
...
-  val defaultNamespace = "SecurityHQ"
+  private val defaultNamespace = "SecurityHQ"
 
   object DataType extends Enumeration {
-    val s3Total = Value("s3/total")
-    val iamCredentialsTotal = Value("iam/credentials/total")
-    val iamCredentialsCritical = Value("iam/credentials/critical")
-    val iamCredentialsWarning = Value("iam/credentials/warning")
-    val iamKeysTotal = Value("iam/keys/total")
+    val s3Total: Value = Value("s3/total")
```
aug24, line 20: Add types.

core/src/main/scala/logging/Cloudwatch.scala
```diff
@@ -12,21 +12,28 @@ import scala.util.{Failure, Success, Try}
...
+    val iamCredentialsTotal: Value = Value("iam/credentials/total")
+    val iamCredentialsCritical: Value = Value("iam/credentials/critical")
+    val iamCredentialsWarning: Value = Value("iam/credentials/warning")
+    val iamKeysTotal: Value = Value("iam/keys/total")
+  }
+
+  private object MetricName extends Enumeration {
```
aug24, line 27: Create an enumeration rather than scatter values in code.

core/src/main/scala/logging/Cloudwatch.scala
```diff
@@ -73,7 +80,16 @@ object Cloudwatch extends LazyLogging {
...
     putMetric(
       defaultNamespace,
-      "IamDisableAccessKey",
+      MetricName.iamDisableAccessKey,
+      Seq(("ReaperExecutionStatus", reaperExecutionStatus.toString)),
+      1
+    )
+  }
+
+  def putIamDisableOutdatedKeysMetric(reaperExecutionStatus: ReaperExecutionStatus.Value): Unit = {
```
aug24: Add a metrics emitter for success/fail of disabling outdated keys, which was not present.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -120,6 +121,7 @@ class IamOutdatedCredentials(
         securityNotificationIdMaybe <- sendSecurityNotification(notificationTopicArn, notificationDevXSecurityMaybe)
 
         // only now do we actually disable the credential
+        _ = Cloudwatch.putIamDisableOutdatedKeysMetric(ReaperExecutionStatus.success, 0)
```
aug24: First emit a value of zero, in case the list is empty, then we won't have missing metrics.

iam-unrecognised-users/src/main/scala/unrecognised/UnrecognisedUsers.scala
```diff
@@ -78,16 +80,19 @@ object UnrecognisedUsers extends LazyLogging {
...
       accountCredsReports = getCredsReportDisplayForAccount(credentialReports.toMap)
       // determine the unrecognised users by comparing Janus usernames to the IAM users (filtered to allowed accounts)
       unrecognisedUsers = unrecognisedUsersForAllowedAccounts(accountCredsReports, janusUsernames, allowedAccountIds)
       accessKeys <- Attempt.traverse(unrecognisedUsers)(listAccountAccessKeys(_, iamClients))
+
       // deactivate access keys and remove login profiles for unrecognised users (skipped in dry run)
-      _ <- Attempt.traverse(accessKeys)(disableAccountAccessKeys(_, iamClients, settings.dryRun)).map(_ => ())
-      _ <- Attempt.traverse(unrecognisedUsers)(removeAccountPasswords(_, iamClients, settings.dryRun)).map(_ => ())
+      // First emit metrics of zero, in case there are none.
+      _ = Cloudwatch.putIamDisableAccessKeyMetric(ReaperExecutionStatus.success, 0)
```
aug24, line 92: Same here - avoid missing metrics on an empty list by emitting a zero first for both data sets.

core/src/main/scala/aws/iam/IAMClient.scala
```diff
@@ -217,4 +232,69 @@ object IAMClient extends LazyLogging {
...
+            logger.error(s"Failed to disable access key: ${failure.logMessage}")
+            Cloudwatch.putIamDisableAccessKeyMetric(ReaperExecutionStatus.failure)
+          },
+          { updateAccessKeyResults =>
+            logger.info(
+              s"Attempt to disable access keys was successful. ${updateAccessKeyResults.length} key(s) were disabled in ${account.name}."
+            )
+            if (updateAccessKeyResults.nonEmpty) {
+              Cloudwatch.putIamDisableAccessKeyMetric(ReaperExecutionStatus.success)
+            }
```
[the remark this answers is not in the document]
aug24: This is tricky.  As is, it emits 1 for every user.  As suggested by Copilot, it emits 1 for every key disabled for each user.  So currently a user with 3 disabled keys gets 1, with this suggestion, we get 3.

I think I'll take this as the metric is called IamDisableAccessKey.

PR #1475, opened 2026-08-04
Reorganise metrics

aug24:
## What does this change?

Metrics are reorganised so that they are emitted in the lambdas.

We emit the first three metrics from the Lambdas, while the app still emits `vulnerabilities`.  Each of the three is emitted in both Success and Failure forms: the count successfully remediated and the count where remediation failed.

```
  private object MetricName extends Enumeration {
    val iamDisableOutdatedKeys = "IamDisableOutdatedKeys"
    val iamDisableAccessKey = "IamDisableAccessKey"
    val iamRemovePassword = "IamRemovePassword"
    val vulnerabilities = "Vulnerabilities"
  }
```

In addition, the result of metrics is now included in the result instead of fire-and-forget in an unchecked Future (in an Attempt).  This is now included in the tests (where it couldn't be, before).

### DIFF
--- a/core/src/main/scala/aws/iam/IAMClient.scala
+++ b/core/src/main/scala/aws/iam/IAMClient.scala
@@ -103,7 +103,7 @@ object IAMClient extends LazyLogging {
       account: AwsAccount,
       iamClients: AwsClients[IamAsyncClient],
       delay: FiniteDuration = 3.seconds
-  )(implicit executionContext: ExecutionContext): Attempt[CredentialReportDisplay] = {
+  )(using ExecutionContext): Attempt[CredentialReportDisplay] = {
     for {
       client <- iamClients.get(account)
       _ <- Retry.until(
@@ -218,11 +218,12 @@ object IAMClient extends LazyLogging {
     for {
       client <- iamClients.get(awsAccount)
       result <- handleDeleteLoginProfileErrs(client, username)(asScala(client.client.deleteLoginProfile(request)))
-      message = if (result.isDefined) {
-        s"Deleted login profile for IAM user $username in account ${awsAccount.name}."
-      } else {
-        s"No login profile found for IAM user $username in account ${awsAccount.name}; nothing to delete."
-      }
+      message =
+        if (result.isDefined) {
+          s"Deleted login profile for IAM user $username in account ${awsAccount.name}."
+        } else {
+          s"No login profile found for IAM user $username in account ${awsAccount.name}; nothing to delete."
+        }
       _ = logger.info(message)
     } yield result
   }
--- a/core/src/main/scala/aws/iam/IAMClient.scala
+++ b/core/src/main/scala/aws/iam/IAMClient.scala
@@ -17,7 +17,7 @@ import scala.jdk.CollectionConverters.*
 
 object IAMClient extends LazyLogging {
 
-  val SOLE_REGION = Region.of("us-east-1")
+  val SOLE_REGION: Region = Region.of("us-east-1")
 
   /*
    * Note: the report is actually generated a maximum of once every 4 hours.
@@ -86,7 +86,7 @@ object IAMClient extends LazyLogging {
     }
   }
 
-  def getCredentialReportDisplay(
+  private def getCredentialReportDisplay(
       account: AwsAccount,
       currentData: Either[FailedAttempt, CredentialReportDisplay],
       iamClients: AwsClients[IamAsyncClient]
@@ -194,6 +194,9 @@ object IAMClient extends LazyLogging {
     for {
       client <- iamClients.get(awsAccount)
       result <- handleAWSErrs(client)(asScala(client.client.updateAccessKey(request)))
+      _ = logger.info(
+        s"Disabled access key $accessKeyId for IAM user $username in account ${awsAccount.name}."
+      )
     } yield result
   }
 
@@ -215,6 +218,13 @@ object IAMClient extends LazyLogging {
     for {
       client <- iamClients.get(awsAccount)
       result <- handleDeleteLoginProfileErrs(client, username)(asScala(client.client.deleteLoginProfile(request)))
+      message = if (result.isDefined) {
+        s"Deleted login profile for IAM user $username in account ${awsAccount.name}."
+      } else {
+        s"No login profile found for IAM user $username in account ${awsAccount.name}; nothing to delete."
+      }
+      _ = logger.info(message)
     } yield result
   }
+
 }
--- a/core/src/main/scala/logging/Cloudwatch.scala
+++ b/core/src/main/scala/logging/Cloudwatch.scala
@@ -12,21 +12,28 @@ import scala.util.{Failure, Success, Try}
 
 object Cloudwatch extends LazyLogging {
 
-  val cloudwatchClient = CloudWatchClient.builder.build()
+  private val cloudwatchClient = CloudWatchClient.builder.build()
 
-  val defaultNamespace = "SecurityHQ"
+  private val defaultNamespace = "SecurityHQ"
 
   object DataType extends Enumeration {
-    val s3Total = Value("s3/total")
-    val iamCredentialsTotal = Value("iam/credentials/total")
-    val iamCredentialsCritical = Value("iam/credentials/critical")
-    val iamCredentialsWarning = Value("iam/credentials/warning")
-    val iamKeysTotal = Value("iam/keys/total")
+    val s3Total: Value = Value("s3/total")
+    val iamCredentialsTotal: Value = Value("iam/credentials/total")
+    val iamCredentialsCritical: Value = Value("iam/credentials/critical")
+    val iamCredentialsWarning: Value = Value("iam/credentials/warning")
+    val iamKeysTotal: Value = Value("iam/keys/total")
+  }
+
+  private object MetricName extends Enumeration {
+    val iamDisableOutdatedKeys: Value = Value("IamDisableOutdatedKeys")
+    val iamDisableAccessKey: Value = Value("IamDisableAccessKey")
+    val iamRemovePassword: Value = Value("IamRemovePassword")
+    val vulnerabilities: Value = Value("Vulnerabilities")
   }
 
   object ReaperExecutionStatus extends Enumeration {
-    val success = Value("Success")
-    val failure = Value("Failure")
+    val success: Value = Value("Success")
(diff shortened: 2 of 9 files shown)


core/src/main/scala/aws/iam/IAMClient.scala
```diff
@@ -184,7 +186,8 @@ object IAMClient extends LazyLogging {
       awsAccount: AwsAccount,
       username: String,
       accessKeyId: String,
-      iamClients: AwsClients[IamAsyncClient]
+      iamClients: AwsClients[IamAsyncClient],
+      tap: Either[FailedAttempt, UpdateAccessKeyResponse] => Unit = _ => {}
```
aug24: Introduced so that the logging / metering is defined on the calling code side.

core/src/main/scala/logic/IamUnrecognisedUsers.scala
```diff
@@ -99,70 +96,6 @@ object IamUnrecognisedUsers extends LazyLogging {
     }
   }
 
-  def disableAccountAccessKeys(
```
aug24, line 102: This has moved to `UnrecognisedUsers` (not to be confused with `IamUnrecognisedUsers` because it's only ever called from there.

iam-unrecognised-users/src/main/scala/unrecognised/UnrecognisedUsers.scala
```diff
@@ -78,23 +83,82 @@ object UnrecognisedUsers extends LazyLogging {
...
+    result.fold(
+      { (failure: FailedAttempt) =>
+        logger.error(s"failed to delete at least one password: ${failure.logMessage}")
+        Cloudwatch.putIamRemovePasswordMetric(ReaperExecutionStatus.failure, 1)
+      },
+      { (_: T) => Cloudwatch.putIamRemovePasswordMetric(ReaperExecutionStatus.success, 1) }
+    )
+  }
+
+  private[unrecognised] def disableAccountAccessKeys(
```
aug24, line 134: Moved here from `IamUnrecognisedUsers`

iam-unrecognised-users/src/main/scala/unrecognised/UnrecognisedUsers.scala
```diff
@@ -78,23 +83,82 @@ object UnrecognisedUsers extends LazyLogging {
...
+      _ <- Attempt.traverse(accessKeys)(disableAccountAccessKeys(_, iamClients, settings.dryRun))
+      _ <- Attempt.traverse(unrecognisedUsers)(removeAccountPasswords(_, iamClients, settings.dryRun))
       // construct and send a notification for each unrecognised user
       notifications = unrecognisedUserNotifications(unrecognisedUsers, settings.dryRun)
       // if in dry run, notifications list is empty
       notificationIds <- Attempt.traverse(notifications)(AnghammaradNotifications.send(_, anghammaradSnsArn, snsClient))
     } yield notificationIds
   }
 
+  private def removeAccountPasswords(
```
aug24, line 105: Moved here from `IamUnrecognisedUsers`

core/src/main/scala/logic/IamUnrecognisedUsers.scala
```diff
@@ -99,70 +96,6 @@ object IamUnrecognisedUsers extends LazyLogging {
...
-      if (accountUnrecognisedKeys.vulnerableAccessKey.nonEmpty) {
-        logger.info(
-          s"DRY RUN: Would disable access keys in account '${accountUnrecognisedKeys.account.name}' for IAM users: ${accountUnrecognisedKeys.vulnerableAccessKey.map(_.username).mkString("'", "', '", "'")}."
-        )
-      }
-      Attempt.Right(Nil)
-    }
-  }
-
-  def removeAccountPasswords(
```
aug24, line 139: This has moved to `UnrecognisedUsers` (not to be confused with `IamUnrecognisedUsers` because it's only ever called from there.

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -130,18 +132,28 @@ class IamOutdatedCredentials(
...
-        )
-      } yield List(userNotificationId) ++ securityNotificationIdMaybe
+        _ <- IAMClient
+          .disableAccessKey(
+            awsAccount,
+            credentialToDisable.username,
+            credentialToDisable.accessKeyId,
+            iamClients
+          )
+      } yield List(userNotificationId) ++ securityNotificationIdMaybe).tap(emitMetricsTap)
```
aug24: `.tap` is where we emit non-zero metrics

iam-unrecognised-users/src/main/scala/unrecognised/UnrecognisedUsers.scala
```diff
@@ -78,23 +83,82 @@ object UnrecognisedUsers extends LazyLogging {
...
+  private[unrecognised] def disableAccountAccessKeys(
+      accountUnrecognisedKeys: AccountUnrecognisedAccessKeys,
+      iamClients: AwsClients[IamAsyncClient],
+      dryRun: Boolean
+  )(implicit ec: ExecutionContext): Attempt[List[UpdateAccessKeyResponse]] = {
+    if (!dryRun) {
+      val AccountUnrecognisedAccessKeys(account, accessKeys) = accountUnrecognisedKeys
+      val activeAccessKeys = accessKeys.filter(_.status == CredentialActive)
+      Attempt.traverse(activeAccessKeys)(key =>
+        IAMClient.disableAccessKey(account, key.username, key.accessKeyId, iamClients).tap(emitDisabledAccessKeyMetrics)
```
aug24: `.tap` is where we emit non-zero metrics

iam-outdated-credentials/src/main/scala/logic/IamOutdatedCredentials.scala
```diff
@@ -262,6 +274,8 @@ class IamOutdatedCredentials(
       // we won't execute these operations, but can log them instead
       _ = filteredOperations.operationsOnAccountsThatAreNotAllowed.foreach(logOperationOnly)
 
+      // First record a zero so that we know we ran, even if there ends up being nothing to do.
+      _ = Cloudwatch.putIamDisableOutdatedKeysMetric(ReaperExecutionStatus.success, 0)
```
aug24, line 283: Emit a zero count.  If we don't have anything to do, we don't want to trigger the missing data alarm.

iam-unrecognised-users/src/main/scala/unrecognised/UnrecognisedUsers.scala
```diff
@@ -78,23 +83,86 @@ object UnrecognisedUsers extends LazyLogging {
...
         .map(_.tap(logCredentialReportResults))
       accountCredsReports = getCredsReportDisplayForAccount(credentialReports.toMap)
       // determine the unrecognised users by comparing Janus usernames to the IAM users (filtered to allowed accounts)
       unrecognisedUsers = unrecognisedUsersForAllowedAccounts(accountCredsReports, janusUsernames, allowedAccountIds)
       accessKeys <- Attempt.traverse(unrecognisedUsers)(listAccountAccessKeys(_, iamClients))
       // deactivate access keys and remove login profiles for unrecognised users (skipped in dry run)
-      _ <- Attempt.traverse(accessKeys)(disableAccountAccessKeys(_, iamClients, settings.dryRun)).map(_ => ())
-      _ <- Attempt.traverse(unrecognisedUsers)(removeAccountPasswords(_, iamClients, settings.dryRun)).map(_ => ())
+      // First emit a zero metric in case there's no keys to disable
+      _ = Cloudwatch.putIamDisableAccessKeyMetric(ReaperExecutionStatus.success, 0)
```
aug24, line 93: Emit a zero count. If we don't have anything to do, we don't want to trigger the missing data alarm.

iam-unrecognised-users/src/main/scala/unrecognised/UnrecognisedUsers.scala
```diff
@@ -78,23 +83,86 @@ object UnrecognisedUsers extends LazyLogging {
...
+      _ <- Attempt.traverse(accessKeys)(disableAccountAccessKeys(_, iamClients, settings.dryRun))
+      // First emit a zero metric in case there's no passwords to remove
+      _ = Cloudwatch.putIamRemovePasswordMetric(ReaperExecutionStatus.success, 0)
```
aug24: Emit a zero count. If we don't have anything to do, we don't want to trigger the missing data alarm.

iam-unrecognised-users/src/main/scala/unrecognised/UnrecognisedUsers.scala
aug24: This applies to removeAccountPasswords - it's got slightly confused!

core/src/main/scala/logging/Cloudwatch.scala
```diff
@@ -1,102 +1,218 @@
...
+    * @param onSuccess
+    *   The action to perform if the result is successful, typically emitting a success metric.
+    * @param onFailure
+    *   The action to perform if the result is a failure, typically emitting a failure metric.
+    * @param actionLabel
+    *   A label describing the action being performed, used for logging purposes.
+    * @return
+    *   An Attempt that represents the outcome of the original result and the metric emission.
+    */
+  def emitOutcomeMetric[T](
```
aug24, line 199: This has been abstracted out to here to make sure we have consistent behaviour.

We get the result of an attempt.  We attempt to conditionally emit a success or failure metric according to that.

If the metric attempt fails, we return that.  If not, we return the original failure, or unit.

This also has the benefit of making testing much easier by overriding the success and failure actions.

core/src/main/scala/logging/Cloudwatch.scala
```diff
@@ -1,102 +1,178 @@
...
+                friendlyMessage = "Failed to log cloudwatch metric",
+                statusCode = 500,
+                context = None,
+                throwable = None
+              )
+            )
+          )
+        )
+      }
+      .map(_ => ())
```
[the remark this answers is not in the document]
andrew-nowak: Isn't this a problem? Even if all 3 putMetrics succeed, the overall return will be a failure

andrew-nowak (approved):

PR #1476, opened 2026-08-04
Add better alerts

aug24:
## What does this change?

Each of the three lambda metrics is now alerted on (1) >0 remediation failure and (2) extended absence.

In addition, the lambdas already have an alert for failure of the process as a whole.

Future task: We need to consider what the consumer of the `vulnerabilities` metrics will be or remove them.

### DIFF
--- a/cdk/lib/security-hq.ts
+++ b/cdk/lib/security-hq.ts
@@ -346,6 +346,9 @@ export class SecurityHQ extends GuStack {
         namespace: "SecurityHQ",
         period: Duration.days(3), // we do not run sat/sun
         statistic: "sum",
+        dimensionsMap: {
+          ReaperExecutionStatus: "Success",
+        },
       }),
       treatMissingData: TreatMissingData.BREACHING,
       comparisonOperator: ComparisonOperator.LESS_THAN_THRESHOLD,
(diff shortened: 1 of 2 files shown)


cdk/lib/security-hq.ts
```diff
@@ -308,6 +332,67 @@ export class SecurityHQ extends GuStack {
...
+        " Check the application logs for more details https://logs.gutools.co.uk/s/devx/goto/f9915a6e4e94a000732d67026cea91be.",
+      snsTopicName: notificationTopic.topicName,
+      threshold: 0,
+      evaluationPeriods: 1,
+      metric: new Metric({
+        metricName: "IamDisableAccessKey",
+        namespace: "SecurityHQ",
+        period: Duration.days(3), // we do not run sat/sun
+        statistic: "sum",
+      }),
```
andrew-nowak, line 352: These metrics are only published with `ReaperExecutionStatus` dimensions, so I think you need to specify them for the alarm to pick up the right metric (I don't think Cloudwatch will aggregate across all the dimensions, although it'd be nice if it did https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#Dimension:~:text=CloudWatch%20does%20not%20aggregate%20across%20dimensions%20for%20your%20custom%20metrics.)

andrew-nowak (approved):

PR #1490, opened 2026-08-11
Remove SHQ App

aug24:
## What does this change?

We are removing all the Security HQ App code.

Current step: *Remove the resources from the cloudformation stack*

To follow:

 * Remove the app from the WAF list (https://github.com/guardian/waf/pull/96)
 * Remove all unused code (https://github.com/guardian/security-hq/pull/1492)
 * Remove / update documentation and scripts (https://github.com/guardian/security-hq/pull/1493)

## Any additional notes?

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1216038883987064


andrew-nowak (approved):
LGTM. I'm not sure what will happen if you deploy this before the WAF PR though, will AWS be happy to delete an ALB which is still registered with a WAF?

Remember also that deletion protection is turned on for the ALB which will need turning off manually for the deploy to complete

PR #1491, opened 2026-08-11
Verify Janus data export before running unrecognised users process

andrew-nowak:
## What does this change?

Verify that the Janus data export is fresh (generated within seven days) before running the unrecognised user lambda. We were already throwing if the data export was not in the expected format, but that's now better handled into the standard `Attempt` error handling pattern. We also check that the number of users is a plausible amount (more than 10, ie. the export contained some users, and wasn't empty).

## What is the value of this?

Greater confidence that the app won't attempt to remediate new users that have only recently been added to the Janus config or fail to remediate recent leavers who have been removed from Janus.

### DIFF
--- a/core/src/main/scala/aws/s3/S3.scala
+++ b/core/src/main/scala/aws/s3/S3.scala
@@ -1,7 +1,8 @@
 package aws.s3
 
 import model.{BucketEncryptionResponse, BucketNotFound, Encrypted, NotEncrypted}
-import software.amazon.awssdk.core.ResponseInputStream
+import software.amazon.awssdk.core.ResponseBytes
+import software.amazon.awssdk.core.sync.ResponseTransformer
 import software.amazon.awssdk.services.s3.S3Client
 import software.amazon.awssdk.services.s3.model.{
   GetBucketEncryptionRequest,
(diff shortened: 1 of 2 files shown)


core/src/main/scala/aws/s3/S3.scala
```diff
@@ -1,20 +1,24 @@
...
 
 object S3 {
-  def getS3Object(s3Client: S3Client, bucket: String, key: String): Attempt[BufferedSource] = {
+  def getS3Object(s3Client: S3Client, bucket: String, key: String): Attempt[ResponseInputStream[GetObjectResponse]] = {
     val request = GetObjectRequest.builder().bucket(bucket).key(key).build()
     try {
       Attempt.Right {
-        scala.io.Source
-          .fromInputStream(s3Client.getObject(request))
+        s3Client.getObject(request)
```
aug24: Might be nice to go the whole hog and do this the same way as `loadConfigFromS3` in CoreConfig?

iam-unrecognised-users/src/main/scala/unrecognised/UnrecognisedUsers.scala
```diff
@@ -59,6 +66,54 @@ object UnrecognisedUsers extends LazyLogging {
...
     Left(Failure.notYetLoaded(account.id, "credentials").attempt)
 
+  private def janusConfigIsFresh(objectResponse: GetObjectResponse): Attempt[Boolean] = {
+    val lastModified = objectResponse.lastModified()
+    val isStale = lastModified.isBefore(Instant.now().minus(7, DAYS))
+    if (isStale) {
+      Attempt.Left(
+        FailedAttempt(
+          Failure(
+            "Janus config has not been updated in over 7 days",
```
aug24: Suggest:
1) Give the magic number 7 a name 
2)  Make it less than that.  We never go more than a few days without an access change.
3) Put the lastModified value in the error message

andrew-nowak: 1 & 3: 👍 
2: _never_? We'll go at least 4 days without an access change every Easter, and probably at least that amount every Christmas too. I'd expect the days either side of that to have lower quantities of changes too, since it's likely a lot of people will be off around those dates. Once we take that into account we're up to 5 or 6 days, at which point it might as well be a round week

aug24 (approved):

